PR #7, opened 2018-11-01
[MM-11825] Unify build process 

cpanato:


hanzei:
I think `server/.depensure` should not be checked in.

cpanato:
yep @hanzei thanks!

PR #14, opened 2018-12-23
Release 1.0.4

hanzei:
Since v1.0.3 is broken we should get this out rather soon.


jwilander:
@hanzei did you test your fix with a JIRA instance? I can do it if you haven't already

hanzei:
FYI: Joram will test the release.

PR #22, opened 2019-02-14
MM-13967: Add support for more issue events

levb:
- Fixed MM-13967: Added support for various ticket update events. Did not add support for worklogs and issue links because these events have only limited data
- Fixed MM-8650: removed enable/disable
- Replaced SlackAttachments with regular markdown for display


jasonblais:
Awesome!! 🎉 

> Did not add support for worklogs and issue links because these events have only limited data

is the data so limited, that any info we post back would be more-or-less useless?

Not a big deal if we don't support them (less important than the other issue-related events). Just curious so that when someone asks why we don't support it, I'll have the justification ready (we have at least one request for worklogs).

> Replaced SlackAttachments with regular markdown for display

UX and I can work out the formatting after this is merged - thinking it might be easier when we see the data sent back.

Curious if there was a reason for replacing SlackAttachments with regular markdown?

server/configuration.go
```diff
@@ -21,7 +21,6 @@ import (
 // If you add non-reference types to your configuration struct, be sure to rewrite Clone as a deep
 // copy appropriate for your types.
 type configuration struct {
-	Enabled  bool
```
hanzei: :+1: on removing this

server/plugin.go
```diff
@@ -28,102 +29,88 @@ func (p *Plugin) ServeHTTP(c *plugin.Context, w http.ResponseWriter, r *http.Req
 	p.API.LogDebug("HTTP request", "Host", r.Host, "RequestURI", r.RequestURI, "Method", r.Method, "query", r.URL.Query().Encode())
 
 	config := p.getConfiguration()
-	if !config.Enabled || config.Secret == "" || config.UserName == "" {
-		errorMessage := "This plugin is not configured"
-		p.postHTTPDebugMessage(errorMessage)
-		http.Error(w, errorMessage, http.StatusForbidden)
+	if config.Secret == "" || config.UserName == "" {
+		http.Error(w, p.debug("JIRA plugin not configured correctly; must provide Secret and UserName"), http.StatusForbidden)
```
hanzei: Maybe we can just warp this whole function into a new one `handleRequest(r *http.Request) (int, error)` and get rid of this duplicate code pieces?

0/5 on the function name
2/5 that the error handling is not optimal

server/webhook.go
```diff
@@ -4,29 +4,36 @@
...
 
 import (
-	"bytes"
-	"errors"
+	"fmt"
 	"strings"
-	"text/template"
-
-	"github.com/mattermost/mattermost-server/model"
+	// "github.com/mattermost/mattermost-server/model"
```
hanzei: Mind removing this?

server/webhook.go
```diff
@@ -38,145 +45,327 @@ type Webhook struct {
...
+	md_RootStyle   = "## "
+	md_UpdateStyle = "###### "
+)
+
+var formatters = map[string]func(w *Webhook) string{
+	"jira:issue_created": mdIssueCreated,
+	"jira:issue_updated": mdIssueUpdated,
+	"comment_created":    mdComment,
+	"comment_updated":    mdComment,
+	"comment_deleted":    mdComment,
```
hanzei: Interesting way to do this :+1:

server/webhook.go
```diff
@@ -38,145 +45,327 @@ type Webhook struct {
...
+			return fmt.Sprintf("%v %v", strings.ToLower(to), md_IssueLongLink(w))
+
+		case item.Field == "Attachment":
+			return fmt.Sprintf("%v %v", md_AddRemove(from, to, "attached", "removed attachments"), md_IssueLongLink(w))
+
+		case item.Field == "labels":
+			return fmt.Sprintf("%v %v", md_AddRemove(from, to, "added labels", "removed labels"), md_IssueLongLink(w))
+
+		default:
+			// return fmt.Sprintf("updated %v from %v to %v on %v", item.Field, from, to, md_IssueLongLink(w))
```
hanzei: Do we really want a default case here? Is the list of Fields exhaustive? Then we can add an log message.

hanzei:
I propose to use https://ci-linux-mysql.mattermost.com for testing this. I can set up a test environment, once the initial review is done.

hanzei:
I propose to use https://ci-linux-mysql.mattermost.com for testing this. I can set up a test environment, once the initial review is done.

server/plugin.go
levb: -negative -2/5 on breaking up the function, it's a good single-purpose webhook handler now. When we have more kinds of http messages, yes.
- strongly agreed that the error handling is not optimal, out of scope for now

server/webhook.go
levb: I'll change it back to a switch actually, now that you pointed it out. I was anticipating more different event types when I set it up, but ended up with only a handful. a 'switch' would be consistent with other places.

server/webhook.go
levb: No we don't. This is where I was just outputting raw JSON before to see what was coming through. Will remove.

server/webhook.go
```diff
@@ -38,145 +45,327 @@ type Webhook struct {
...
+		headline = md_HeadlineFromChangeLog(w)
+		if w.showDescription {
+			extra = md_IssueDescription(w)
 		}
-	case "jira:issue_deleted":
-		if w.Issue.Fields.Resolution != nil {
-			return nil, nil
+	}
+	if headline == "" {
+		// return fmt.Sprintf("```%v```\n", w.RawJSON)
```
jwilander: Unneeded comment?

server/webhook.go
```diff
@@ -38,145 +45,327 @@ type Webhook struct {
...
+	s += md_User(&w.Comment.UpdateAuthor) + " " + headline + " " + md_IssueHashtags(w)
+	s += "\n"
+	if extra != "" {
+		s += extra
+	}
+	s += "\n"
+	return s
+}
+
+func md_HeadlineFromChangeLog(w *Webhook) string {
```
jwilander: Some functions are named without the underscore and some have it (eg `mdComment` vs `md_HeadlineFromChangeLog `). Is that an intentional distinction or just accidental?

levb: This was intentional (top-level vs "blocks") but thinking about your comment, and being true to the Go style I'll remove the `_`s.

levb:
@jasonblais:
RE: _"...the data so limited, that any info we post back would be more-or-less useless"_: there are neither issue keys nor the user data readily available in these messages. There are internal IDs, so with proper auth/API usage we can get the details later, but the webhooks themselves appear very limited. So, if we need it, will come in v2.

RE: Reason for replacing SlackAttachments with regular markdown. Just seems a lot cleaner without them.Was there a reason that they were used to start with?

levb:
@jasonblais @jwilander @hanzei @DHaussermann I updated the code with the PR feedback, you can see it in action at https://i-0cba36c59c7d65ed7.test.spinmint.com/ad-1/channels/autem-2 (sysadmin/sysadmin). It posts from https://levbrouk.atlassian.net/secure/RapidBoard.jspa?rapidView=1&view=planning.nodetail&selectedIssue=TES-39.

jasonblais:
@levb Sounds good for issue links and worklogs, not a top priority so let's cut. That explanation helps when going back to customer(s).

Re: attachments, that was a design decision in V1. However, Asaad and I are working on updating them for V2, and we'll compare how the messages look with just plain Markdown.

server/plugin.go
levb: Apologies, completely misunderstood your comment about wrapping into a function; did that now, muchas gracias for the suggestion, much cleaner.

levb:
@hanzei I think I got your change request covered, sorry for the initial misunderstanding.

@jasonblais please confirm with regards to whether to commit this with markdown messages once fully dev-approved, or hold until further UX review.

jasonblais:
@levb Please hold for now, I got UI designs from Asaad (unrelated to changes in this PR), will see how the designs compare with the Markdown messages.

hanzei:
My 2 cents on attachments:

For me (and this is a personal opinion) attachments are the "bot way" to show information in Mattermost. When I search for e.g. a new JIRA issue in a channel, I just look out for the style of attachments. It help keeping bot messages separated from user messages.

I know of the technical flaws of slack attachments, but I think this we have decided to have them this way in our system and should stick (for now) with this.

2/5 on sticking with attachments.

hanzei:
@levb Is this ready for a final round of review?

levb:
@hanzei Yes it is. I rewrote much of it to simplify dual-mode md/slack attachment, but stopped short of making it configurable.

jasonblais:
Thanks @levb. It would be good to see the formatting of the new events before merging.

jasonblais:
@levb As discussed last week, we can merge and test on community-daily. We have designs from Asaad, but they can be done on a separate PR.

PR #41, opened 2019-05-01
Proposed text updates for /jira slash commands

jasonblais:
Feedback welcome. Includes
 - text updates for /jira help
 - text updates for error responses
 - JIRA > Jira

### DIFF
--- a/server/command.go
+++ b/server/command.go
@@ -10,21 +10,21 @@ import (
 	"github.com/mattermost/mattermost-server/plugin"
 )
 
-const helpText = "###### Mattermost JIRA Plugin - Slash Command Help\n" +
-	"* `/jira connect` - Connect your Mattermost account to your JIRA account\n" +
-	"* `/jira disconnect` - Disonnect your Mattermost account to your JIRA account\n" +
-	"* `/jira instance` - Manage JIRA instances connected to Mattermost\n" +
-	"  * `list` - List known JIRA instances\n" +
-	"  * `select <key or number>` - Select a known instance as current\n" +
-	"  * `add server <URL>` - Add a JIRA Server instance\n" +
-	"  * `add cloud` - Add a JIRA Cloud instance\n" +
+const helpText = "###### Mattermost Jira Plugin - Slash Command Help\n" +
+	"* `/jira connect` - Connect your Mattermost account to your Jira account and subscribe to events\n" +
+	"* `/jira disconnect` - Disonnect your Mattermost account from your Jira account\n" +
+	"* `/jira instance [add/list/select]` - Manage connected Jira instances. Must have System Admin role in Mattermost\n" +
+	"  * `/jira instance add server <URL>` - Connect a Jira Server instance to Mattermost\n" +
+	"  * `/jira instance add cloud` - Connect a Jira Cloud instance to Mattermost\n" +
+	"  * `/jira instance list` - List connected Jira instances\n" +
+	"  * `/jira instance select <key or number>` - Select the active Jira instance. At most one active Jira instance is currently supported\n" +
 	""
 
 func getCommand() *model.Command {
 	return &model.Command{
 		Trigger:          "jira",
-		DisplayName:      "JIRA",
-		Description:      "Integration with JIRA.",
+		DisplayName:      "Jira",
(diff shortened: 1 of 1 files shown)


server/command.go
```diff
@@ -10,21 +10,21 @@ import (
...
-	"* `/jira instance` - Manage JIRA instances connected to Mattermost\n" +
-	"  * `list` - List known JIRA instances\n" +
-	"  * `select <key or number>` - Select a known instance as current\n" +
-	"  * `add server <URL>` - Add a JIRA Server instance\n" +
-	"  * `add cloud` - Add a JIRA Cloud instance\n" +
+const helpText = "###### Mattermost Jira Plugin - Slash Command Help\n" +
+	"* `/jira connect` - Connect your Mattermost account to your Jira account and subscribe to events\n" +
+	"* `/jira disconnect` - Disonnect your Mattermost account from your Jira account\n" +
+	"* `/jira instance [add/list/select]` - Manage connected Jira instances. Must have System Admin role in Mattermost\n" +
+	"  * `/jira instance add server <URL>` - Connect a Jira Server instance to Mattermost\n" +
```
levb: LOL, I added `/jira instance`  for debugging purposes. If we want to keep  these commands we should add permission checks to the code.

server/command.go
```diff
@@ -10,21 +10,21 @@ import (
...
+	"  * `/jira instance add cloud` - Connect a Jira Cloud instance to Mattermost\n" +
+	"  * `/jira instance list` - List connected Jira instances\n" +
+	"  * `/jira instance select <key or number>` - Select the active Jira instance. At most one active Jira instance is currently supported\n" +
```
levb: s/key/URL/

server/command.go
```diff
@@ -78,29 +78,29 @@ func executeInstance(p *Plugin, c *plugin.Context, args []string) *model.Command
...
+		return responsef("Failed to load current Jira instance: %v", err)
 	}
 
 	keys := []string{}
 	for key := range known {
 		keys = append(keys, key)
 	}
 	sort.Strings(keys)
-	text := "Known JIRA instances (selected instance is **bold**)\n\n| |Key|Type|\n|--|--|--|\n"
+	text := "Known Jira instances (selected instance is **bold**)\n\n| |Key|Type|\n|--|--|--|\n"
```
levb: s/Key/URL/ (the "key" is always the URL of the instance, so might as well call it such)

server/command.go
```diff
@@ -166,11 +166,11 @@ func executeInstanceSelect(p *Plugin, c *plugin.Context, args []string) *model.C
 
 	ji, err := p.LoadJIRAInstance(instanceKey)
 	if err != nil {
-		return responsef("failed to load JIRA instance %s: %v", instanceKey, err)
+		return responsef("Failed to load JIRA instance %s: %v", instanceKey, err)
```
levb: s/JIRA/Jira/

server/command.go
```diff
@@ -166,11 +166,11 @@ func executeInstanceSelect(p *Plugin, c *plugin.Context, args []string) *model.C
...
 	}
 	err = p.StoreJIRAInstance(ji, true)
 	if err != nil {
-		return responsef("failed to store JIRA instance %s: %v", instanceKey, err)
+		return responsef("Failed to store JIRA instance %s: %v", instanceKey, err)
```
levb: ditto

server/command.go
cpoile: I would change to `<number or URL>` to match the order in the table.

Also, someone should probably change the actual table column headings to be correct (I'll do that in my PR). :)

server/command.go
cpoile: ah, this is where we can change to:
```
"Known JIRA instances (selected instance is **bold**)\n\n|Key|URL|Type|\n|--|--|--|\n"
```

server/command.go
jasonblais: Cool, I thought it had a permission check. I wouldn't worry about it though, will just remove from the text.

I'm aware it's just for debugging purposes at this point :)

jasonblais:
@cpoile @levb Feedback addressed, thanks for the quick review!

PR #43, opened 2019-05-03
[MM-15425] Improve plugin deploy/redeploy/debug cycle

cpoile:
Improve the workflow for plugin development. 

make debug command:
- only recompile for the architecture the developer is actually using (speeds up the change/test cycle)
- compile with debug flags for the server plugin (allows us to attach a dlv process and debug the server component. I'll write instructions.)

make webapp-debug command:
- compile in dev mode (with source maps) for the webapp
- setup npm correctly so that webapp changes are compiled immediately, and only changes need to be compiled
- symlink the webapp directory so that webapp changes don't need to redeployed (a plugin restart will pick up the changes immediately)
  - this means that the webapp can be debugged directly from the development directory (yay)

make reset command:
- disable the plugin
- enable the plugin (to pick up new changes in the server and webapp)
  - this stops the current instance of the plugin (to prevent 20+ zombie plugin processes)

make stop command:
- stop the webapp watcher

Quick Instructions (I'll write more detailed instructions when I merge this into the demo plugin... so that everyone can use this)
1. Instead of running `make deploy`, run `make debug`. 
    - It will set up the link between the mattermost server and the plugin's directory.
1. Run `make webapp-debug` to build the webapp with source maps, and start the npm watch process.
1. Whenever a change is made to the plugin's server component, run `make debug` to recompile and reload the plugin.
1. Whenever a change is made to the webapp, run `make reset` to stop and restart the plugin process on the server. Your changes will now appear (you may need to refresh the page)
1. To debug the server part of the plugin, run the script `./debug-plugin.sh` which will attach a delve headless process to the plugin. Attach your debugger of choice to localhost:2346 
1. To debug the webapp part of the plugin, attach the debugger to localhost:8065 (as you would normally debug the mattermost webapp).

## Links
[MM-15425](https://mattermost.atlassian.net/browse/MM-15425)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -168,7 +168,7 @@ endif
 
 # debug builds and deploys a debug version of the webapp, for only my architecture.
 .PHONY: debug
-debug:
+debug: stop
 	./build/bin/manifest apply
 ifneq ($(HAS_SERVER),)
 	mkdir -p server/dist
@@ -213,7 +213,7 @@ reset:
 ifneq ($(and $(MM_SERVICESETTINGS_SITEURL),$(MM_ADMIN_USERNAME),$(MM_ADMIN_PASSWORD),$(CURL)),)
 	@echo "\n\nRestarting plugin via API"
 	$(eval TOKEN := $(shell curl -i -X POST $(MM_SERVICESETTINGS_SITEURL)/api/v4/users/login -d '{"login_id": "$(MM_ADMIN_USERNAME)", "password": "$(MM_ADMIN_PASSWORD)"}' | grep Token | cut -f2 -d' '))
-	@curl -s -H "Authorization: Bearer $(TOKEN)" -X POST $(MM_SERVICESETTINGS_SITEURL)/api/v4/plugins/$(PLUGIN_ID)/disable && \
+	@curl -s -H "Authorization: Bearer $(TOKEN)" -X POST $(MM_SERVICESETTINGS_SITEURL)/api/v4/plugins/$(PLUGIN_ID)/disable > /dev/null && \
 		curl -s -H "Authorization: Bearer $(TOKEN)" -X POST $(MM_SERVICESETTINGS_SITEURL)/api/v4/plugins/$(PLUGIN_ID)/enable > /dev/null
 endif
 
--- a/Makefile
+++ b/Makefile
@@ -166,39 +166,42 @@ ifneq ($(HAS_WEBAPP),)
 endif
 	rm -fr build/bin/
 
-# debug builds and deploys a debug version of the webapp, for only my architecture.
+# server-debug builds and deploys a debug version of the plugin for your architecture.
+# Then resets the plugin to pick up the changes.
 .PHONY: debug
-debug: stop
-	./build/bin/manifest apply
-ifneq ($(HAS_SERVER),)
-	mkdir -p server/dist
-	cd server && env GOOS=darwin GOARCH=amd64 $(GO) build -gcflags "all=-N -l" -o dist/plugin-darwin-amd64
+debug: server-debug reset
 
-	# uncomment for your own architecture. This speeds up the cycle.
-	#cd server && env GOOS=linux GOARCH=amd64 $(GO) build -gcflags "all=-N -l" -o dist/plugin-linux-amd64;
-	#cd server && env GOOS=windows GOARCH=amd64 $(GO) build -gcflags "all=-N -l" -o dist/plugin-windows-amd64.exe;
-	cd server && cp -r templates dist/templates
+.PHONY: server-debug
+server-debug: server/.depensure
+
+	./build/bin/manifest apply
+	mkdir -p server/dist
+
+ifeq ($(OS),OSX)
+	cd server && env GOOS=darwin GOARCH=amd64 $(GO) build -gcflags "all=-N -l" -o dist/plugin-darwin-amd64;
+else ifeq ($(OS),Linux)
+	cd server && env GOOS=linux GOARCH=amd64 $(GO) build -gcflags "all=-N -l" -o dist/plugin-linux-amd64;
+else ifeq ($(OS),Windows_NT)
+	cd server && env GOOS=windows GOARCH=amd64 $(GO) build -gcflags "all=-N -l" -o dist/plugin-windows-amd64.exe;
+else
+	$(error Please set your OS system env variable to: 'OSX', 'Windows_NT', or 'Linux')
 endif
+
+	cd server && cp -r templates dist/templates
 	rm -rf dist/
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -164,3 +165,67 @@ ifneq ($(HAS_WEBAPP),)
...
 endif
 	rm -fr build/bin/
+
+# debug builds and deploys a debug version of the webapp, for only my architecture.
+.PHONY: debug
+debug: stop
+	./build/bin/manifest apply
+ifneq ($(HAS_SERVER),)
+	mkdir -p server/dist
+	cd server && env GOOS=darwin GOARCH=amd64 $(GO) build -gcflags "all=-N -l" -o dist/plugin-darwin-amd64
```
levb: we should use set this up for a local arch, not hardcoded; otherwise I'd suggest OS-specific targets, maybe `debug-osx`, etc.

Makefile
```diff
@@ -164,3 +165,67 @@ ifneq ($(HAS_WEBAPP),)
...
+	cp $(MANIFEST_FILE) dist/$(PLUGIN_ID)/
+ifneq ($(HAS_SERVER),)
+	mkdir -p dist/$(PLUGIN_ID)/server/dist
+	cp -r server/dist/* dist/$(PLUGIN_ID)/server/dist/
+endif
+
+	mkdir -p ../mattermost-server/plugins
+	cp -r dist/* ../mattermost-server/plugins/
+
+	# if there is a current plugin running, kill it
```
levb: This is a workaround for a bug, as opposed to a standard operating procedure. Put a comment, with a ticket reference, so we remove these lines later?

Makefile
cpoile: I'm now just requiring an OS env variable (I'll put it in the instructions). Personally I like that better than typing more every time I want to reload (3/5). Or do you prefer `make debug-OSX`?

Makefile
cpoile: I fixed it so we don't need to do this. When disabling/reenabling the plugin, the process is killed (as long as there isn't a delve process attached -- which it checks for now). The original bug hasn't been fixed, but fixing it won't affect this now.

Makefile
```diff
@@ -164,3 +165,96 @@ ifneq ($(HAS_WEBAPP),)
...
+
+.PHONY: server-debug
+server-debug: server/.depensure
+
+	./build/bin/manifest apply
+	mkdir -p server/dist
+
+ifeq ($(OS),OSX)
+	cd server && env GOOS=darwin GOARCH=amd64 $(GO) build -gcflags "all=-N -l" -o dist/plugin-darwin-amd64;
+else ifeq ($(OS),Linux)
```
crspeller: Can we make this automatic for our two most common dev platforms? We should be able to use `uname` to detect between those.

Makefile
```diff
@@ -164,3 +165,96 @@ ifneq ($(HAS_WEBAPP),)
...
+
+	# If we were debugging, we have to unattach the delve process or else we can't disable the plugin.
+	# NOTE: we are assuming the dlv was listening on port 2346, as in the debug-plugin.sh script.
+	@DELVE_PID=$(shell ps aux | grep "dlv attach.*2346" | grep -v "grep" | awk -F " " '{print $$2}') && \
+	if [ "$$DELVE_PID" -gt 0 ] > /dev/null 2>&1 ; then \
+		echo "Located existing delve process running with PID: $$DELVE_PID. Killing." ; \
+		kill -9 $$DELVE_PID ; \
+	fi
+
+ifneq ($(HTTP),)
```
crspeller: We dropped support for this non-standard tool in the base plugins. We should do that here too.

Makefile
```diff
@@ -164,3 +165,96 @@ ifneq ($(HAS_WEBAPP),)
...
+	@echo "\nRestarting plugin via API"
+		(TOKEN=`http --print h POST $(MM_SERVICESETTINGS_SITEURL)/api/v4/users/login login_id=$(MM_ADMIN_USERNAME) password=$(MM_ADMIN_PASSWORD) X-Requested-With:"XMLHttpRequest" | grep Token | cut -f2 -d' '` && \
+		  http --print b GET $(MM_SERVICESETTINGS_SITEURL)/api/v4/users/me Authorization:"Bearer $$TOKEN" && \
+			http --print b POST $(MM_SERVICESETTINGS_SITEURL)/api/v4/plugins/$(PLUGIN_ID)/disable Authorization:"Bearer $$TOKEN" && \
+		  http --print b POST $(MM_SERVICESETTINGS_SITEURL)/api/v4/plugins/$(PLUGIN_ID)/enable Authorization:"Bearer $$TOKEN" && \
+		  http --print b POST $(MM_SERVICESETTINGS_SITEURL)/api/v4/users/logout Authorization:"Bearer $$TOKEN" \
+	  )
+else ifneq ($(CURL),)
+	@echo "\nRestarting plugin via API"
+	$(eval TOKEN := $(shell curl -i -X POST $(MM_SERVICESETTINGS_SITEURL)/api/v4/users/login -d '{"login_id": "$(MM_ADMIN_USERNAME)", "password": "$(MM_ADMIN_PASSWORD)"}' | grep Token | cut -f2 -d' '))
```
crspeller: This doesn't work. Maybe it's just a linux thing. `curl -i -X POST $(MM_SERVICESETTINGS_SITEURL)/api/v4/users/login -d '{"login_id": "$(MM_ADMIN_USERNAME)", "password": "$(MM_ADMIN_PASSWORD)"}' | grep -o MMAUTHTOKEN=[0-9a-z]\* | cut -f2 -d'=' 2> /dev/null` Is what I modified mine to so it works.

Makefile
```diff
@@ -164,3 +165,96 @@ ifneq ($(HAS_WEBAPP),)
...
+	mkdir -p ./webapp
+	ln -nfs $(PWD)/webapp/dist ../mattermost-server/plugins/$(PLUGIN_ID)/webapp/dist
+	# start an npm watch
+	cd webapp && $(NPM) run run &
+endif
+
+	@echo "\n\n*** After the frontend is compiled, run 'make reset' to reset the plugin. Run reset every time a change is made to force the server to serve the chages in your webapp portion of the plugin.\n\n"
+
+# Reset the plugin
+.PHONY: reset
```
crspeller: There is a bunch of duplication here that should be resolved.

Makefile
cpoile: Good call, I forgot about uname.

Makefile
cpoile: Great, thanks. That seems to work on OSX as well, so we'll use that.

Makefile
cpoile: Great, thanks. That seems to work on OSX as well, so we'll use that.

PR #45, opened 2019-05-06
Fixed /jira subcommand parsing consistency

levb:
Cleaned up `/jira` sub-command parsing

Ticket: https://mattermost.atlassian.net/browse/MM-15335


crspeller:
@levb Considered using https://github.com/spf13/cobra ? That's what MM server uses.

levb:
@crspeller sure thing. IMO we should consider extracting some of this functionality into p.Helpers, and can then choose whatever implementation we want, right? Separate ticket?

crspeller:
@levb Certainly a separate ticket for moving stuff out to p.Helpers. But you might consider doing the cobra migration for this ticket. Your call.

cpoile:
@crspeller isn't Cobra for CLI, and this is for slash command parsing?

crspeller:
@cpoile Aren't slash commands like a CLI?

It have used cobra for this before. An example is matterbuild here: https://github.com/mattermost/matterbuild/blob/master/server/server.go

cpoile:
@crspeller that's awesome, didn't realize.. :)

server/command.go
```diff
@@ -21,70 +21,68 @@ const helpText = "###### Mattermost Jira Plugin - Slash Command Help\n" +
...
+		"connect":             executeConnect,
+		"disconnect":          executeDisconnect,
+	},
+	defaultHandler: commandHelp,
+}
+
+func (ch CommandHandler) Handle(p *Plugin, c *plugin.Context, args ...string) *model.CommandResponse {
+	for n := len(args); n > 0; n-- {
+		h := ch.handlers[strings.Join(args[:n], "/")]
+		if h != nil {
```
cpoile: nit: more idiomatic way:
```
if h, ok := ch.handlers[strings.Join(args[:n], "/")]; ok {
   return h(p, c, args[n:]...)
}
```
But 0/5 personally (because there's no chance you'll have a nil value instead of a missing entry).

levb: I disagree. In this case there is no semantic difference between "no entry" and a `nil` entry, so checking the value is the _right way_.

cpoile: Sounds good.

hanzei:
I didn't know a `SetArgs` method exists in Cobra. I was looking for a smarter way to parse slash commands for a long time. Locking forward to seeing this as a helper.

levb:
Will consider COBRA later, as a separate PR

PR #46, opened 2019-05-06
MM-15101 Adding /jira transition

crspeller:
https://mattermost.atlassian.net/browse/MM-15101

- Ticket didn't specify how exactly to match the text entered with the transition. Method implemented is a case-insensitive `strings.Contains()` on the transition target state name.

### DIFF
--- a/server/command.go
+++ b/server/command.go
@@ -6,7 +6,6 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/andygrunwald/go-jira"
 	"github.com/mattermost/mattermost-server/model"
 	"github.com/mattermost/mattermost-server/plugin"
 )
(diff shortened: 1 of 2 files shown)


server/command.go
```diff
@@ -250,6 +254,64 @@ func executeInstanceDelete(p *Plugin, c *plugin.Context, args ...string) *model.
 	return executeInstanceSelect(p, c, "1")
 }
 
+func (p *Plugin) transitionJiraIssue(mmUserId, issueKey, toState string) error {
```
levb: 1/5 `issue.go`? (not a blocking issue though)

crspeller: Oh ya, I totally meant to do that. Fixed.
Unrelated, we really need to separate our concerns api vs app logic like transitionJiraIssue.

PR #47, opened 2019-05-06
MM-15400 Added some protections to /installed link

levb:
- Now require the sysadmin to enter the Jira URL for cloud
- Match the incoming /installed with the JiraURL
- Only allow /installed to be invoked once, with a 15 min expiration

https://mattermost.atlassian.net/browse/MM-15400


server/command.go
```diff
@@ -173,8 +173,14 @@ func executeInstanceAdd(p *Plugin, c *plugin.Context, args ...string) *model.Com
...
 	case JIRATypeCloud:
+		// Create an "uninitialized" instance of Jira Cloud that will
+		// receive the /installed callback
+		err := p.StoreNewCloudInstance(jiraURL)
+		if err != nil {
+			return responsef(err.Error())
+		}
 		// TODO the exact group membership in Jira?
-		return responsef(`As an admin, upload an application from %s/%s. The link can be found in **Jira Settings > Applications > Manage**`,
+		return responsef(`As an admin, upload an application from %s/%s (expires in 15 minutes). The link can be found in **Jira Settings > Applications > Manage**`,
```
jasonblais: Why 15 minutes? (instead of shorter/longer time)

levb: No particular reason, seems sufficient. Would you like me to change this or make it configurable?

jasonblais: 15 minutes feels a bit short. I'd make it as long as is reasonable security-wise.

If it's 15 minutes, someone starting to set up a Jira integration before their meeting, holds their meeting, then continues after their meeting will result in an error they may not immediately understand why. It's not forgiving for the admin.

server/atlassian_connect.go
```diff
@@ -50,6 +51,26 @@ func httpACInstalled(p *Plugin, w http.ResponseWriter, r *http.Request) (int, er
...
+	// Only allow this operation once, a JIRA instance must already exist
+	// for asc.BaseURL but its EventType would be empty.
+	ji, err := p.LoadJIRAInstance(asc.BaseURL)
+	if err != nil {
+		return http.StatusInternalServerError,
+			errors.WithMessage(err, "failed to load instance "+asc.BaseURL)
+	}
+	if ji == nil {
+		return http.StatusNotFound,
+			errors.New(fmt.Sprintf("Jira instance %s must first be added to Mattermost", asc.BaseURL))
```
cpoile: alternative:
`fmt.Errorf("Jira instance %s must first be added to Mattermost", asc.BaseURL)`

server/atlassian_connect.go
```diff
@@ -50,6 +51,26 @@ func httpACInstalled(p *Plugin, w http.ResponseWriter, r *http.Request) (int, er
...
+	}
+	jci, ok := ji.(*jiraCloudInstance)
+	if !ok {
+		return http.StatusBadRequest, errors.New("Must be a JIRA Cloud instance, is " + ji.GetType())
+	}
+	if jci.EventType == "installed" {
+		return http.StatusUnauthorized,
+			errors.New(fmt.Sprintf("Jira instance %s is already installed", asc.BaseURL))
```
cpoile: same

server/atlassian_connect.go
```diff
@@ -50,6 +51,26 @@ func httpACInstalled(p *Plugin, w http.ResponseWriter, r *http.Request) (int, er
...
+	}
+	if ji == nil {
+		return http.StatusNotFound,
+			errors.New(fmt.Sprintf("Jira instance %s must first be added to Mattermost", asc.BaseURL))
+	}
+	jci, ok := ji.(*jiraCloudInstance)
+	if !ok {
+		return http.StatusBadRequest, errors.New("Must be a JIRA Cloud instance, is " + ji.GetType())
+	}
+	if jci.EventType == "installed" {
```
crspeller: This just seems like a strange thing to check here. Maybe something dedicated to this purpose instead of storing a blank instance?

server/command.go
levb: @crspeller do you have a suggestion on what is "as long as is reasonable security-wise"?

server/atlassian_connect.go
levb: I converted (almost) everything to `errors.New` and `errors.Errorf` for consistency. We should probably have a separate conversation on tradeoffs and the best practices of using simple errors (`fmt.Errorf`, `errors.WithMessage`) vs those embedding the call stacks (`errors.New/Errorf`, `errors.Wrap`). I am 0/5 just would like consistency.

server/atlassian_connect.go
levb: Now an `Installed` flag in jci

server/atlassian_connect.go
```diff
@@ -50,9 +51,28 @@ func httpACInstalled(p *Plugin, w http.ResponseWriter, r *http.Request) (int, er
...
+	if ji == nil {
+		return http.StatusNotFound,
+			errors.New(fmt.Sprintf("Jira instance %s must first be added to Mattermost", asc.BaseURL))
+	}
+	jci, ok := ji.(*jiraCloudInstance)
+	if !ok {
+		return http.StatusBadRequest, errors.New("Must be a JIRA Cloud instance, is " + ji.GetType())
+	}
+	if jci.Installed {
+		return http.StatusUnauthorized,
```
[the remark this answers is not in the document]
levb: Right, 401 is totally wrong here (not an authentication issue). But I think the right response is a 403 (forbidden). https://tools.ietf.org/html/rfc7231#section-6.5.3, will change

levb:
> ...work with the system console configuration...
@crspeller this is now between @jasonblais and @cpoile, they've been talking about having a `/jira instructions` command that would display the proper steps, with dynamically generated contents.

PR #49, opened 2019-05-08
Proposed update to response after a successful /jira transition

jasonblais:
#### Context

When "/jira transition" results in a successful jira issue transition, we post back an ephemeral message `Transition completed.`

image

In the ticket (https://mattermost.atlassian.net/browse/MM-15101), the proposal had been to use an ephemeral message with the UI matching a notification that is sent to a Mattermost channel when the state of an issue is changed.

An example of that is shown below.

image

I'm not sure if the UI was a bigger effort, and hence a simpler message was used. If so, this is okay because of our tight timeline.

#### Proposed Change

However, propose we update the response to

```
<issue-key> transition to <state>
```

to confirm for the user which issue was moved to which state.

### DIFF
--- a/server/command.go
+++ b/server/command.go
@@ -264,7 +264,7 @@ func executeTransition(p *Plugin, c *plugin.Context, args []string, userId strin
 		return responsef("%v", err)
 	}
 
-	return responsef("Transition completed.")
+	return responsef("issueKey transitioned to toState.")
 }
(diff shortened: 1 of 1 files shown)


jasonblais:
hah, yea..I was hoping it was this simple :P 

Will create a ticket

PR #50, opened 2019-05-08
Fixed creating server/dist/templates/templates on a 2nd make

levb:


levb:
@cpoile there are a couple other `cp -r`'s there to change then?

cpoile:
@levb Yah, `-r` is all over the place in all the makefiles (but sometimes it's `-R`, even in the same target. So I'm 0/5 on whether to even bother making this change (it doesn't seem to affect any of our devs). I probably shouldn't even have mentioned it.

hanzei:
Good catch on `-R` vs. `-r`. :+1: for consistency. Maybe make a small HW out of this?

crspeller:
@hanzei Made one: https://mattermost.atlassian.net/browse/MM-15496

PR #51, opened 2019-05-08
[MM-15411] /jira create command

cpoile:
### Summary

Add “/jira create” command

1) “/jira create” opens the same modal as when user selects "Create Jira Issue" from the post menu.
2) “/jira create [some text here]” opens the same modal, but with [some text here] copied for description.

In both cases, if user doesn't have permissions for the action, return ephemeral message "You do not have the appropriate permissions to perform this action. Please contact your Jira administrator.

### Links
[MM-15411](https://mattermost.atlassian.net/browse/MM-15411)
requires: https://github.com/mattermost/mattermost-webapp/pull/2526  ticket: [MM-15531](https://mattermost.atlassian.net/browse/MM-15531)


server/issue.go
```diff
@@ -25,8 +25,10 @@ func httpAPICreateIssue(ji Instance, w http.ResponseWriter, r *http.Request) (in
 	api := ji.GetPlugin().API
 
 	create := &struct {
-		PostId string           `json:"post_id"`
-		Fields jira.IssueFields `json:"fields"`
+		PostId         string           `json:"post_id"`
+		AttachedToPost bool             `json:"attached_to_post"`
```
levb: nit/naming: suggest `s/AttachedToPost/FromPost/`

levb: Also, shouldn't `PostId != ""` (and subsequently `post != nil` be enough? Do we really need a separate flag?

server/issue.go
```diff
@@ -49,42 +51,58 @@ func httpAPICreateIssue(ji Instance, w http.ResponseWriter, r *http.Request) (in
...
+		postId = create.PostId
+	} else {
+		channelId = create.ChannelId
+		postId = ""
+	}
+
 	if err != nil {
 		return http.StatusInternalServerError,
-			errors.WithMessage(err, "failed to create the issue, postId: "+create.PostId)
+			errors.WithMessage(err, "failed to create the issue, postId (if attached to a post): "+postId)
```
levb: nit: 1/5 I'd drop ` (if attached to a post)` from the error, `postId: nil` is intention-revealing enough.

server/issue.go
```diff
@@ -49,42 +51,58 @@ func httpAPICreateIssue(ji Instance, w http.ResponseWriter, r *http.Request) (in
...
 	})
+
+	// For now, if we are not attaching to a post, just postId blank (this will only affect the error message)
+	var channelId, postId string
+	if create.AttachedToPost {
+		channelId = post.ChannelId
+		postId = create.PostId
+	} else {
+		channelId = create.ChannelId
+		postId = ""
```
levb: nit: remove, redundant `postId = ""`

levb: 1/5 I'd do 
```go
postId := ""
channelId := create.ChannelId
if post != nil {
    postId = post.Id
    channelId = post.ChannelId
}

File: webapp/src/plugin.jsx [hunk 4]
```diff
@@ -18,6 +18,14 @@ export default class Plugin {
 
             registry.registerRootComponent(CreateIssueModal);
             registry.registerPostDropdownMenuComponent(CreateIssuePostMenuAction);
+
```
levb: 1/5 This seems to be the wrong place to do it

server/issue.go
cpoile: Good point. I usually like to make execution paths explicit, but `create.PostId != ""` should be clear enough.

server/issue.go
cpoile: much nicer, thanks

webapp/src/plugin.jsx
cpoile: I agree, it doesn't seem the right place. But:
1. I can't think of a more appropriate place
1. plugin-demo has a lot more than that in its initialize function. :)

server/issue.go
```diff
@@ -113,26 +128,26 @@ func httpAPICreateIssue(ji Instance, w http.ResponseWriter, r *http.Request) (in
...
-		RootId:    create.PostId,
+		ChannelId: channelId,
+		RootId:    postId,
 		UserId:    mattermostUserId,
 	}
 	_, appErr = api.CreatePost(reply)
 	if appErr != nil {
 		return http.StatusInternalServerError,
-			errors.WithMessage(appErr, "failed to create notification post "+create.PostId)
+			errors.WithMessage(appErr, "failed to create notification post, postId: "+postId)
```
crspeller: It's really the root post Id is it not? The one we are replying to?  Maybe add the channel ID here to make this error message more useful.

webapp/src/components/modals/create_issue/create_issue.jsx
```diff
@@ -51,6 +53,11 @@ export default class CreateIssueModal extends PureComponent {
             const fields = {...this.state.fields};
             fields.description = this.props.post.message;
             this.setState({fields}); //eslint-disable-line react/no-did-update-set-state
+        } else if (this.props.channelId && (this.props.channelId !== prevProps.channelId || this.props.description !== prevProps.description)) {
```
crspeller: This is getting unruly. Can we derive the description in the mapStateToProps function and just pass it correctly? Then we can simplify this. 

I would like to get rid of the derived state here. But I am not sure how.

webapp/src/components/modals/create_issue/create_issue.jsx
```diff
@@ -59,9 +66,14 @@ export default class CreateIssueModal extends PureComponent {
             e.preventDefault();
         }
 
+        const {post, channelId} = this.props;
+
+        const postId = (typeof post === 'undefined') ? '' : post.id;
```
crspeller: You don't need to do the typeof here because undefined is false.

webapp/src/plugin.jsx
```diff
@@ -18,6 +18,14 @@ export default class Plugin {
...
+            registry.registerWebSocketEventHandler(
+                'custom_' + PluginId + '_create_issue',
+                (payload) => {
```
crspeller: Not sure this is the right way to activate a modal client side. Will this not activate the create issue modal for all clients for that user?

I propose we cut this from jira 2.0 and leave it for 2.1. That way we can extend the plugin system to allow for client-side slash command interception and use that here.

webapp/src/actions/index.js
```diff
@@ -14,6 +14,14 @@ export const openCreateModal = (postId) => {
     };
 };
 
+export const openCreateModelEmpty = (description, channelId) => (dispatch) => dispatch({
```
crspeller: What do you mean by empty? You are passing in a description and channelId.

webapp/src/reducers/index.js
```diff
@@ -25,6 +26,20 @@ const createModalVisible = (state = false, action) => {
     }
 };
 
+const createModalEmpty = (state = '', action) => {
```
crspeller: This can probably be combined with the reducer below so we only have one thing to check for the create issue modal.

webapp/src/components/modals/create_issue/create_issue.jsx
cpoile: Like we chatted about over zoom, this will be split into a separate refactoring PR. 
I'd love to move all this derived state into props, and then just letting the component update itself when the state changes.

webapp/src/components/modals/create_issue/create_issue.jsx
cpoile: true true, I was just being really 'correct' (protecting against all the other falsey values... which will never happen here... :)

webapp/src/reducers/index.js
cpoile: Good point, done. Much nicer now. I also made it more idiomatic reduxy.

webapp/src/actions/index.js
cpoile: Good point. d'oh. Fixed. It's a little wordy now, but clear.

webapp/src/plugin.jsx
cpoile: So... it /does/ activate the modal in other clients. I guess that's a breaking problem? @levb @jasonblais? 
I can expand this ticket to include making a `handleSlashCommandWillBeSent` hook, if you want to include this in v2.0?

webapp/src/components/modals/create_issue/index.js
```diff
@@ -6,21 +6,23 @@ import {bindActionCreators} from 'redux';
 
 import {getPost} from 'mattermost-redux/selectors/entities/posts';
 
-import {closeCreateModal, createIssue, fetchJiraIssueMetadata} from 'actions';
-import {isCreateModalVisible, getCreateModalForPostId, getJiraIssueMetadata} from 'selectors';
+import {closeCreateModal, createIssue, fetchJiraIssueMetadata} from '../../../actions';
+import {isCreateModalVisible, getCreateModal, getJiraIssueMetadata} from '../../../selectors';
```
crspeller: Why the `../../../`? Webpack should handle resolving.

cpoile: it was giving me a warning in the file, and I want to clear those out. :)

crspeller: A warning? What warning?

cpoile: image

cpoile: I think it's because it's not an external module, so we should be specific about where we're getting it

crspeller: We use this kind of resolution all over the place in the webapp. I think you just need to fix your IDE.

cpoile: ok, sounds good

webapp/src/plugin.jsx
jasonblais: @crspeller @cpoile Given the issue at hand (opening a create issue modal on every client), agree: let's cut from v2.0, add the `handleSlashCommandWillBeSent` via [this ticket](https://mattermost.atlassian.net/browse/MM-15531), and then merge the command for v2.1.

cpoile:
@crspeller It's not giving me a `re-request review` for you. But it's ready for a re-review. Thanks!

levb:
@cpoile can you please rebase and retarget this to `master`? I gave it a try but there were some non-trivial conflicts.

cpoile:
@levb done. That was... non-trivial. :)

Issue #33, opened 2019-03-19
Secrets with a plus sign not recognized


icelander:
# Summary

Testing the Jira plugin with a `curl` is a common way of figuring out if it's working. However, if there is a `+` in the secret it is not recognized by the plugin.

# Environment

* Mattermost Server: 5.8.0

# Steps to Reproduce

1. Edit `config.json` so that the `PluginSettings.Plugins.jira.secret` value has a `+` in it. For example: `KBcCZm_JWQWM2WD9tr5IMo+fUo5zGJwS`
2. `curl` the URL:

```bash
curl -XPOST "https://mattermost.example.com/plugins/jira/webhook?secret=KBcCZm_JWQWM2WD9tr5IMo+fUo5zGJwS&team=a-team&channel=town-square
```

# Expected Behavior

`curl` command outputs this:

```
EOF
```

# Observed Behavior

`curl` command outputs this:

```
You must provide the configured secret
```

# Recommended Fixes

Scrub secrets of `+` character when generating them.

levb:
Right. We can (must?!)  URI-encode the secret.

BTW, your secret should be fine if you replace `+` with `%2B` in that URL

jasonblais:
Resolved by https://github.com/mattermost/mattermost-plugin-jira/pull/53, to be included in Jira v2.0 build next month.

PR #53, opened 2019-05-09
Fixed a bug in unescaping webhook secret

levb:
This PR fixes the issue that we'd receive the secret double-%-escaped. This is to unescape the secret repeatedly until there are no more valid %hh sequences left, comparing each version.

@jasonblais I am not sure there was a prior (1.0) ticket on this, couldn't find it. This may or may not be the same issue, but should be a general improvement anyway. I'd prefer that the secret was base64-encoded but I don't think it's better to stick with a backwards-compatible approach.

Also added `/jira webhook` command to see a URL custom-fit to the current channel.

### DIFF
--- a/server/webhook.go
+++ b/server/webhook.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"strings"
 
 	"github.com/andygrunwald/go-jira"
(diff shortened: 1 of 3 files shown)


jasonblais:
There wasn't a Jira ticket, it was reported here https://github.com/mattermost/mattermost-plugin-jira/issues/33

server/webhook.go
```diff
@@ -329,3 +335,24 @@ func (p *Plugin) notify(ji Instance, parsed *parsedJIRAWebhook, text string) {
...
+		return "", appErr
+	}
+
+	channel, appErr := p.API.GetChannel(channelId)
+	if appErr != nil {
+		return "", appErr
+	}
+
+	v := url.Values{}
+	secret, _ := url.QueryUnescape(cf.Secret)
```
cpoile: Just for my own understanding, why are you unescaping the cf.Secret? Is there any reason it would be escaped before storing in the config?

levb: That's actually a good question. I might have gone too far there - never really found the code that generated the value. I'll double-check.

PR #55, opened 2019-05-10
MM-15455 Some improvments to cloud connect flow.

crspeller:


server/instance_cloud.go
```diff
@@ -165,7 +165,7 @@ func (jci jiraCloudInstance) parseHTTPRequestJWT(r *http.Request) (*jwt.Token, s
 		// HMAC secret is a []byte
 		return []byte(jci.AtlassianSecurityContext.SharedSecret), nil
 	})
-	if err != nil {
+	if err != nil || !token.Valid {
```
crspeller: It will return an error if not valid, but just for sanity. (they check this in their docs)

server/user_cloud.go
```diff
@@ -32,10 +30,7 @@ func httpACUserRedirect(jci *jiraCloudInstance, w http.ResponseWriter, r *http.R
 		return http.StatusBadRequest, err
 	}
 
-	submitURL := path.Join(jci.Plugin.GetPluginURLPath(), routeACUserConnected)
-	if requireUserApproval {
-		submitURL = path.Join(jci.Plugin.GetPluginURLPath(), routeACUserConfirm)
-	}
+	submitURL := path.Join(jci.Plugin.GetPluginURLPath(), routeACUserConfirm)
```
crspeller: The confirmation is required to be secure. Otherwise you can trick people into clicking the link.

levb: @jasonblais is requesting the opposite :) Can you two please resolve it? (I'm 3/5 that security should win, and we should have the explicit warning/confirmation)

crspeller: Yup. Will chat with @jasonblais

jasonblais: Chatted with @crspeller, given the level of a security vulnerability if this button wasn't added, I'm good with this change. 

Approving PR.

levb:
@crspeller do you mind also removing the definition of requiresUserApproval? (nit, can wait)

crspeller:
@levb Not sure what you mean.

PR #56, opened 2019-05-13
[MM-14513] Implement attaching messages to Jira issues as comment

jfrerich:
**Summary**
Add ability to attach a MM message to a Jira Issue, via webapp

**Ticket Link**
The Jira Ticket describes UI change in detail.
https://mattermost.atlassian.net/browse/MM-14513

**Checklist**
[x]  Has UI changes


server/issue.go
```diff
@@ -158,6 +158,159 @@ func httpAPIGetCreateIssueMetadata(ji Instance, w http.ResponseWriter, r *http.R
...
+			} else {
+				comment.Fields.Body = permalinkMessage
+			}
+
+			comment.Fields.Body += post.Message
+		}
+	}
+
+	commentAdded, _, err := jiraClient.Issue.AddComment(comment.Fields.ID, &comment.Fields)
+	fmt.Printf("commentAdded = %+v\n", commentAdded)
```
crspeller: Debugging?

server/issue.go
```diff
@@ -158,6 +158,159 @@ func httpAPIGetCreateIssueMetadata(ji Instance, w http.ResponseWriter, r *http.R
...
+
+	if err != nil {
+		return http.StatusInternalServerError,
+			errors.WithMessage(err, "failed to attach the comment, postId: "+comment.PostId)
+	}
+
+	// Reply to the post with the issue link that was created
+	reply := &model.Post{
+		// TODO: Why is this not created.Self?
```
crspeller: TODO? Not sure what you mean.

webapp/src/components/modals/attach_issue/attach_issue.jsx
```diff
@@ -0,0 +1,241 @@
...
+                        disabled={false}
+                    />
+                    <Input
+                        label='Message Attached to Jira Issue'
+                        type='textarea'
+                        isDisabled={true}
+                        value={this.state.fields.description}
+                        disabled={false}
+                    />
+                    <JiraFields
```
crspeller: What are the fields for?

server/issue.go
```diff
@@ -158,6 +158,159 @@ func httpAPIGetCreateIssueMetadata(ji Instance, w http.ResponseWriter, r *http.R
...
+	if err != nil {
+		return http.StatusInternalServerError, err
+	}
+
+	jiraClient, err := ji.GetJIRAClient(jiraUser)
+	if err != nil {
+		return http.StatusInternalServerError, err
+	}
+
+	// TODO : redundant.  Maybe this entire fucntion is..
```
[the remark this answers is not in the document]
crspeller: We don't' event use this in the client code.

server/issue.go
levb: @crspeller I think this may be a leftover from my comment in CreateIssue. We compose the issue URL, rather than using what I think is a "self" reference provided by Jira. Wanted to test out and maybe switch over.

server/http.go
```diff
@@ -18,6 +18,8 @@ import (
 const (
 	routeAPICreateIssue            = "/api/v2/create-issue"
 	routeAPIGetCreateIssueMetadata = "/api/v2/get-create-issue-metadata"
+	routeAPIAttachIssue            = "/api/v2/attach-issue"
```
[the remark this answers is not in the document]
jfrerich: I agree with this proposal.  I should probably rename the files and directories too. What do you think? 

Move every `attach_issue` to `attach_comment_to_issue`

server/issue.go
```diff
@@ -158,6 +158,159 @@ func httpAPIGetCreateIssueMetadata(ji Instance, w http.ResponseWriter, r *http.R
...
+	}
+
+	jiraClient, err := ji.GetJIRAClient(jiraUser)
+	if err != nil {
+		return http.StatusInternalServerError, err
+	}
+
+	// Lets add a permalink to the post in the Jira Description
+	post, appErr := api.GetPost(comment.PostId)
+
```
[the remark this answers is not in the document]
jfrerich: Done.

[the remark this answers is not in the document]
jfrerich: Done

server/issue.go
```diff
@@ -158,6 +158,159 @@ func httpAPIGetCreateIssueMetadata(ji Instance, w http.ResponseWriter, r *http.R
...
+	if appErr != nil {
+		return http.StatusInternalServerError,
+			errors.WithMessage(appErr, "failed to load post "+comment.PostId)
+	}
+
+	mattermostUser, err := api.GetUser(post.UserId)
```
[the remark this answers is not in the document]
jfrerich: agreed and done.  This "user" is the user who posted the comment that is getting attached, not the user who is attaching. Thanks.

server/issue.go
jfrerich: yes.  Removed.

webapp/src/components/modals/attach_issue/attach_issue.jsx
jfrerich: These aren't getting used and I can remove them.  In `create issue`, this was used to populate issue types, from using the metadata, in a more programmatic way.  I just used two <input /> components to 1) allow user to input an issue Key (ie. EXT-20) and 2) show the comment message from MM.  (with disabled prop so cannot be edited)

server/issue.go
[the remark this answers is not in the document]
jfrerich: For further understanding:

The change is only a naming convention to save the error returned from the plugin api method and to specify that the return error is typed (*model.AppError)?  

Thanks for the catch and link to the explanation.

server/issue.go
```diff
@@ -158,6 +158,159 @@ func httpAPIGetCreateIssueMetadata(ji Instance, w http.ResponseWriter, r *http.R
...
+		// return http.StatusInternalServerError, err
+	}
+
+	if post == nil {
+		return http.StatusInternalServerError,
+			errors.New("failed to load post " + comment.PostId + ": not found")
+	}
+
+	if channel, _ := api.GetChannel(post.ChannelId); channel != nil {
+		if team, _ := api.GetTeam(channel.TeamId); team != nil {
```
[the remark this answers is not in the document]
jfrerich: Will wrap this into the next PR commit.

server/issue.go
```diff
@@ -158,6 +158,159 @@ func httpAPIGetCreateIssueMetadata(ji Instance, w http.ResponseWriter, r *http.R
...
+		// TODO : appropriate error
+		// return http.StatusInternalServerError, err
+	}
+
+	if post == nil {
+		return http.StatusInternalServerError,
+			errors.New("failed to load post " + comment.PostId + ": not found")
+	}
+
+	if channel, _ := api.GetChannel(post.ChannelId); channel != nil {
```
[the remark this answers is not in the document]
jfrerich: Agree and will wrap this into the next commit

server/http.go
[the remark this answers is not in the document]
jfrerich: New PR will address this issue.  Some of the function names are rather long, but there will no longer be confusion as to what the function, modal, or component does.

server/issue.go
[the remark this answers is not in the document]
levb: No, the issue is that when you wrap a non-interface nil pointer to an interface, it's not a nil value of that interface. I actually don't understand how this even worked before, but I didn't even bother since the change was worth it on stylistic grounds alone.

server/issue.go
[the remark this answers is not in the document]
jfrerich: I'm going to delete this function, but leave the httpAPIGetCreateIssueMetadata.

server/http.go
```diff
@@ -18,6 +18,8 @@ import (
...
+	routeAPIGetAttachIssueMetadata = "/api/v2/get-attach-issue-metadata"
```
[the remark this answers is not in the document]
jfrerich: Correct.  I removed this endpoint and httpAPIGetAttachIssueMetadata function.

server/issue.go
jfrerich: self returns a Jira Rest endpoint for the ticket, which provides quite a bit of information about the ticket, but no single property that contains the complete URL to the Jira issue.  Looks like we will have to construct the URL

levb: You can totally just kill the comment here for now

server/issue.go
```diff
@@ -181,6 +178,126 @@ func httpAPIGetCreateIssueMetadata(ji Instance, w http.ResponseWriter, r *http.R
...
+	if r.Method != http.MethodPost {
+		return http.StatusMethodNotAllowed,
+			errors.New("method " + r.Method + " is not allowed, must be POST")
+	}
+
+	api := ji.GetPlugin().API
+
+	comment := &struct {
+		PostId string       `json:"post_id"`
+		Fields jira.Comment `json:"fields"`
```
crspeller: Why do we decode into a jira.Comment here?

server/issue.go
```diff
@@ -181,6 +178,126 @@ func httpAPIGetCreateIssueMetadata(ji Instance, w http.ResponseWriter, r *http.R
...
+	permalinkMessage := fmt.Sprintf("*@%s attached a* [message|%s] *from @%s*\n", jiraUser.User.Name, permalink, commentUser.Username)
+
+	if len(comment.Fields.Body) > 0 {
+		comment.Fields.Body += fmt.Sprintf("\n%v", permalinkMessage)
+	} else {
+		comment.Fields.Body = permalinkMessage
+	}
+	comment.Fields.Body += post.Message
+
+	commentAdded, _, err := jiraClient.Issue.AddComment(comment.Fields.ID, &comment.Fields)
```
crspeller: The first argument to `AddComment` is the issue Id. It's a bit confusing to see it stored in `Fields.ID`

webapp/src/components/modals/attach_comment_to_issue/attach_comment_to_issue.jsx
```diff
@@ -0,0 +1,241 @@
...
+                        disabled={false}
+                    />
+                    <Input
+                        label='Message Attached to Jira Issue'
+                        type='textarea'
+                        isDisabled={true}
+                        value={this.state.fields.description}
+                        disabled={false}
+                    />
+                    <JiraFields
```
crspeller: What is this used for? Since we don't pick and `issueType` it won't work anyway.

webapp/src/components/modals/attach_comment_to_issue/attach_comment_to_issue.jsx
```diff
@@ -0,0 +1,241 @@
...
+            projectKey,
+            fields,
+        });
+    }
+
+    handleIssueKeyChange = (id, value) => {
+        const fields = {...this.state.fields};
+        const issueKey = value;
+
+        fields.ID = value;
```
crspeller: Why set this on fields.id and not just issueKey?

server/issue.go
jfrerich: Agree and am making changes to the fields of the struct and variable naming so make this clearer.

server/issue.go
jfrerich: After coming back to this after a couple of days I realize we only need the post_id and IssueKey from the response and do not need to decode into a jira.comment.  I am making changes that

webapp/src/components/modals/attach_comment_to_issue/attach_comment_to_issue.jsx
jfrerich: Agree and clearing this up in the next release.  Thanks @crspeller

webapp/src/components/modals/attach_comment_to_issue/attach_comment_to_issue.jsx
jfrerich: Correct.  I'm pulling this out in the next release. Thanks.

jfrerich:
performed `git merge origin/jira2`
fixed conflict with server/issue.go
ready to merge

PR #60, opened 2019-05-16
[MM-15100] Add /jira assign <issue-key> <assignee> command

jfrerich:
*Summary*
This PR adds the ability to change the assignee of an issue via the MM `jira assign` slash command.

The go-jira module was also updated to the latest version. The JIRA REST API uses AccountID and version 1.6 of go-jira did not have this field in the jira.User struct fields.

*Ticket Link* 
Please view the following ticket for list of use cases and error handling.
https://mattermost.atlassian.net/browse/MM-15100


server/command.go
```diff
@@ -16,6 +16,7 @@ const helpText = "###### Mattermost Jira Plugin - Slash Command Help\n" +
 	"\nFor system administrators:\n" +
 	"* `/jira install cloud <URL>` - connect Mattermost to a cloud Jira instance located at <URL>\n" +
 	"* `/jira install server <URL>` - connect Mattermost to a server Jira instance located at <URL>\n" +
+	"* `/jira assign <issuekey> <assignee>` - Changes the assignee of a Jira issue.\n" +
```
cpoile: Is it restricted to system administrators? If so, you need to check in executeAssign.

server/issue.go
```diff
@@ -193,6 +193,78 @@ func httpAPIGetCreateIssueMetadata(ji Instance, w http.ResponseWriter, r *http.R
...
+	url := fmt.Sprintf("rest/api/2/user/assignable/search?issueKey=%s&query=%s", issueKey, jiraAssignee)
+	req, err := jiraClient.NewRequest("GET", url, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	res, err := jiraClient.Do(req, nil)
+	if err != nil {
+		if res.Response.StatusCode == 401 {
+			errorMsg := fmt.Sprintf("You do not have the appropriate permissions to perform this action. Please contact your Jira administrator.")
```
levb: nit: no need for the var, nor Sprintf, just return responsef("....", nil)?

server/issue.go
```diff
@@ -193,6 +193,78 @@ func httpAPIGetCreateIssueMetadata(ji Instance, w http.ResponseWriter, r *http.R
...
+			return responsef(errorMsg), nil
+		}
+		return nil, err
+	}
+
+	body, err := ioutil.ReadAll(res.Body)
```
levb: 1/5 use a NewDecoder and decode directly from the body. In other places we need the raw JSON for one reason or another, but not here?

server/command.go
```diff
@@ -239,6 +240,23 @@ If you see an option to create a Jira issue, you're all set! If not, refer to ou
 	return responsef(addResponseFormat, ji.GetURL(), p.GetSiteURL(), ji.GetMattermostKey(), pkey)
 }
 
+func executeAssign(p *Plugin, c *plugin.Context, header *model.CommandArgs, args ...string) *model.CommandResponse {
+
+	if len(args) <= 1 {
```
cpoile: Unless there can be 3 or more args, I would suggest `len(args) != 2`
Below you're joining them with spaces -- if that's intentional, you have to send over the wire with URL encoding

server/issue.go
```diff
@@ -193,6 +193,78 @@ func httpAPIGetCreateIssueMetadata(ji Instance, w http.ResponseWriter, r *http.R
...
+	var data []*jira.User
+	if err := json.Unmarshal([]byte(body), &data); err != nil {
+		return nil, err
+	}
+
+	// handle number of returned jira users
+	if len(data) == 0 {
+		errorMsg := fmt.Sprintf("We couldn't find the assignee. Please use a Jira member and try again.")
+		return nil, fmt.Errorf(errorMsg)
+	} else if len(data) > 1 {
```
levb: nit: 2/5 no need for an else, use a blank line and a new `if len(data) > 1 {`

server/issue.go
```diff
@@ -193,6 +193,78 @@ func httpAPIGetCreateIssueMetadata(ji Instance, w http.ResponseWriter, r *http.R
 	return http.StatusOK, nil
 }
 
+func (p *Plugin) assignJiraIssue(mmUserId, issueKey, jiraAssignee string) (*model.CommandResponse, error) {
```
levb: Why return both a response and an err?

server/issue.go
cpoile: should have 
`defer res.Body.Close()` before the ReadAll (to prevent resource leaks)

server/issue.go
```diff
@@ -193,6 +193,78 @@ func httpAPIGetCreateIssueMetadata(ji Instance, w http.ResponseWriter, r *http.R
...
+		return nil, err
+	}
+
+	// handle number of returned jira users
+	if len(data) == 0 {
+		errorMsg := fmt.Sprintf("We couldn't find the assignee. Please use a Jira member and try again.")
+		return nil, fmt.Errorf(errorMsg)
+	} else if len(data) > 1 {
+		errorMsg := fmt.Sprintf("Your <assignee>, `%s`, matches %d users.  Please make your user request unique.", jiraAssignee, len(data))
+		return nil, fmt.Errorf(errorMsg)
```
cpoile: as Lev said. It keeps if-elses to a minimum

server/issue.go
cpoile: My fault -- I suggested that, since you need to provide the positive response... But I see now that you can provide the positive response in the command handler, and you only need to return errors from this function.

server/issue.go
jfrerich: Correct. Thanks for the catch

server/command.go
jfrerich: Good question.   @jasonblais?  I think it should be left open.  The slash command will check if the user has jira permissions on the issue.

server/command.go
jfrerich: That was not intentional.  using your suggestions and removing join from below.

server/issue.go
jfrerich: Good catch.  using your suggestion.

server/issue.go
jfrerich: Done

server/issue.go
jfrerich: done.  Thanks

server/issue.go
jfrerich: @cpoile How do I provide the non-error type response if I only return an error type and not the *model.CommandResponse?

server/command.go
jasonblais: No need to restrict to system administrators. It should be dependent on Jira permissions, similar to `/jira transition`

jfrerich: Thanks @jasonblais.  Will leave as is.  Thanks for the question too, @cpoile

server/issue.go
cpoile: @jfrerich well, you know a return of nil will be successful, so something like:
```
err = assignJiraIssue(...)
if err == nil {
  // do your error handling
}
return responsef(return the successful feedback here)
```

levb: model.CommandResponse does not provide for a separate stdin vs stderr, 0/5 do the simplest thing? (return a `string`? now where would that fit in an OO book?)

jfrerich: @cpoile assignJiraIssue(...) returns the following if successful.  I return information gained from the function.  That is why I also have the *model.CommandResponse returned.  Otherwise, I would have just checked err == nil, like you said above.  Does that explanation make sense?

`	msg := fmt.Sprintf("`%s` assigned to Jira issue [%s](%s)", assignee.DisplayName, issueKey, permalink)
	return responsef(msg), nil

`

server/command.go
jasonblais:
```suggestion
	"* `/jira assign <issuekey> <assignee>` - Change the assignee of a Jira issue.\n" +
```

jasonblais: Changes > Change - for consistency with other commands (e.g. `connect`/`disconnect`)

server/issue.go
cpoile: @jfrerich gotcha, I misread assignee as something you were passing in. :) I think returning a (string, error) would probably work then, and put the string in a responsef in the caller. Keep the function as pure as possible (it doesn't need to know its result will be sent in a response).

jfrerich: Thanks. I'm settling on (string, error) and will have the caller wrap the string in a responsef().

server/issue.go
```diff
@@ -193,6 +193,78 @@ func httpAPIGetCreateIssueMetadata(ji Instance, w http.ResponseWriter, r *http.R
...
+
+	// check for valid issue key
+	_, _, err = jiraClient.Issue.Get(issueKey, nil)
+	if err != nil {
+		errorMsg := fmt.Sprintf("We couldn't find the issue key `%s`.  Please confirm the issue key and try again.", issueKey)
+		return responsef(errorMsg), nil
+	}
+
+	// Get list of assignable assignees
+	url := fmt.Sprintf("rest/api/2/user/assignable/search?issueKey=%s&query=%s", issueKey, jiraAssignee)
```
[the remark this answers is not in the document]
jfrerich: updated to `/api/3` and retested use cases.  I agree with adding a GetRESTURL() method on Instance.

server/issue.go
jfrerich: I am 0/5 and will make the change.  Thanks.

server/command.go
jfrerich: Got it.  Thanks J

jfrerich: @jasonblais Do you want me to do the same for /jira transistion while I'm in this area of the code?

jasonblais: @jfrerich go for it 👍

server/issue.go
```diff
@@ -193,6 +193,75 @@ func httpAPIGetCreateIssueMetadata(ji Instance, w http.ResponseWriter, r *http.R
...
+			return "You do not have the appropriate permissions to perform this action. Please contact your Jira administrator.", nil
+		}
+		return "", err
+	}
+
+	var data []*jira.User
+	if err := json.NewDecoder(res.Body).Decode(&data); err != nil {
+		return "", errors.WithMessage(err, "failed to decode incoming request")
+	}
+	defer res.Body.Close()
```
cpoile: the defer should go before you read from res.Body (that way if something happens and you error out, the body will still be closed)

jfrerich: Thanks for the added explanation to the request.  Adding and will commit asap

server/issue.go
```diff
@@ -193,6 +193,75 @@ func httpAPIGetCreateIssueMetadata(ji Instance, w http.ResponseWriter, r *http.R
...
+	res, err := jiraClient.Do(req, nil)
+	if err != nil {
+		if res.Response.StatusCode == 401 {
+			return "You do not have the appropriate permissions to perform this action. Please contact your Jira administrator.", nil
+		}
+		return "", err
+	}
+
+	var data []*jira.User
+	defer res.Body.Close()
```
levb: res.Body.Close() must be called even in the case of an error, if `resp` != nil

server/issue.go
```diff
@@ -193,6 +193,75 @@ func httpAPIGetCreateIssueMetadata(ji Instance, w http.ResponseWriter, r *http.R
...
+	}
+
+	// Get list of assignable assignees
+	url := fmt.Sprintf("rest/api/3/user/assignable/search?issueKey=%s&query=%s", issueKey, assignee)
+	req, err := jiraClient.NewRequest("GET", url, nil)
+	if err != nil {
+		return "", err
+	}
+
+	res, err := jiraClient.Do(req, nil)
```
levb: nit: `resp` is the usual short name for the response.

server/issue.go
```diff
@@ -193,6 +193,75 @@ func httpAPIGetCreateIssueMetadata(ji Instance, w http.ResponseWriter, r *http.R
...
+	if err != nil {
+		if res.Response.StatusCode == 401 {
+			return "You do not have the appropriate permissions to perform this action. Please contact your Jira administrator.", nil
+		}
+		return "", err
+	}
+
+	var data []*jira.User
```
levb: nit/naming: suggest `s/data/found/` as more intention-revealing (alternatively, `[jira]users`)

server/issue.go
```diff
@@ -193,6 +193,75 @@ func httpAPIGetCreateIssueMetadata(ji Instance, w http.ResponseWriter, r *http.R
...
+		return "", fmt.Errorf(errorMsg)
+	}
+
+	if len(data) > 1 {
+		errorMsg := fmt.Sprintf("Your <assignee>, `%s`, matches %d users.  Please make your user request unique.", assignee, len(data))
+		return "", fmt.Errorf(errorMsg)
+	}
+
+	// User is array of one object
+	User := data[0]
```
levb: style: local variables are not to start with a Capital letter. suggest `jiraUser`

server/issue.go
jfrerich: @levb see proposed change shown below.  This code would be placed immediately after the request on line 220.  Is that correct?  We don't have to close the body if resp == nil?
```
	if res != nil {
		defer res.Body.Close()
	}
```

server/issue.go
jfrerich: I've seen res, resp, and just r.  Will use resp for MM code in the future.

server/issue.go
jfrerich: I like jiraUsers as the name of the list.  Incorporating suggestion. Thanks.

server/issue.go
jfrerich: jiraUser is already defined and occupied earlier in the function.  Will how about `user` with lower case?

server/issue.go
cpoile: I suggest putting it as your first line after checking the error from the request (line 227 above):
```go
if err != nil {
    // do stuff
}
defer resp.Body.Close()
```

cpoile: Alternatively, pass in `&data` to `jiraClient.Do` and it will do all this for you.

jfrerich: Nice catch.  I pulled this section of code from hackathon code and didn't realize the optional interface input to `jiraClient.Do` 

That will handle this code for me and reduce the custom code nicely. Thanks @cpoile

jfrerich:
Thanks for the feedback @cpoile and @levb.

jfrerich:
@levb, I can add that functionality. 

Additionally, we could be even more creative with any of the following ideas:

1. open a modal with a select field prepopulated with the user's previous search and use autocomplete to find the correct match
2. show matches from search and allow the user to click the correct name to complete the assign task

jasonblais:
> show matches from search and allow the user to click the correct name to complete the assign task

That sounds like a great idea. Perhaps included within an ephemeral message.

I'd also be happy to smoke test this feature prior to merge, let me know.

jfrerich:
@jasonblais After some discussion with @levb we limited the scope of this ticket to the following use cases:

**Require minimum of 3 characters in assignee value**
image

**Require show bullet list of matched users**
image

**Show first `x` users is # users > 10.  (used max of 3 for testing)**
image

**issue key not found** 
image

**User successfully assigned** 
image

**help text for `/jira assign`
image

server/issue.go
```diff
@@ -540,6 +540,91 @@ func (p *Plugin) getIssueAsSlackAttachment(ji Instance, jiraUser JIRAUser, issue
...
+	// handle number of returned jira users
+	if len(jiraUsers) == 0 {
+		errorMsg := fmt.Sprintf("We couldn't find the assignee. Please use a Jira member and try again.")
+		return "", fmt.Errorf(errorMsg)
+	}
+
+	maxDisplayedUsers := 10
+	if len(jiraUsers) > 1 {
+
+		errorMsg := fmt.Sprintf("`%s` matches %d users.  Please make your user request unique.\n", assignee, len(jiraUsers))
```
jasonblais:
```suggestion
		errorMsg := fmt.Sprintf("`%s` matches %d users.  Please specify a unique assignee.\n", assignee, len(jiraUsers))
```

jfrerich: @jasonblais, yes the three-character minimum was to avoid returning too many search results and limit Jira from auto-selecting a user that wasn't really intended.  

For example:
I used `/jira assign EXT-20 a`.  I was expecting to see use, 'jason', `asaad`, and anyone else with an `a`.  Instead, the Jira assigned directly to `Asaad`.

server/issue.go
```diff
@@ -540,6 +540,91 @@ func (p *Plugin) getIssueAsSlackAttachment(ji Instance, jiraUser JIRAUser, issue
...
+
+	// required minimum of three letters in assignee value
+	minChars := 3
+	if len(assignee) < minChars {
+		errorMsg := fmt.Sprintf("`%s` contains less than %v characters.", assignee, minChars)
+		return errorMsg, nil
+	}
+
+	// check for valid issue key
+	_, _, err = jiraClient.Issue.Get(issueKey, nil)
```
levb: Can this check be skipped? I'd imagine the subsequent request(s) would fail on a non-existing ticket anyway...

jfrerich: @levb, this does get checked later it the code if a non-existent issue key is provided.  Unfortunately, the message returned by `jiraClient.Issue.UpdateAssignee()` is generic.  In the image below, you can see the two cases.  The first is where I handle the err and the second is the return value from Jira. 

image

PR #66, opened 2019-05-21
[MM-15536/MM-15832] Only show one "Connect to Jira" menu option.

jfrerich:
**Summary**
This PR addresses the issue when a Jira instance exists, but the user has not yet attached their Jira account to MM.  The Post Menu pulldown previously showed "Connect to Jira" for each action, where each available action is `Create Jira Issue` and `Attach Jira Issue`.  There is not a need to show this menu option twice. 

This ticket resolves the issue by only showing a single "Connect to Jira" menu option.

Note to reviewers:  A more complete solution would be to encapsulate the `create jira` and `attach jira` components with a parent component.  For simplicity and time, this solution uses the `create issue` post modal as the master to serve this function and removes the connecting feature from the `attach jira issue` modal.

**Ticket Link**
https://mattermost.atlassian.net/browse/MM-15536
https://mattermost.atlassian.net/browse/MM-15832


webapp/src/components/post_menu_actions/attach_comment_to_issue/attach_comment_to_issue.jsx
```diff
@@ -56,16 +56,6 @@ export default class AttachCommentToIssuePostMenuAction extends PureComponent {
                     {this.getLocalizedTitle()}
                 </button>
             );
-        } else {
```
levb: 1/5 consolidate lines 43-48, smething like
```
    const conn = this.props.connected || {};
    if (this.props.isSystemMessage || !conn.connected) {
            return null;
    }

    let content;
    ...
```

jfrerich: I like this idea as it removes the extra if statement.  I think we will eventually refactor these post menu options components and end up doing the same for the create issue post menu modal.  Submitting change soon.

webapp/src/components/post_menu_actions/attach_comment_to_issue/attach_comment_to_issue.jsx
```diff
@@ -40,33 +40,20 @@ export default class AttachCommentToIssuePostMenuAction extends PureComponent {
     }
 
     render() {
-        if (this.props.isSystemMessage) {
+        const conn = this.props.connected || {};
```
crspeller: Why is this an object? Looking at the reducer for `connected` it looks like we save a bool. But the mapStateToProps doesn't seem to use that, it seems to just grab any state. Can we fix this strange behavior?

jfrerich: Hey @crspeller.  I add an isConnected selector in the latest commit.  Please review to see if that fix is what you had in mind.  Thanks

jfrerich:
@jasonblais FYI, the latest commit will add the jira icon to the "Connect to Jira" post menu item

PR #71, opened 2019-05-23
MM-15535 - UI updates for Jira 2

asaadmahmood:
Some minor changes.


webapp/src/components/input.jsx
```diff
@@ -77,6 +77,7 @@ export default class Input extends PureComponent {
                     id={this.props.id}
                     className='form-control'
                     rows='5'
+                    readOnly
```
crspeller: Needs to be:
```suggestion
                    readOnly={true}
```
to make the linter happy.

PR #72, opened 2019-05-23
MM-15439 Restores 1.0 compatibility for webhook filtering

levb:
- By default, only the events supported by 1.0 are shown (eventIssueCreated, eventIssueReopened, eventIssueResolved, eventUnresolvedIssueDeleted)
- Extra parameters can be added to the webhook URL to post more event types to Mattermost, see https://github.com/mattermost/mattermost-plugin-jira/pull/72/files#diff-afb678e7f5aecce88e775e4aa8c70174R56

Ticket: 
https://mattermost.atlassian.net/browse/MM-15439


server/webhook.go
```diff
@@ -192,39 +225,48 @@ func parse(in io.Reader, linkf func(w *JIRAWebhook) string) (*parsedJIRAWebhook,
 	issue := parsed.mdIssueType() + " " + linkf(parsed.JIRAWebhook)
 	switch parsed.WebhookEvent {
 	case "jira:issue_created":
+		parsed.events = parsed.events | eventIssueCreated
 		parsed.style = mdRootStyle
 		headline = fmt.Sprintf("created %v", issue)
 		parsed.details = parsed.mdIssueCreatedDetails()
 		parsed.text = parsed.mdIssueDescription()
 	case "jira:issue_deleted":
+		parsed.events = parsed.events | eventIssueDeleted
```
levb: See https://github.com/mattermost/mattermost-plugin-jira/blob/c6f5fc6df455522509fbe95e22d94a9ee620b3d3/server/webhook.go#L66

server/webhook.go
```diff
@@ -18,56 +18,73 @@ import (
...
+	eventIssueUpdateRank
+	eventIssueUpdatedAttachment
+	eventIssueUpdatedLabels
+	eventCommentCreated
+	eventCommentDeleted
+	eventCommentUpdated
+	eventMax = iota
+)
+
+const maskLegacy = eventIssueCreated |
```
levb: see https://github.com/mattermost/mattermost-plugin-jira/blob/c6f5fc6df455522509fbe95e22d94a9ee620b3d3/server/webhook.go#L63-L79 and https://github.com/mattermost/mattermost-plugin-jira/blob/c6f5fc6df455522509fbe95e22d94a9ee620b3d3/server/webhook.go#L146-L151

jasonblais:
> By default, only the events supported by 1.0 are shown (eventIssueCreated, eventIssueReopened, eventIssueResolved, eventUnresolvedIssueDeleted)

Does this mean the following events are supported, as listed in ticket

- Issue created
- Issue updated - issue assignee updated; issue transitioned (e.g. Reopened, In Progress, Resolved, Submitted)
- Issue deleted
- Comment created
- Comment updated
- Comment deleted

server/webhook.go
```diff
@@ -18,56 +18,74 @@ import (
...
+const maskComments = eventCommentCreated |
+	eventCommentDeleted |
+	eventCommentUpdated
+
+const maskDefault = maskLegacy
+
+// The keys listed here can be used in the Webhook URL to control what events
+// are posted to Mattermost. A matching parameter with a non-empty value must
+// be added to turn on the event display.
+var eventParamMasks = map[string]uint64{
```
jasonblais: Propose
```
	"updated_assignee":    eventIssueAssigned,                        // updated assignee
	"updated_priority":       eventIssueUpdatePriority,               // updated priority
	"updated_summary":    eventIssueUpdateSummary,          // updated summary
	"updated_description": eventIssueUpdateDescription,     // updated description
	"updated_sprint":          eventIssueUpdateSprint,               // updated sprint
	"updated_rank":             eventIssueUpdateRank,                // updated rank
	"updated_attachment": eventIssueUpdatedAttachments, // updated attachments
	"updated_labels":          eventIssueUpdatedLabels,           // updated labels
	"updated_all":                 ^(-1 << eventMax),                       // all update events
```

1. because Jira handles issue creation/deletion and comment creation/update/deletion separately, so these don't need to be included.

image

2. add all supported events (doesn't hurt to include?)
3. remove issue updated status, since already included by default

levb: @jasonblais 
1. 2/5 to keep the param(s) to explicitly control the visibility (vs what's sent down). Do not include comment events in the default mask. Specifically I am concerned about any existing webhooks that are set up to include all events, we should make sure that their traffic does not change materially.
2. Sure. 0/5 I'd prefer shorter param names.
3. Good catch. Will do.

server/utils.go
```diff
@@ -92,11 +92,18 @@ func parseJIRAIssuesFromText(text string, keys []string) []string {
...
+func getIssueURL(issue *jira.Issue) string {
+	if issue == nil {
+		return ""
+	}
+	u, _ := url.Parse(issue.Self)
+	return u.Scheme + "://" + u.Host + "/browse/" + issue.Key
 }
 
-func getUserURL(issue *JIRAWebhookIssue, user *jira.User) string {
+func getUserURL(issue *jira.Issue, user *jira.User) string {
```
crspeller: Why does this take an issue?

server/webhook.go
```diff
@@ -18,56 +18,74 @@ import (
...
 type JIRAWebhook struct {
-	WebhookEvent string
-	Issue        JIRAWebhookIssue
-	User         jira.User
-	Comment      struct {
-		Body         string
-		UpdateAuthor jira.User
-	}
-	ChangeLog struct {
+	WebhookEvent string       `json:"webhookEvent,omitempty"`
```
crspeller: standard to use lower snake case.

server/webhook.go
```diff
@@ -192,39 +226,48 @@ func parse(in io.Reader, linkf func(w *JIRAWebhook) string) (*parsedJIRAWebhook,
 	issue := parsed.mdIssueType() + " " + linkf(parsed.JIRAWebhook)
 	switch parsed.WebhookEvent {
 	case "jira:issue_created":
+		parsed.events = parsed.events | eventIssueCreated
```
crspeller: I feel like the concern of populating this bitset should be removed from everything else that's going on here. Then the code here can use this bitset.

server/webhook.go
crspeller: Let's make this more clear. I don't think there is a need to confuse this public interface with cryptic parameter names.

server/webhook.go
levb: These come from Jira

crspeller: Oh ya! Well I will take it up with Atlassian then :P

server/webhook.go
jasonblais: Honestly, I'm confused. With ignoring my proposal and only considering what's currently implemented in this PR, if I want to send notifications for
 - issue creation
 - issue updates for assignee, sprint change
 - comment creation

What do I need to configure as an admin?

levb: `...&updated_created_comment=1&updated_assignee=1&updated_sprint=1`

you will also get "Reopened", "Resolved", and "DeletedUnresolved" - there's no way to turn them off (short of using JQL in Jira)

jasonblais: @levb and what if we simply reduced the scope of the supported webhook events to v1.1, and only added the issue assignment event?

Then the admin wouldn't need to configure any of these extra parameters?

This, assuming the admin can control whether comments creation/update/deletion is handled through the attached screen
image

jasonblais: Chatted with @levb - the proposal is to support the following events by default, configurable in Jira Settings > System > WebHooks, and not document the additional flags.

- Issue: Created (no change)
- Issue: Updated (ADD issue assigned event)
- Issue: Deleted (no change)
- Comment: Created (new)
- Comment: Updated (new)
- Comment: Updated (new)

image

levb:
@crspeller @cpoile this is ready re-review

server/webhook.go
```diff
@@ -192,43 +228,52 @@ func parse(in io.Reader, linkf func(w *JIRAWebhook) string) (*parsedJIRAWebhook,
 	issue := parsed.mdIssueType() + " " + linkf(parsed.JIRAWebhook)
 	switch parsed.WebhookEvent {
 	case "jira:issue_created":
+		parsed.event(eventCreated)
```
crspeller: I like the encapsulation, maybe consider completely encapsulating the functionality of the bitset? Or to save time use a library like: https://github.com/willf/bitset

The other thing is there is a mix of concerns in this function. We are creating the bitset but also deriving a headline and other stuff. Maybe split this into multiple functions? Like `createEventsBitset` and `createHeadline`, etc.

server/utils.go
```diff
@@ -92,11 +92,15 @@ func parseJIRAIssuesFromText(text string, keys []string) []string {
...
+	if issue == nil {
+		return ""
+	}
+	u, _ := url.Parse(issue.Self)
+	return u.Scheme + "://" + u.Host + "/browse/" + issue.Key
 }
 
-func getUserURL(issue *JIRAWebhookIssue, user *jira.User) string {
+func getUserURL(user *jira.User) string {
+	// TODO is this right?
```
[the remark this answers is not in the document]
levb: This function is not used in 2.0, and will likely be deprecated with @cpoile's implementation of `/jira settings notify`. 

I can cut a separate PR and simply remove all "notify" functionality from v2.0 since it's not used. This function will disappear then.

server/webhook.go
levb: I'll do a separate PR for 2.1 for this. I am inclined not to bring in an extra dependency for what I consider temporary feature flags. If exposing the bit-logic is not sufficiently intention-revealing, I'd rather just use a go map, it's not a material performance difference in the context.

crspeller: Alright.

jasonblais:
@levb Please create tickets for any follow-up PRs for 2.0/2.1 if applicable, e.g. https://github.com/mattermost/mattermost-plugin-jira/pull/72#discussion_r287403109 and https://github.com/mattermost/mattermost-plugin-jira/pull/72#discussion_r287409699

PR #75, opened 2019-05-24
[MM-15537] Fix 'connect to jira' menu item when not connected

cpoile:
#### Summary
The 'Connect to Jira' menu item will not appear if there is no installed Jira instance.

Notes:
- One edge case: When connecting server, if a user was already connected to that server, their client won't be updated until they reload. Possible second ticket @jasonblais ?
- Some refactoring  (eg. state variables to make it more clear, connected -> userConnected)
- now using `/jira uninstall cloud/server <URL>`, and the cloud server's callback to ACUnistalled does nothing because we can't authenticate it with the JWT. 

#### Links
[MM-15537](https://mattermost.atlassian.net/browse/MM-15537)
[MM-15828](https://mattermost.atlassian.net/browse/MM-15828)

### DIFF
--- a/server/atlassian_connect.go
+++ b/server/atlassian_connect.go
@@ -5,6 +5,7 @@ package main
 
 import (
 	"encoding/json"
+	"github.com/mattermost/mattermost-server/model"
 	"io/ioutil"
 	"net/http"
 
@@ -76,7 +77,7 @@ func httpACInstalled(p *Plugin, w http.ResponseWriter, r *http.Request) (int, er
 	if err != nil {
 		return http.StatusInternalServerError, err
 	}
-	err = p.StoreCurrentJIRAInstance(jiraInstance)
+	err = p.StoreCurrentJIRAInstanceAndNotify(jiraInstance)
 	if err != nil {
 		return http.StatusInternalServerError, err
 	}
@@ -93,6 +94,32 @@ func httpACUninstalled(p *Plugin, w http.ResponseWriter, r *http.Request) (int, 
 	if r.Method != http.MethodPost {
 		return http.StatusMethodNotAllowed,
 			errors.New("method " + r.Method + " is not allowed, must be POST")
+	}
+
+	// Notify users we have uninstalled an instance
+	// We're doing this first because there's a chance of getting into a state where the clients think there is an
+	// instance_installed, but the admin has deleted the instance somehow (eg. through the slash commands if we
+	// expose them). If that happens, then we'll error out of this function before we get to the end.
+	// We want to make sure to send the notification to clients that there is no instance_installed.
+	p.API.PublishWebSocketEvent(
+		wSEventInstanceStatus,
+		map[string]interface{}{
+			"instance_installed": false,
+		},
+		&model.WebsocketBroadcast{},
+	)
+
+	ji, err := p.LoadCurrentJIRAInstance()
+	if err != nil {
+		return http.StatusInternalServerError,
+			errors.WithMessage(err, "no current Jira instance to uninstall")
+	}
+
+	// Remove the instance. If we don't, we will mistakenly think that there is a connected instance.
+	err = p.DeleteJiraInstance(ji.GetURL())
+	if err != nil {
+		return http.StatusInternalServerError,
+			errors.WithMessage(err, "failed to delete Jira instance"+ji.GetURL())
 	}
 
 	_ = json.NewEncoder(w).Encode([]string{"OK"})
--- a/server/atlassian_connect.go
+++ b/server/atlassian_connect.go
@@ -5,7 +5,6 @@ package main
 
 import (
 	"encoding/json"
-	"github.com/mattermost/mattermost-server/model"
 	"io/ioutil"
 	"net/http"
 
@@ -96,32 +95,7 @@ func httpACUninstalled(p *Plugin, w http.ResponseWriter, r *http.Request) (int, 
 			errors.New("method " + r.Method + " is not allowed, must be POST")
 	}
 
-	// Notify users we have uninstalled an instance
-	// We're doing this first because there's a chance of getting into a state where the clients think there is an
-	// instance_installed, but the admin has deleted the instance somehow (eg. through the slash commands if we
-	// expose them). If that happens, then we'll error out of this function before we get to the end.
-	// We want to make sure to send the notification to clients that there is no instance_installed.
-	p.API.PublishWebSocketEvent(
-		wSEventInstanceStatus,
-		map[string]interface{}{
-			"instance_installed": false,
-		},
-		&model.WebsocketBroadcast{},
-	)
-
-	ji, err := p.LoadCurrentJIRAInstance()
-	if err != nil {
-		return http.StatusInternalServerError,
-			errors.WithMessage(err, "no current Jira instance to uninstall")
-	}
-
-	// Remove the instance. If we don't, we will mistakenly think that there is a connected instance.
-	err = p.DeleteJiraInstance(ji.GetURL())
-	if err != nil {
-		return http.StatusInternalServerError,
-			errors.WithMessage(err, "failed to delete Jira instance"+ji.GetURL())
-	}
-
+	// Just send an ok to the Jira server, even though we're not doing anything.
 	_ = json.NewEncoder(w).Encode([]string{"OK"})
 	return http.StatusOK, nil
 }
--- a/server/command.go
+++ b/server/command.go
@@ -334,7 +334,7 @@ func executeUninstallServer(p *Plugin, c *plugin.Context, header *model.CommandA
 		&model.WebsocketBroadcast{},
 	)
 
-	const uninstallInstructions = `Jira instance successfully disconnected. Go to XXX to remove the application in your Jira Server instance.`
+	const uninstallInstructions = `Jira instance successfully disconnected. Go to **Settings > Applications > Application Links** to remove the application in your Jira Server or Data Center instance.`
 
 	return responsef(uninstallInstructions)
 }
--- a/server/command.go
+++ b/server/command.go
@@ -299,7 +299,7 @@ func executeUninstallServer(p *Plugin, c *plugin.Context, header *model.CommandA
 		return responsef("%v", err)
 	}
 	if !authorized {
-		return responsef("`/jira uninstall` can only be run by a system administrator.")
+		return responsef("`/jira uninstall` can only be run by a System Administrator.")
 	}
 	if len(args) != 1 {
 		return help()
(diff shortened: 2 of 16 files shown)


jasonblais:
> One edge case: When connecting server, if a user was already connected to that server, their client won't be updated until they reload. Possible second ticket @jasonblais ?

Do you mean that if
1. User-A had already connected to a Jira server, call it Jira-A
2. Admin-A connects MM to another Jira server, call it Jira-B

Then User-A's client will continue to be connected to Jira-A until a page reload?

/cc @DHaussermann this is the PR that should resolve remaining issues with the upgrade path.

cpoile:
@jasonblais What I mean is: 
1. Admin-A connects to Jira-A, User-A connects their account to Jira-A.
1. Admin-A disconnects from Jira-A, User-A is notified that they are no longer connected to a Jira instance (they are shown no menu items related to Jira).
1. User-A logs off, logs back on. Their state is "jira intance not connected" AND "user account not connected" (because there was no jira instance to test if their account was connected)
1. Admin-A connects back to Jira-A.
1. User-A is notified that they are now connected to a Jira instance. But their client state still thinks their user account is NOT connected.
1. User-A reloads their webapp, and it correctly pulls that their account is connected.

It's a weird edge case that might never happen, but it's there.

jasonblais:
Thank you. I created a separate ticket for this one, but not critical for 2.0 as you suggested: https://mattermost.atlassian.net/browse/MM-15776

Nice find!

server/user.go
```diff
@@ -106,17 +107,20 @@ func httpAPIGetUserInfo(ji Instance, w http.ResponseWriter, r *http.Request) (in
 	}
 
 	resp := UserInfo{}
-	jiraUser, err := ji.GetPlugin().LoadJIRAUser(ji, mattermostUserId)
-	if err == nil {
-		resp = UserInfo{
-			JIRAUser:    jiraUser,
-			IsConnected: true,
-			JIRAURL:     ji.GetURL(),
+
```
levb: nit: empty line?

server/atlassian_connect.go
```diff
@@ -95,6 +96,32 @@ func httpACUninstalled(p *Plugin, w http.ResponseWriter, r *http.Request) (int,
 			errors.New("method " + r.Method + " is not allowed, must be POST")
 	}
 
+	// Notify users we have uninstalled an instance
```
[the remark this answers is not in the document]
crspeller: Let me clarify what @levb means here. Since this is a public route I can create a DOS attack here by spamming uninstall. So we need to at least check the request is valid before sending this event.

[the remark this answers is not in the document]
cpoile: @crspeller @levb I get that, and makes sense. I'll reproduce a message I sent to Lev on Friday:

I've been having trouble with JWT... when Atlassian sends the Uninstalled POST, it puts the JWT in the Header under "Authorization". Not in the body, like it does for the other requests. The problem is: I believe the Authorization header gets stripped out on redirect, so I can't access it through ngrok, so I can't test that it's working. 

This is what's sent in the body... Is anything here secure enough to use as a proxy to verify the request?
```javascript
{
    "key": "mattermost_https_chrisp_ngrok_io",
    "clientKey": "5236f67b-1021-3c6f-94da-b4e016d62ac8",
    "publicKey": "MIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQCvtU0OP6fwW1JsB4tfBESeObHF/tMqvJLa18NZo7wyTYQHqncE6ucJF0hzuPc/0lEOT28DvxJZRXF2WBDqRc2XRkrZnkfEnFPoQU4NLvv9DYY+/zLp0hwrDNasY0qVL5SxaBeA1PafjDqvwt+rn/bRsSyUlDyxzVYEsb7qc0hGdQIDAQAB",
    "serverVersion": "100103",
    "pluginsVersion": "1.403.0",
    "baseUrl": "https://mmtest.atlassian.net",
    "productType": "jira",
    "description": "Atlassian JIRA at https://mmtest.atlassian.net ",
    "eventType": "uninstalled"
}
```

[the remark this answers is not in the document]
levb: 1/5 I suggest removing /uninstalled support if we can't easily make this work. I was talking to @jasonblais this morning about the (eventual, 2.1) need for a `/jira uninstall` command to remove the current Jira Server instance, it can also be used in lieu of `/uninstall` http callback.

For 2.0 there will be no way to properly uninstall the instance, we will recommend disabling the plugin.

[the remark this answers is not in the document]
cpoile: @levb For 2.0 there is a proper way to uninstall (for cloud only), and it's through this uninstall endpoint. The only problem is the JWT authorization. I'm 95% sure what I have will work, I just can't test it on my machine (and no-one using ngrok can test it either, if I'm right in assuming the problem is that the Authorization header gets stripped).

Howabout I make `/jira uninstall` work for v2.0? Maybe a separate ticket for that. @jasonblais ?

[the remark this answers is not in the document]
jasonblais: I've created a ticket, we can discuss tomorrow during sprint planning if that goes to v2.0 or v2.1: https://mattermost.atlassian.net/browse/MM-15828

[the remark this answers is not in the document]
cpoile: resolving because we're going with https://mattermost.atlassian.net/browse/MM-15828

server/atlassian_connect.go
```diff
@@ -95,6 +96,32 @@ func httpACUninstalled(p *Plugin, w http.ResponseWriter, r *http.Request) (int,
...
+	// We're doing this first because there's a chance of getting into a state where the clients think there is an
+	// instance_installed, but the admin has deleted the instance somehow (eg. through the slash commands if we
+	// expose them). If that happens, then we'll error out of this function before we get to the end.
+	// We want to make sure to send the notification to clients that there is no instance_installed.
+	p.API.PublishWebSocketEvent(
```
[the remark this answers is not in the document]
cpoile: Originally I thought I wanted to send out the notification just in case. Because if any of the other calls failed, there is a good chance the instance wouldn't be in a good state, and I didn't want users to be left with menu options for a jira that isn't connected.

0/5 I think the chances of that are so low that we don't need to worry about it. What do you think? 

For now I'm putting the websocket event after everything (in the uninstall commands in command.go).

cpoile:
@jasonblais updated. Waiting for new responsef instructions for both `executeUninstallCloud` and `executeUninstallServer` in `command.go`.  Thanks!

server/atlassian_connect.go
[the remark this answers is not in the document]
levb: I'm fine with that - ", there is a good chance the instance wouldn't be in a good state, "

server/command.go
```diff
@@ -13,9 +13,13 @@ const helpText = "###### Mattermost Jira Plugin - Slash Command Help\n" +
 	"* `/jira connect` - Connect your Mattermost account to your Jira account\n" +
 	"* `/jira disconnect` - Disconnect your Mattermost account from your Jira account\n" +
 	"* `/jira transition <issue-key> <state>` - Change the state of a Jira issue\n" +
-	"\nFor System Administrators:\n" +
+	"\n###### For system administrators:\n" +
```
jasonblais:
```suggestion
	"\n###### For System Administrators:\n" +
```

server/command.go
```diff
@@ -239,6 +245,100 @@ If you see an option to create a Jira issue, you're all set! If not, refer to ou
...
 
+// executeUninstallCloud will uninstall the jira cloud instance if the url matches, and then update all connected
+// clients so that their Jira-related menu options are removed.
+func executeUninstallCloud(p *Plugin, c *plugin.Context, header *model.CommandArgs, args ...string) *model.CommandResponse {
+	authorized, err := authorizedSysAdmin(p, header.UserId)
+	if err != nil {
+		return responsef("%v", err)
+	}
+	if !authorized {
+		return responsef("`/jira uninstall` can only be run by a system administrator.")
```
jasonblais:
```suggestion
		return responsef("`/jira uninstall` can only be run by a System Administrator.")
```

server/command.go
```diff
@@ -239,6 +245,100 @@ If you see an option to create a Jira issue, you're all set! If not, refer to ou
...
+	return responsef(uninstallInstructions)
+}
+
+func executeUninstallServer(p *Plugin, c *plugin.Context, header *model.CommandArgs, args ...string) *model.CommandResponse {
+	authorized, err := authorizedSysAdmin(p, header.UserId)
+	if err != nil {
+		return responsef("%v", err)
+	}
+	if !authorized {
+		return responsef("`/jira uninstall` can only be run by a system administrator.")
```
jasonblais:
```suggestion
		return responsef("`/jira uninstall` can only be run by a System Administrator.")
```

server/command.go
```diff
@@ -239,6 +245,100 @@ If you see an option to create a Jira issue, you're all set! If not, refer to ou
...
+	// Notify users we have uninstalled an instance
+	p.API.PublishWebSocketEvent(
+		wSEventInstanceStatus,
+		map[string]interface{}{
+			"instance_installed": false,
+		},
+		&model.WebsocketBroadcast{},
+	)
+
+	const uninstallInstructions = `Jira instance successfully disconnected. Go to XXX to remove the application in your Jira Cloud instance.`
```
jasonblais:
```suggestion
	const uninstallInstructions = `Jira instance successfully disconnected. Go to **Settings > Apps > Manage Apps** to remove the application in your Jira Cloud instance.`
```

server/command.go
```diff
@@ -239,6 +245,100 @@ If you see an option to create a Jira issue, you're all set! If not, refer to ou
...
+	// Notify users we have uninstalled an instance
+	p.API.PublishWebSocketEvent(
+		wSEventInstanceStatus,
+		map[string]interface{}{
+			"instance_installed": false,
+		},
+		&model.WebsocketBroadcast{},
+	)
+
+	const uninstallInstructions = `Jira instance successfully disconnected. Go to XXX to remove the application in your Jira Server instance.`
```
jasonblais:
```suggestion
	const uninstallInstructions = `Jira instance successfully disconnected. Go to **Settings > Applications > Application Links** to remove the application in your Jira Server or Data Center instance.`
```

server/command.go
jasonblais: (Not blocking this PR) We would ideally make this a link to the appropriate page, similar to what we do in `/jira install` command output https://github.com/mattermost/mattermost-plugin-jira/blob/master/server/command.go#L209

cpoile: @jasonblais can we make that a quick ticket? I'll take it. I want to merge this. :)

jasonblais: Yes, like I said, not blocking this PR :)

jasonblais: https://mattermost.atlassian.net/browse/MM-15903

PR #78, opened 2019-05-26
MM-15733 - Making react select theme compatible

asaadmahmood:
Screenshot 2019-05-26 at 6 17 55 PM

### DIFF
--- a/webapp/src/components/react_select_setting.jsx
+++ b/webapp/src/components/react_select_setting.jsx
@@ -6,12 +6,15 @@ import PropTypes from 'prop-types';
 
 import ReactSelect from 'react-select';
 
+import {changeOpacity} from 'mattermost-redux/utils/theme_utils';
+
 import Setting from 'components/setting';
 
 export default class ReactSelectSetting extends React.PureComponent {
     static propTypes = {
         name: PropTypes.string.isRequired,
         onChange: PropTypes.func,
+        theme: PropTypes.object,
     };
 
     handleChange = (value) => {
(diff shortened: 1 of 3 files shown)


webapp/src/components/react_select_setting.jsx
```diff
@@ -21,13 +24,71 @@ export default class ReactSelectSetting extends React.PureComponent {
     }
 
     render() {
+        const theme = this.props.theme;
+        let styles = {};
+
+        if (typeof theme !== 'undefined') {
```
crspeller:
```suggestion
        if (theme) {
```

webapp/src/components/react_select_setting.jsx
```diff
@@ -21,13 +24,71 @@ export default class ReactSelectSetting extends React.PureComponent {
...
+            styles = {
```
crspeller: Can we move this out to a separate function? Kinda like `getStyle`.

asaadmahmood:
Done @crspeller

webapp/src/components/react_select_setting.jsx
```diff
@@ -20,6 +23,58 @@ export default class ReactSelectSetting extends React.PureComponent {
         }
     }
 
+    getStyle = (theme) => ({
```
cpoile: what happens now if theme is undefined or null?

asaadmahmood: @cpoile Looks like previously for some reason, the theme code that I was using to set the style was running at an earlier point where at times the theme was undefined, here however, I do not get the undefined error, so I would assume that the theme would always be there in this case.

jasonblais: @cpoile on above

cpoile: @jasonblais we need to put a note for Dylan to test this a lot and be on the lookout for errors right here then, just in case.

PR #79, opened 2019-05-27
[MM-15677] autocomplete for issue keys in Attach to Jira Issue

cpoile:
#### Summary
- autocomplete for issues when attaching a post to a Jira issue
- uses debouncing to prevent swamping the server with intermediate requests as the user types.
- removed project selector from the attach modal -- it wasn't needed, and kind of confusing if you're also searching (what does it do? it's not required... it just acts as a filter.) We'll wait to see if anyone asks for it.
  - so that meant we could get rid of that supporting code
- see pictures below for UX

#### Links
[MM-15677](https://mattermost.atlassian.net/browse/MM-15677)

### DIFF
--- a/server/issue.go
+++ b/server/issue.go
@@ -190,6 +190,78 @@ func httpAPIGetCreateIssueMetadata(ji Instance, w http.ResponseWriter, r *http.R
 	return http.StatusOK, nil
 }
 
+func httpAPIGetSearchIssues(ji Instance, w http.ResponseWriter, r *http.Request) (int, error) {
+	if r.Method != http.MethodGet {
+		return http.StatusMethodNotAllowed,
+			errors.New("Request: " + r.Method + " is not allowed, must be GET")
+	}
+
+	mattermostUserId := r.Header.Get("Mattermost-User-Id")
+	if mattermostUserId == "" {
+		return http.StatusUnauthorized, errors.New("not authorized")
+	}
+
+	jiraUser, err := ji.GetPlugin().LoadJIRAUser(ji, mattermostUserId)
+	if err != nil {
+		return http.StatusInternalServerError, err
+	}
+
+	jiraClient, err := ji.GetJIRAClient(jiraUser)
+	if err != nil {
+		return http.StatusInternalServerError, err
+	}
+
+	err = r.ParseForm()
+	if err != nil {
+		return http.StatusInternalServerError, errors.WithMessage(err, "failed to parse request")
+	}
+
+	jqlString := r.Form.Get("jql")
+
+	searchRes, resp, err := jiraClient.Issue.Search(jqlString, &jira.SearchOptions{
+		MaxResults: 50,
+		Fields:     []string{"key", "summary"},
+	})
+
+	if err != nil {
+		message := "failed to get search results"
+		if resp != nil {
+			bb, _ := ioutil.ReadAll(resp.Body)
+			resp.Body.Close()
+			message += ", details: " + string(bb)
+		}
+		return http.StatusInternalServerError,
+			errors.WithMessage(err, message)
+	}
+
+	// We only need to send down a summary of the data
+	type keySummary struct {
+		Key     string `json:"key"`
+		Summary string `json:"summary"`
+	}
+	resSummary := make([]keySummary, 0, len(searchRes))
+	for _, res := range searchRes {
+		resSummary = append(resSummary, keySummary{
+			Key:     res.Key,
+			Summary: res.Fields.Summary,
+		})
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	b, err := json.Marshal(resSummary)
+	if err != nil {
+		return http.StatusInternalServerError,
+			errors.WithMessage(err, "failed to marshal response")
+	}
+	_, err = w.Write(b)
+	if err != nil {
+		return http.StatusInternalServerError,
+			errors.WithMessage(err, "failed to write response")
+	}
+
+	return http.StatusOK, nil
+}
+
 func httpAPIAttachCommentToIssue(ji Instance, w http.ResponseWriter, r *http.Request) (int, error) {
 	if r.Method != http.MethodPost {
 		return http.StatusMethodNotAllowed,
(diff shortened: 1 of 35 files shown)


jasonblais:
@cpoile I'm leaning towards only supporting one way of searching to keep the user experience simple and obvious.

Between the two options (searching by issue key or searching by issue text), I'd propose only supporting searching by issue text because
 - (assumption) users typically remember something about the issue summary or description, rather than the specific issue key, when searching for an issue
 - (for reference) other integrations such as Slack's Jira Cloud integration does this

What are your thoughts on that approach?

Also adding @asaadmahmood for UX review.

cpoile:
@jasonblais I agree that keeping one search box is definitely better. But, the trouble with using only the one search box is that when the user types, they expect to be narrowing down the options in the select box -- so they start typing "ext" or even "1" and then nothing appears because the words "ext" or "1" don't appear in the text of the issues. So it doesn't match what they expect from every other select-box they've ever used before.

jasonblais:
> so they start typing "ext" or even "1" and then nothing appears because the words "ext" or "1" don't appear in the text of the issues. So it doesn't match what they expect from every other select-box they've ever used before.

That's a fair point. Thoughts:

1 - Are all the fields in a select box or autocomplete always searchable?

E.g. when using slash command `/`, only the command name is searchable, not the help text. Same when searching for users in System Console, only the user attributes (name, email address, ..) are searchable.

That said, both of the above examples are a little different from select boxes. @asaadmahmood (and you) likely have more insight on the UX for select boxes.

2 - We could style the issue key in the autocomplete in such a way that signifies it's not a searchable field. E.g. grey it out.

3 - To use the issue key or issue number to narrow down search results is helpful when it belongs to the initial list of 50 issues returned in the autocomplete.

I wonder how common use case that is for large Enterprises and DevOps teams.

cpoile:
@jasonblais that's what I was struggling with. If only your search terms in JQL also searched the issue-keys...

1. usually. You usually narrow down by typing. This is different than a slash command. I think users expect to be able to type and match anything in a select box (right @asaadmahmood ?)
1. That's a possibility. But then they can't narrow down by issue number no matter what -- they'll have to scroll or narrow down by text searching (and it will take a few times for them to learn that this is what's happening here)
1. to be honest, it's non-intuitive either way. The way I have it now they _can_ narrow down by issue key, but only if it's in those first 50 (or whatever we choose) issues. If it's not there they have to search using text, which is a change of context and thinking -- what was the issue called again? what words did we use?

That's why this is PM and UX review first. :)

jasonblais:
Thanks @cpoile - I'll see what feedback or thoughts @asaadmahmood will have.

One thing we should consider is whether to optimize for narrowing search by issue key within the first 50 issues, or to optimize for narrowing search by issue text.

Once we know what to optimize for, it'll help guide the ultimate solution.

jasonblais:
@cpoile - we can discuss more tomorrow. To summarize, we're deciding whether typing `ext-1` in the autocomplete field
1. narrows the existing list of 50 entries based on the issue key, or
2. does a JQL search query using `ext-1` as a search term (which ignores the issue key)

Is there an option to get the best of both worlds and do a "union" of the two, i.e., typing `ext-1` narrows the existing list of 50 entries based on the issue key AND pulls additional entries via a JQL search query using `ext-1` as a search term. I'm still thinking about the UX, but it may be an option UX-wise.

Otherwise, I'm 3/5 we should optimize for the second option and return results based on JQL search query [proposed previously](https://github.com/mattermost/mattermost-plugin-jira/pull/79#issuecomment-496241900). Using two search fields isn't obvious.

I have additional thoughts, but the above is a summarized version of it which we can discuss further tomorrow.

/cc @asaad.mahmood

cpoile:
@jasonblais @asaadmahmood  Notes:
- Pressing enter while in the select-box would trigger the `Attach` button and submit. I removed that.
- Added a star at the end of the input string (after chatting with @jasonblais) -- this seems to help when searching.

cpoile:
@asaadmahmood I'm passing in theme to jira_issue_selector.jsx -- can you style it (keeping in mind to check for null theme, as in `react_select_setting.jsx`)?
@jasonblais it's probably easier to make a separate PR for that, right?
Thanks!

jasonblais:
@cpoile Sure, I've queued it in a ticket for Asaad to take a look

server/issue.go
```diff
@@ -199,6 +199,78 @@ func httpAPIGetCreateIssueMetadata(ji Instance, w http.ResponseWriter, r *http.R
...
+	if err != nil {
+		return http.StatusInternalServerError, err
+	}
+
+	jiraClient, err := ji.GetJIRAClient(jiraUser)
+	if err != nil {
+		return http.StatusInternalServerError, err
+	}
+
+	err = r.ParseForm()
```
levb: nit: prefer https://golang.org/pkg/net/http/#Request.FormValue

server/issue.go
```diff
@@ -199,6 +199,78 @@ func httpAPIGetCreateIssueMetadata(ji Instance, w http.ResponseWriter, r *http.R
...
+	}
+
+	jqlString := r.Form.Get("jql")
+
+	searchRes, resp, err := jiraClient.Issue.Search(jqlString, &jira.SearchOptions{
+		MaxResults: 50,
+		Fields:     []string{"key", "summary"},
+	})
+
+	if err != nil {
```
levb: for 2.1, we should refactor into a function, maybe `RespondWithJiraError(status, err, resp)`

asaadmahmood:
@cpoile haven't I already styled it in a previous PR?
https://github.com/mattermost/mattermost-plugin-jira/pull/78

Is this something different?

cpoile:
@asaadmahmood yep, different component. react-select has an async component that is less styled (remember I mentioned that we have to add our own label, help-text, everything else we want). So it's in a new component I made called `jira_issue_selector`. That needs to be theme styled (and the inline syling needs to be cleaned up if you can think of a better way). Maybe it's as easy as copy-pasting, or maybe there's a better way of doing it, up to you.

Best way (I think?) would be to make a new branch off this branch, start making your changes, then merge #79's changes when it gets merged to jira2, then submit a PR with your changes.  (unless you can think of a better way -- up to you of course)

PR #81, opened 2019-05-27
[MM-15589] Create Issue: (Jira server) Priority field fails to select

jfrerich:
**Summary**
This PR addresses the issue when a user tries to select a priority field when creating a ticket.  

The priority field type, inside the fields object, is an object, not a string.  This requires handling priority differently to satisfy the embedded priority struct inside the IssueFields struct.

For the priority ReactSelectSetting component, filter against
this.props.value.id to get the value, not this.props.value which is an
object

Below is the working case which appropriately selects the priority field.
image

An example of a ticket created with the priority field set can be viewed [Here](https://mmtest.atlassian.net/browse/EXT-128)

**Ticket Link**
https://mattermost.atlassian.net/browse/MM-15536

### DIFF
--- a/webapp/src/components/jira_field.jsx
+++ b/webapp/src/components/jira_field.jsx
@@ -11,6 +11,7 @@ export default class JiraField extends React.PureComponent {
     static propTypes = {
         id: PropTypes.object.isRequired,
         field: PropTypes.object.isRequired,
+        fieldKey: PropTypes.string.isRequired,
         obeyRequired: PropTypes.bool,
         onChange: PropTypes.func,
         value: PropTypes.any,
@@ -40,16 +41,16 @@ export default class JiraField extends React.PureComponent {
         return (
             {value: allowedValue.id, label: iconLabel}
         );
-    }
+    };
 
     render() {
-        const field = this.props.field;
+        const {field, fieldKey, obeyRequired} = this.props;
 
         if (field.schema.system === 'description') {
             return (
                 <Input
-                    key={field.key}
-                    id={field.key}
+                    key={fieldKey}
+                    id={fieldKey}
                     label={field.name}
                     type='textarea'
                     onChange={this.handleChange}
@@ -62,8 +63,8 @@ export default class JiraField extends React.PureComponent {
         if (field.schema.type === 'string') {
             return (
                 <Input
-                    key={field.key}
-                    id={field.key}
+                    key={fieldKey}
+                    id={fieldKey}
                     label={field.name}
                     type='input'
                     onChange={this.handleChange}
@@ -79,20 +80,14 @@ export default class JiraField extends React.PureComponent {
 
             return (
                 <ReactSelectSetting
-                    key={field.key}
-                    name={field.key}
+                    key={fieldKey}
+                    name={fieldKey}
                     label={field.name}
                     options={options}
-                    required={this.props.obeyRequired && field.required}
-                    onChange={this.handleChange}
+                    required={obeyRequired && field.required}
+                    onChange={(id, val) => this.props.onChange(id, {id: val})}
                     isMulti={false}
-                    value={options.filter((option) => {
-                        if (!this.props.value) {
-                            return false;
-                        }
-                        return option.value === this.props.value.id;
-                    }
-                    )}
+                    value={options.find((option) => option.value === this.props.value)}
                 />
             );
         }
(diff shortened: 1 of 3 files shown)


webapp/src/components/jira_field.jsx
```diff
@@ -75,6 +75,22 @@ export default class JiraField extends React.PureComponent {
...
+                return (
+                    <ReactSelectSetting
+                        key={field.key}
+                        name={field.key}
+                        label={field.name}
+                        options={options}
+                        required={this.props.obeyRequired && field.required}
+                        onChange={this.handleChange}
+                        isMulti={false}
+                        value={options.filter((option) => option.value === this.props.value.id)}
```
jfrerich: Note to reviewers.  I kept with the spirit of the other components and created a unique component for the priority special case.  I am curious about your thoughts regarding the following alternate solution:

Instead of creating an additional component, I could use the if statement to set a variable to `this.props.value.id` or `this.props.value`.  Note this is the only difference between the two components, in setting the value property.

cpoile: @crspeller is the SelectSetting below the check `if (field.allowedValues && field.allowedValues.length)` used for any other field other than priority? I don't think so, since there isn't any other field that has `allowedValues` (other than project and and issueType which are handled previously anyway).

In which case you don't have to make this special case, just use the original below (and maybe make it more specific like you have here).

Then, can you just alter priority's field structure to conform to what we expect. That would mean you could remove the special case you added to `initialState` and `handleFieldChange` in `create_issue.jsx`. Make sense?

jasonblais:
@jfrerich Do you know if this resolves cases where other Jira fields may fail issue creation?

webapp/src/components/jira_field.jsx
crspeller: Ya I think it is used in some other fields. You can also create a custom field that would have allowed values.
This should be generic though special casing this here is not what we want. Just apply your modifications to below.

webapp/src/components/modals/create_issue/create_issue.jsx
```diff
@@ -24,6 +24,9 @@ const initialState = {
         issuetype: {
             name: '',
         },
+        priority: {
```
crspeller: No need for a default here. There are lots of possible values and we can't default them all.

jfrerich:
> @jfrerich Do you know if this resolves cases where other Jira fields may fail issue creation?

@jasonblais no, this does not solve all Jira fields that may fail issue creation.  We currently handle the following fields: project, issue type, summary, description, and priority.

If we want to cover all of the cases, there is still more work to be done.  Here's a screenshot of just some of the additional fields
 
image

jasonblais:
> @jasonblais no, this does not solve all Jira fields that may fail issue creation. We currently handle the following fields: project, issue type, summary, description, and priority.

> If we want to cover all of the cases, there is still more work to be done. Here's a screenshot of just some of the additional fields

Got it thanks, Jason. I've filed a separate ticket to handle the required fields in general. https://mattermost.atlassian.net/browse/MM-15838

jfrerich:
Thanks @jasonblais   How do you want to proceed for this ticket?  Should we keep the field options to the current set?   To have a modal with 100% coverage of all create issue options will require quite a bit more time.

jasonblais:
For this particular ticket/PR (https://mattermost.atlassian.net/browse/MM-15536), we can keep the scope as-is. (cc: @levb unless you have other thoughts)

For the broader ticket (https://mattermost.atlassian.net/browse/MM-15838), we can discuss on Tuesday's sprint planning. I'm not sure of the development effort, but customers will have different sets of required fields - and from my understanding issue creation will fail unless we support those different fields. So it won't be a good user experience. It actually happened to one prospect testing the plugin last week.

I was hoping there would be a generic approach that works for the different fields, but that doesn't seem to be the case..

webapp/src/components/jira_field.jsx
```diff
@@ -62,29 +57,31 @@ export default class JiraField extends React.PureComponent {
...
+                    name={fieldKey}
                     label={field.name}
                     options={options}
-                    required={this.props.obeyRequired && field.required}
-                    onChange={this.handleChange}
+                    required={obeyRequired && field.required}
+                    onChange={(id, val) => this.props.onChange(id, {id: val})}
                     isMulti={false}
-                    value={options.filter((option) => option.value === this.props.value)}
+                    value={options.find((option) => option.value === this.props.value)}
```
jfrerich: A couple of notes here: 

Nice catch here to use `.find()` to return the object instead of `.filter()` which returns an array.
Also nice choice to assign onChange here so that `handleFieldChange` does not have to handle values of string and objects.

jasonblais:
@levb @jfrerich This is good to merge so we can deploy to community

PR #82, opened 2019-05-27
MM-15439 refactor and merge webhooks #72 from jira2

levb:
This PR merges #72 into master, and adds refactoring to address the PR feedback, testability and webhook tests.

- Refactored webhook code, added tests
- Added separate interfaces (`UserStore`, `InstanceStore`, etc.) to KV methods
- Added store (kv) interface pointers to `type Plugin`
- Added POC mocks for some test stores (as was needed for the webhook tests).


jasonblais:
@levb Is there anything specific you'd like me to review from PM side?

levb:
@jasonblais No, added you reflexively. This should be (almost) identical to v2 as far as functionality of the webhooks is concerned.

server/webhook_http_test.go
```diff
@@ -0,0 +1,255 @@
...
+				mock.AnythingOfTypeArgument("string"),
+				mock.AnythingOfTypeArgument("string"),
+				mock.AnythingOfTypeArgument("string"),
+				mock.AnythingOfTypeArgument("string"),
+				mock.AnythingOfTypeArgument("string"),
+				mock.AnythingOfTypeArgument("string"),
+				mock.AnythingOfTypeArgument("string"),
+				mock.AnythingOfTypeArgument("string")).Return(nil)
+
+			// api.On("KVGet", mock.AnythingOfTypeArgument("string")).Return(make([]byte, 0), (*model.AppError)(nil))
```
cpoile: nit: these are still here for later?

server/webhook_http_test.go
```diff
@@ -0,0 +1,255 @@
...
+			})
+			p.SetAPI(api)
+			p.currentInstanceStore = mockCurrentInstanceStore{&p}
+			p.userStore = mockUserStore{}
+
+			w := httptest.NewRecorder()
+			recorder := &testWebhookWrapper{}
+			prev := webhookWrapperFunc
+			defer func() { webhookWrapperFunc = prev }()
+			webhookWrapperFunc = func(wh Webhook) Webhook {
```
cpoile: this is cool...

server/webhook_parser_misc_test.go
```diff
@@ -0,0 +1,66 @@
...
+			},
+		},
+	}
+
+	assert.Equal(t, "", wh.mdJiraLink("test", "/test"))
+	assert.Equal(t, "", wh.mdIssueDescription())
+	assert.Equal(t, "", wh.mdIssueSummary())
+	assert.Equal(t, "_nobody_", wh.mdIssueAssignee())
+	assert.Equal(t, "", wh.mdIssueType())
+	assert.Equal(t, " ", wh.mdSummaryLink())
```
crspeller: A single space?

server/webhook_parser.go
```diff
@@ -0,0 +1,253 @@
...
+	// For HTTP testing, so we can capture the output of the interface
+	if webhookWrapperFunc != nil {
+		wh = webhookWrapperFunc(wh)
+	}
+
+	return wh, jwh, nil
+}
+
+func parseWebhookChangeLog(jwh *JiraWebhook) Webhook {
+	for _, item := range jwh.ChangeLog.Items {
```
crspeller: We only ever pay attention to the first change-log item we can parse. Is this the indented behavior? We are kinda locking our selves a bit in with this architecture.

server/kv_mock_test.go
```diff
@@ -0,0 +1,59 @@
+// Copyright (c) 2017-present Mattermost, Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
levb: I'm 0/5 on this - the mocks used in the server seem pretty complicated, but either way this is "... POC mocks ...". How about I leave this comment open until it's (almost) merged, and then file a ticket to re-factor these.

server/webhook_http.go
```diff
@@ -0,0 +1,148 @@
...
+	if r.Method != http.MethodPost {
+		return http.StatusMethodNotAllowed,
+			fmt.Errorf("Request: " + r.Method + " is not allowed, must be POST")
+	}
+	cfg := p.getConfig()
+	if cfg.Secret == "" || cfg.UserName == "" {
+		return http.StatusForbidden, fmt.Errorf("Jira plugin not configured correctly; must provide Secret and UserName")
+	}
+	secret := r.FormValue("secret")
+	for {
```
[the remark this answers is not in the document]
levb: This comes from https://github.com/mattermost/mattermost-plugin-jira/pull/53

server/webhook_parser.go
```diff
@@ -0,0 +1,253 @@
...
+		return nil, nil, err
+	}
+	if jwh.WebhookEvent == "" {
+		return nil, jwh, errors.New("No webhook event")
+	}
+	if jwh.Issue.Fields == nil {
+		return nil, jwh, errors.New("Invalid webhook event")
+	}
+
+	wh := func() Webhook {
```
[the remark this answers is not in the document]
levb: > an assignment vs return
Yes, I like that. Simpler.

server/webhook_parser_misc_test.go
levb: These tests were originally added for coverage and/or check for nil pointers and default handling. 1/5 These functions do return an `str1 + " " + str2` and are not supposed to be called with nil/defaults.

server/webhook_parser.go
levb: A separate ticket?

2/5 this is not an architecture issue. a 1-1 mapping of `JiraWebhook -> `Webhook` makes sense. I agree that we should scan all records, and include all recognized changes in the post, or potentially even make multiple posts if applicable. That would be a part of `PostToChannel`.

server/kv.go
```diff
@@ -22,6 +22,54 @@ const (
 	prefixOneTimeSecret    = "ots_" // + unique key that will be deleted after the first verification
 )
 
+type KV interface {
```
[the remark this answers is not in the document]
levb: now just `Store`. Because why not?

levb:
@crspeller comments addressed or responded to. Please review.

server/kv_mock_test.go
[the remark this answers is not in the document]
crspeller: Yup they are complicated but they are auto-generated so it doesn't matter. I would like to see this now because otherwise development would be slowed by constantly having to update the mocks rather than just doing a `make mocks` to fix them.

server/webhook_parser_misc_test.go
crspeller: Ya that's fine then.

server/webhook_parser.go
crspeller: Sure make a ticket.

server/kv_mock_test.go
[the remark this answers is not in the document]
levb: Filed https://mattermost.atlassian.net/browse/MM-15926 to address this separately. I tried using mockery, but failed to make it work properly with the source files being in the `main` package. It allows to generate mocks "inpkg" but then their names don't match, and one has to set up multiple "On" hooks for all specific concrete types coming in, or write custom matching functions.

I propose that we merge with the DIY mocks, then restructure the bulk of the source codee into separate package(s), and then add mockery.

server/webhook_parser.go
levb: https://mattermost.atlassian.net/browse/MM-15927

PR #93, opened 2019-05-28
MM-15866: Webhook presentation tweaks in jira2

levb:
Discussed with @jasonblais offline - some small improvements:
- Added the issue description to the headline of the posts
- Remove the (now redundant) headline from the Slack attachment
- Do not use Slack attachments for 1-liners

https://mattermost.atlassian.net/browse/MM-15866

### DIFF
--- a/server/webhook.go
+++ b/server/webhook.go
@@ -295,8 +295,6 @@ func (p *parsedJIRAWebhook) fromChangeLog(issue string) (string, string) {
 	for _, item := range p.ChangeLog.Items {
 		to := item.ToString
 		from := item.FromString
-		lto := strings.ToLower(item.ToString)
-		lfrom := strings.ToLower(item.FromString)
 		switch {
 		case item.Field == "resolution" && to == "" && from != "":
 			p.event(eventUpdatedReopened)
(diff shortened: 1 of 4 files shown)


server/webhook.go
```diff
@@ -296,17 +306,21 @@ func (p *parsedJIRAWebhook) fromChangeLog(issue string) (string, string) {
 			p.event(eventUpdatedResolved)
 			return fmt.Sprintf("resolved %v", issue), ""
 
-		case item.Field == "status" && to == "Backlog":
+		case item.Field == "status" && lto == "in progress":
```
jasonblais: @levb Instead of hardcoding to `in progress`, etc., I'm wondering if we're able to use the statuses defined in Jira.

server/webhook.go
```diff
@@ -296,17 +306,21 @@ func (p *parsedJIRAWebhook) fromChangeLog(issue string) (string, string) {
...
 			p.event(eventUpdatedStatus)
-			return fmt.Sprintf("moved %v to backlog", issue), ""
+			return fmt.Sprintf("started working on %v", issue), ""
```
jasonblais: same here (and below)

server/webhook.go
```diff
@@ -296,17 +306,21 @@ func (p *parsedJIRAWebhook) fromChangeLog(issue string) (string, string) {
...
+
+		case item.Field == "status" && lfrom == "in progress" && lto == "done":
+			p.event(eventUpdatedStatus)
+			return fmt.Sprintf("finished work on %v", issue), ""
 
-		case item.Field == "status" && to == "Selected for Development":
+		case item.Field == "status":
 			p.event(eventUpdatedStatus)
-			return fmt.Sprintf("selected %v for development", issue), ""
+			return fmt.Sprintf("transitioned %v to %q", issue, to), ""
```
jasonblais: this is the more generic statement I would expect to see from issue transitions / status updates.

server/webhook.go
```diff
@@ -287,6 +295,8 @@ func (p *parsedJIRAWebhook) fromChangeLog(issue string) (string, string) {
 	for _, item := range p.ChangeLog.Items {
 		to := item.ToString
 		from := item.FromString
+		lto := strings.ToLower(item.ToString)
```
cpoile: nit: 1/5 I'm worried this will slow down reading the code, can we use longer variable names and prioritize readability over typing speed? (it took me a second to understand the `l` meant lower)

levb: It's gone now. 2/5 https://github.com/golang/go/wiki/CodeReviewComments#variable-names in this case, though `lto` may not have been the best choice, I agree.

server/webhook.go
levb: Done.

PR #96, opened 2019-05-28
MM-15871 strengthened one-time checks for auth links

levb:
The issue was that specific (Jira-side) "connect" links could not be reused, but a user could generate a number of simultaneously valid links. 

For the cloud auth now handle this by storing the mmuserid->secret mapping, and then verifying the secret in the link.

For Server, Jira appears to ignore the secret value, but still protect the link from multiple use by deleting the stored key after its first use.

### DIFF
--- a/server/user_server.go
+++ b/server/user_server.go
@@ -21,20 +21,22 @@ func httpOAuth1Complete(jsi *jiraServerInstance, w http.ResponseWriter, r *http.
 		return http.StatusInternalServerError,
 			errors.WithMessage(err, "failed to parse callback request from Jira")
 	}
-
-	requestSecret, err := jsi.Plugin.LoadOneTimeSecret(requestToken)
-	if err != nil {
-		return http.StatusInternalServerError, err
-	}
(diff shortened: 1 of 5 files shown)


server/user_server.go
```diff
@@ -21,20 +21,22 @@ func httpOAuth1Complete(jsi *jiraServerInstance, w http.ResponseWriter, r *http.
...
 	if err != nil {
-		return http.StatusInternalServerError, err
+		return http.StatusUnauthorized, err
 	}
-	err = jsi.Plugin.DeleteOneTimeSecret(requestToken)
-	if err != nil {
-		return http.StatusInternalServerError, err
+	// Unfortunately, passing the wrong secret to oauth1Config.AccessToken does
+	// not appear to fail as expected, so have to add this extra protection,
+	// making the secret value useless?
```
cpoile: that's hilarious. :)

crspeller:
Taking a look at the security around this more closely.

crspeller:
Replacing with https://github.com/mattermost/mattermost-plugin-jira/pull/110

PR #106, opened 2019-05-30
[MM-15589] Create Issue: (Jira server) Priority field fails to select… #102

cpoile:
… (#81)

The priority field type, inside the fields object, is an object, not a string value.
For the priority ReactSelectSetting component, filter against
this.props.value.id to get the value, not this.props.value which is an
object

Remove default values in initialState object.
Remove special casing for priority field in JiraField component. Note
that fields that have field.allowedValues will return an object.
Otherwise returns a string with is handled when field.schema.type ===
'string'

minor changes

fixup


cpoile:
don't merge, something got merged wrong. Have to test.

hanzei:
Sorry, was to quick with the label.

cpoile:
ok now

PR #108, opened 2019-05-30
[MM- 15838] Handle selected elements in Jira issue creation screen

jfrerich:
Summary
This PR handles two issues, both for custom Jira fields
1) if the custom field is a textarea, mimic that in the create-issue dialog window.
2) Select List (single choice) field.  Values were not showing up in the dropdown.

Ticket Link
Please view the following ticket for a description
https://mattermost.atlassian.net/browse/MM-15838

### DIFF
--- a/webapp/src/components/jira_field.jsx
+++ b/webapp/src/components/jira_field.jsx
@@ -22,6 +22,9 @@ export default class JiraField extends React.PureComponent {
     };
 
     // Creates an option for react-select from an allowedValue from the jira field metadata
+    // includes both .value and .name because allowedValue test cases have used .value or .name
+    // and are mutually exclusive.
+    // should wrap this with some if else logic
     makeReactSelectValue = (allowedValue) => {
         const iconLabel = (
             <React.Fragment>
@@ -29,9 +32,6 @@ export default class JiraField extends React.PureComponent {
                     style={getStyle().jiraIcon}
                     src={allowedValue.iconUrl}
                 />
-                {/* this works because allowedValue test cases have */}
-                {/* been .value or .name and are mutually exclusive  */}
-                {/* should wrap this with some if else logic */}
                 {allowedValue.value}
                 {allowedValue.name}
             </React.Fragment>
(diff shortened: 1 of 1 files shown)


webapp/src/components/jira_field.jsx
```diff
@@ -29,6 +29,10 @@ export default class JiraField extends React.PureComponent {
                     style={getStyle().jiraIcon}
                     src={allowedValue.iconUrl}
                 />
+                {/* this works because allowedValue test cases have */}
```
cpoile: nit: Can we put the comments above the fragment? I dislike this comment style. It looks like you're passing comments into the component. :)

jfrerich:
@jasonblais You are confusing the `jira_field.jsx` file with `jira_fields.jsx`.  I only changed the `jira_field.jsx` file so if that is the file you were looking at, it wouldn't have this code, but `jira_fields.jsx` will  :) 

https://github.com/mattermost/mattermost-plugin-jira/blob/2c4e3a1defbe1857e4b88b9b6dfe9352c82e79b6/webapp/src/components/jira_fields.jsx#L30

jasonblais:
@jfrerich got it - do we need to make any changes to `jira_fields.jsx` so that all the fields we support (multi-text area, etc.) show up in the dialog? 

Or would this be part of your error handling ticket (https://mattermost.atlassian.net/browse/MM-15924)?

webapp/src/components/jira_field.jsx
jfrerich: 0/5.  We probably have a MM doc style guide and will check this after pushing through with these changes!

jfrerich:
@jasonblais I am going to address that issue next.  We have two issues left regarding fields.

1) handle the case where Jira instance requires fields that we do not yet support (and send an ephemeral message in MM)
2) handle the case where a supported field is not filled out when user hits the create button. (and let show them on the create issue dialog window)

With respect to `jira_fields.jsx`, I will have to edit it.  My initial thought is that I will maintain a list of field types we DO support.  Then compare all visible jira fields to that list and if a field doesn't fall under one of those, I'll error out and report back in MM.

jasonblais:
Sounds good, thanks @jfrerich!

PR #109, opened 2019-05-30
MM-15249: cache the current Jira instance

levb:
The default TTL is set to 1s.

Ticket: https://mattermost.atlassian.net/browse/MM-15249

### DIFF
--- a/server/kv.go
+++ b/server/kv.go
@@ -230,8 +230,10 @@ func (p *Plugin) LoadCurrentJIRAInstance() (Instance, error) {
 	now := time.Now()
 
 	if now.Before(conf.currentInstanceExpires) {
+		// if conf.currentInstanceExpires is set and there is no current
+		// instance, it's a cached "Not found"
 		if conf.currentInstance == nil {
-			return nil, errors.New("not found")
+			return nil, errors.New("failed to load current Jira instance: not found")
 		}
 		return conf.currentInstance, nil
 	}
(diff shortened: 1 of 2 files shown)


server/plugin.go
```diff
@@ -29,12 +30,19 @@ type externalConfig struct {
...
+const currentInstanceTTL = 1 * time.Second
+
 type config struct {
 	// externalConfig caches values from the plugin's settings in the server's config.json
 	externalConfig
 
 	// Cached actual bot user ID (derived from c.UserName)
 	botUserID string
+
+	// Cached current Jira instance. A non-0 expires indicates the presense
```
jfrerich: nit: s/presense/presence/

server/plugin.go
```diff
@@ -29,12 +30,19 @@ type externalConfig struct {
 	Secret string `json:"secret"`
 }
 
+const currentInstanceTTL = 1 * time.Second
```
jfrerich: What does multiplying by one achieve?

levb: Just a no-op to make he code more readable.

crspeller: Why 1 second?

server/kv.go
```diff
@@ -208,17 +214,36 @@ func (p *Plugin) DeleteJiraInstance(key string) (returnErr error) {
...
 
 	return nil
 }
 
 func (p *Plugin) LoadCurrentJIRAInstance() (Instance, error) {
+	conf := p.getConfig()
+	now := time.Now()
+
+	if now.Before(conf.currentInstanceExpires) {
+		if conf.currentInstance == nil {
```
crspeller: Why would this happen? Maybe just check `conf.currentInstance == nil` above and let it try to load it if it's nil?

PR #110, opened 2019-05-31
MM-15900 Improving connection flows.

crspeller:
- Prevents multiple sanctimonious valid connect links.
- Prevents am SSRF in the jira server connect flow. 
- Based on and replaces: https://github.com/mattermost/mattermost-plugin-jira/pull/96

### DIFF
--- a/server/user_cloud.go
+++ b/server/user_cloud.go
@@ -71,10 +71,20 @@ func httpACUserInteractive(jci *jiraCloudInstance, w http.ResponseWriter, r *htt
 			Name: username,
 		},
 	}
-	mattermostUserId, secret, err := jci.Plugin.ParseAuthToken(mmToken)
+	mattermostUserId := r.Header.Get("Mattermost-User-ID")
+	if mattermostUserId == "" {
+		return http.StatusUnauthorized, errors.New("not authorized")
+	}
+
+	requestedUserId, secret, err := jci.Plugin.ParseAuthToken(mmToken)
 	if err != nil {
 		return http.StatusUnauthorized, err
 	}
+
+	if mattermostUserId != requestedUserId {
+		return http.StatusUnauthorized, errors.New("not authorized, user id does not match link")
+	}
+
 	mmuser, appErr := jci.Plugin.API.GetUser(mattermostUserId)
 	if appErr != nil {
 		return http.StatusInternalServerError,
@@ -83,19 +93,14 @@ func httpACUserInteractive(jci *jiraCloudInstance, w http.ResponseWriter, r *htt
 
 	switch r.URL.Path {
 	case routeACUserConnected:
-		value := ""
-		value, err = jci.Plugin.LoadOneTimeSecret(secret)
+		storedSecret := ""
+		storedSecret, err = jci.Plugin.LoadOneTimeSecret(mattermostUserId)
 		if err != nil {
 			return http.StatusUnauthorized, err
 		}
-		err = jci.Plugin.DeleteOneTimeSecret(secret)
-		if err != nil {
-			return http.StatusInternalServerError, err
+		if len(storedSecret) == 0 || storedSecret != secret {
+			return http.StatusUnauthorized, errors.New("this link has already been used")
 		}
-		if len(value) == 0 {
-			return http.StatusUnauthorized, errors.New("link expired")
-		}
-
 		err = jci.Plugin.StoreUserInfoNotify(jci, mattermostUserId, uinfo)
 
 	case routeACUserDisconnected:
--- a/server/user_server.go
+++ b/server/user_server.go
@@ -15,20 +15,16 @@ import (
 	"github.com/mattermost/mattermost-server/model"
 )
 
+type OAuth1aTemporaryCredentials struct {
+	Token  string
+	Secret string
+}
+
 func httpOAuth1Complete(jsi *jiraServerInstance, w http.ResponseWriter, r *http.Request) (int, error) {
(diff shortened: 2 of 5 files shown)


server/user_server.go
```diff
@@ -41,13 +37,24 @@ func httpOAuth1Complete(jsi *jiraServerInstance, w http.ResponseWriter, r *http.
 			errors.WithMessage(appErr, "failed to load user "+mattermostUserId)
 	}
 
+	oauthTmpCredentials, err := jsi.Plugin.OneTimeLoadOauth1aTemporaryCredentials(mattermostUserId)
+	if err != nil || oauthTmpCredentials == nil || len(oauthTmpCredentials.Token) <= 0 {
+		return http.StatusInternalServerError, errors.WithMessage(err, "failed to get temporary credentials for "+mattermostUserId)
+	}
+
+	if oauthTmpCredentials.Token != requestToken {
```
crspeller: This check prevents reuse (since the credentials are deleted) and also makes sure the MM user that started the flow is the one finishing it (keyed on userid)

server/user_server.go
```diff
@@ -41,13 +37,24 @@ func httpOAuth1Complete(jsi *jiraServerInstance, w http.ResponseWriter, r *http.
...
 	oauth1Config, err := jsi.GetOAuth1Config()
 	if err != nil {
 		return http.StatusInternalServerError,
 			errors.WithMessage(err, "failed to obtain oauth1 config")
 	}
 
-	accessToken, accessSecret, err := oauth1Config.AccessToken(requestToken, requestSecret, verifier)
+	// Although we pass the oauthTmpCredentials as required here. The JIRA server does not appar to validate it.
+	// We perform the check above for reuse so this is irrelavent to the security from our end.
+	accessToken, accessSecret, err := oauth1Config.AccessToken(requestToken, oauthTmpCredentials.Secret, verifier)
```
crspeller: This I don't quite understand, but I don't think it compromises the security.

server/user_cloud.go
```diff
@@ -83,19 +93,14 @@ func httpACUserInteractive(jci *jiraCloudInstance, w http.ResponseWriter, r *htt
...
 		if err != nil {
 			return http.StatusUnauthorized, err
 		}
-		err = jci.Plugin.DeleteOneTimeSecret(secret)
-		if err != nil {
-			return http.StatusInternalServerError, err
-		}
-		if len(value) == 0 {
-			return http.StatusUnauthorized, errors.New("link expired")
+		if len(storedSecret) == 0 || storedSecret != secret {
```
crspeller: Keyed secret on mmuserid deleted on retrieval prevents reuse.

server/user_cloud.go
```diff
@@ -71,10 +71,20 @@ func httpACUserInteractive(jci *jiraCloudInstance, w http.ResponseWriter, r *htt
...
+	if mattermostUserId == "" {
+		return http.StatusUnauthorized, errors.New("not authorized")
+	}
+
+	requestedUserId, secret, err := jci.Plugin.ParseAuthToken(mmToken)
 	if err != nil {
 		return http.StatusUnauthorized, err
 	}
+
+	if mattermostUserId != requestedUserId {
```
crspeller: Checks to make sure the user starting the flow is the one finishing it. (otherwise the link is portable to unauthorized browsers) Not a huge deal, but nice to prevent.

levb: Well, I gave it thumbs up, then remembered why I didn't do it in the first place, tested it and I doesn't work on my server.

 On my server (with ngrok) I've never seen `Mattermost-User-ID` come through for these requests, always blank. I've been assuming that we rely on explicit request params/headers rather than cookies, and that running the iframe inside Jira somehow messed it up. Probably worth investigating - it'd simplify the flow a lot if we could MM-authenticate these requests as you proposed.

cpoile: I wonder if it's stripped by the server, just like the Authorization header? And if so, can we maybe change that in server 5.14 (for the future)?

If we don't get `Mattermost-User-ID`, can we really secure this from the portability attack you mentioned? Is that a serious problem?

crspeller: This works for me though. It should because the request back to the MM server should include the authentication cookie.

levb: How do you expose your server to Jira, direct NAT?

crspeller: Using a server on AWS.
@levb @cpoile Can ether of you reproduce the problem with these changes?

cpoile: I just tried it. Disconnected and reconnected to the cloud mmtest instance. No problems at all through my ngrok. Works!

levb: Tested some more. It works for me in Safari, does not in Chrome. There is no cookie coming in the request from Chrome. Probably has something to do with my browser settings, which I don't think I customized much. @cpoile @crspeller

levb: Had "Block Third Party Cookies" turned on. Turning it off makes the use-case work. I am not sure what the default in Chrome is, looking.

levb: Confirmed that Chrome does _not_ block by default, so I guess it's ok.

We should maybe add copy to the error page that would recommend checking the browser settings? @jasonblais?

jasonblais: @levb @crspeller So if someone blocks third party cookies on Chrome, they won't be able to successfully connect their account to Jira? Is that correct?

That seems like a potentially bad experience? Do we have no way of avoiding that?

levb: Filed https://mattermost.atlassian.net/browse/MM-15968 to extend (and style?) the error page/message.

PR #111, opened 2019-06-01
MM-15922 - Fixing theme and dropdown issue

asaadmahmood:
**Ticket:**
https://mattermost.atlassian.net/browse/MM-15922

I've also styled the Jira issue selector but I'm basically using the same style function as in the `react_select_setting.jsx` file. Would be ideal if we can just have a separate utils.jsx file from where we can import the function in different places.

Screenshot 2019-06-01 at 2 58 52 PM

Screenshot 2019-06-01 at 2 17 21 PM

### DIFF
--- a/webapp/src/components/react_select_setting.jsx
+++ b/webapp/src/components/react_select_setting.jsx
@@ -14,12 +14,12 @@ export default class ReactSelectSetting extends React.PureComponent {
     static propTypes = {
         name: PropTypes.string.isRequired,
         onChange: PropTypes.func,
-        theme: PropTypes.object,
+        theme: PropTypes.object.isRequired,
     };
 
     handleChange = (value) => {
         if (this.props.onChange) {
-            if (Array.isArray(value)) {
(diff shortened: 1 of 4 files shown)


webapp/src/components/react_select_setting.jsx
```diff
@@ -83,8 +91,9 @@ export default class ReactSelectSetting extends React.PureComponent {
             >
                 <ReactSelect
                     {...this.props}
-                    styles={this.props.theme && this.getStyle(this.props.theme)}
+                    menuPortalTarget={document.body}
                     onChange={this.handleChange}
+                    styles={this.getStyle(this.props.theme)}
```
cpoile: I think the null check belongs in the `getStyle` function above, as I noted earlier, so it's a good change to remove it from here. 👍

cpoile:
@asaadmahmood refactored for you. There's still some non-resolved comments though.

asaadmahmood:
@cpoile I think that was accidental.

PR #112, opened 2019-06-01
[MM-15962] Allow only handled fields to show in Create Issue Dialog

jfrerich:
**Summary** 
This ticket is to reduce the fields that show up in the create issue dialog window to only those currently supported.

- Allow fields that are not required to show up in create issue dialog
- restrict custom field types to textarea, textfield, select (single-select), epic-label (for epic issue types), and project picker
- for non-custom types, with allowedValues, don't show fields with field.schema.type == array.  Fix at later date

**Ticket Link**
https://mattermost.atlassian.net/browse/MM-15962

### DIFF
--- a/webapp/src/components/jira_field.jsx
+++ b/webapp/src/components/jira_field.jsx
@@ -44,6 +44,17 @@ export default class JiraField extends React.PureComponent {
     render() {
         const field = this.props.field;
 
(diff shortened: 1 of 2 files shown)


jfrerich:
@jasonblais including you for comments regarding ticket

PR #116, opened 2019-06-03
[MM-15880] Replaced the S3 icon used for command ephemeral posts

levb:
- Fixed https://mattermost.atlassian.net/browse/MM-15880
- Also removed dead code (was tangentially related)

@asaadmahmood I ended up taking the PNG icon because the SVG was huge (125Kb?)

### DIFF
--- a/server/command.go
+++ b/server/command.go
@@ -58,75 +58,75 @@ func (ch CommandHandler) Handle(p *Plugin, c *plugin.Context, header *model.Comm
 }
 
 func commandHelp(p *Plugin, c *plugin.Context, header *model.CommandArgs, args ...string) *model.CommandResponse {
-	return help()
+	return help(p)
 }
 
-func help() *model.CommandResponse {
-	return responsef(helpText)
+func help(p *Plugin) *model.CommandResponse {
+	return responsef(p, helpText)
 }
 
 func (p *Plugin) ExecuteCommand(c *plugin.Context, commandArgs *model.CommandArgs) (*model.CommandResponse, *model.AppError) {
 	args := strings.Fields(commandArgs.Command)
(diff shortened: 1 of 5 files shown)


server/utils.go
```diff
@@ -1,97 +1,11 @@
...
 	"net/url"
-	"regexp"
 
 	"github.com/andygrunwald/go-jira"
-	"github.com/pkg/errors"
-
-	"github.com/mattermost/mattermost-server/model"
 )
 
-func (p *Plugin) CreateBotDMPost(userId, message, postType string) (returnErr error) {
```
jasonblais: Are all of these changes related to replacing the S3 icon?

server/command.go
```diff
@@ -399,12 +399,13 @@ func getCommand() *model.Command {
...
 
-func responsef(format string, args ...interface{}) *model.CommandResponse {
+func responsef(p *Plugin, format string, args ...interface{}) *model.CommandResponse {
+	conf := p.getConfig()
 	return &model.CommandResponse{
 		ResponseType: model.COMMAND_RESPONSE_TYPE_EPHEMERAL,
 		Text:         fmt.Sprintf(format, args...),
-		Username:     PluginMattermostUsername,
-		IconURL:      PluginIconURL,
+		Username:     conf.UserName,
```
jasonblais: Does this mean it's now configurable?

levb: Yes, it will use the same configured username wee use for webhook posts. I assumed this was desirable, right?

server/utils.go
levb: No, but this function used thee icon, etc.

levb:
For 2.1 we can just serve it from the /public folder, and all the responsef's will be replaced anyway, @cpoile

PR #118, opened 2019-06-04
MM-15893: Fixed Jira Server webhooks

levb:
Ticket: https://mattermost.atlassian.net/browse/MM-15893. I did not observe failures with anything other than comments, which I fixed.

Jira Server and Jira Cloud send different webhooks for comments. 
- Jira Cloud sends 1 webhook (A) for each comment event, with `WebhookEvent == "comment_xxx"`. The event contains all necessary issue data for us to post to Mattermost. 
- Jira Server sends 2 webhooks for each comment change. Like Jira Cloud, it sends a (B) `WebhookEvent == "comment_xxx"` event, but it is stripped of all issue data. It also sends a (C) `WebhookEvent == "jira:issue_updated" && IssueEventTypeName == "comment_xxx"` which are complete, like (A).

For our context, (B) is insufficient, so it now returns `ErrWebhookIgnored` and results in a no-op. (C) is now supported, but it lacks the separate admin controls on the Jira Webhook Admin screen.

@jasonblais please consider removing `maskComments` from https://github.com/mattermost/mattermost-plugin-jira/blob/6bea27e47a2df568389ff494a413328d0f90909e/server/webhook.go#L51-L53 to disable comment event processing by default.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ vendor
 
 # Intellij
 .idea/
+
+.vscode/
--- a/server/webhook.go
+++ b/server/webhook.go
@@ -11,9 +11,12 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"os"
 	"strings"
+	"time"
 
 	"github.com/andygrunwald/go-jira"
+	"github.com/pkg/errors"
 
 	"github.com/mattermost/mattermost-server/model"
 )
@@ -105,6 +108,8 @@ type notifier interface {
 	notify(ji Instance, parsed *parsedJIRAWebhook, text string)
 }
 
+var ErrWebhookIgnored = errors.New("Webhook purposely ignored")
+
 func httpWebhook(p *Plugin, w http.ResponseWriter, r *http.Request) (int, error) {
 	if r.Method != http.MethodPost {
 		return http.StatusMethodNotAllowed,
@@ -152,6 +157,9 @@ func httpWebhook(p *Plugin, w http.ResponseWriter, r *http.Request) (int, error)
 	}
 
 	parsed, err := parse(r.Body, nil)
+	if err == ErrWebhookIgnored {
+		return http.StatusOK, nil
+	}
 	if err != nil {
 		return http.StatusBadRequest, err
 	}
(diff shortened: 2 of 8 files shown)


jasonblais:
> Please consider removing maskComments from

Is the only consequence that in Jira Server, comments created/updated/deleted events won't be sent when `Issue:Updated` is enabled for webhook notifications? 

Do the attached `Comments:Created/Updated/Deleted` not work on Jira Server?
image

levb:
@jasonblais updated the description for clarity, hopefully answering your questions. Basically, in Jira Server, the separately-controlled comment events have insufficient data for our architecture. With this PR we support comment change notifications, but based solely on issue update events.

.gitignore
```diff
@@ -9,3 +9,5 @@ vendor
 
 # Intellij
 .idea/
+
+.vscode/
```
cpoile: For future reference, I realized I could stop adding `.idea/` to all our public repos, and instead add a `.gitignore` file in the `.idea/` directory, containing:
```
*.*
*/**/*.*
```

server/webhook.go
```diff
@@ -152,6 +157,9 @@ func httpWebhook(p *Plugin, w http.ResponseWriter, r *http.Request) (int, error)
 	}
 
 	parsed, err := parse(r.Body, nil)
+	if err == ErrWebhookIgnored {
```
cpoile: nice 👍

.gitignore
hanzei: Or you use your local `.gitignore` file

levb: kk, will take it out

jasonblais:
Thanks @levb, that's helpful. It's definitely not optimal, but not much we can do given the issue is on Jira's end.

For Jira Server only, propose (as you proposed)
1. Let's remove `maskComments` from the defaults for issue updated events.
2. Let's add it as a parameter, so a Jira Admin can add `& maskComments=1` as part of the webhook URL to send these events
3. Let's not support CommentCreated/CommentUpdated/CommentDeleted events at all if it's stripped of all issue data.

No changes to Jira Cloud.

PR #119, opened 2019-06-04
[MM-15913] Update UX when creating issues for projects that require additional fields

jfrerich:
**Summary** 
This ticket addresses the issue when a user tries to create a ticket and the connected Jira instance requires fields we currently don't support.

I'm using the Draft Pull Request feature because there are some steps still remaining, but I would like to receive early feedback due to the time crunch for Jira2.0 

**Description of enhancements made for UX (from Ticket)**

Update UX when creating issues for projects that require additional fields

step 1: allow the user to fill in the parameters we support in the Create Jira Issue dialog
step 2: allow user to hit "Create"
step 3: respond back with an ephemeral message saying

>The project you tried to create an issue for has **required fields** this plugin does not yet support. Please [create your Jira issue manually](...link to Create Jira Issue modal).

step 4: clicking the link brings the user to an issue creation screen in Jira, with the fields they entered previously pre-filled

**NOTES for QA** 
Cases that should be covered following this ticket are listed below:

*Jira Default Fields*
- Project
- Issue Type
- Priority
- Description
- Summary

*Jira Custom Fields*
- Project Picker
- Select List (single choice)
- Text Field (multi-line)
- Text Field (single line)

- Given a field that is not supported, message in the description section above is posted in MM

@jasonblais I would really like to have this version spun up for people to thoroughly test.

**Ticket Link**
Please view the following ticket for a description
https://mattermost.atlassian.net/browse/MM-15913

### DIFF
--- a/server/issue.go
+++ b/server/issue.go
@@ -111,7 +111,7 @@ func httpAPICreateIssue(ji Instance, w http.ResponseWriter, r *http.Request) (in
 			ChannelId: post.ChannelId,
 			RootId:    rootId,
 			ParentId:  parentId,
-			UserId:    mattermostUserId,
+			UserId:    ji.GetPlugin().conf.botUserID,
 		}
 		_ = api.SendEphemeralPost(mattermostUserId, reply)
 
--- a/server/issue.go
+++ b/server/issue.go
@@ -111,7 +111,7 @@ func httpAPICreateIssue(ji Instance, w http.ResponseWriter, r *http.Request) (in
 			ChannelId: post.ChannelId,
 			RootId:    rootId,
 			ParentId:  parentId,
-			UserId:    ji.GetPlugin().conf.botUserID,
+			UserId:    ji.GetPlugin().getConfig().botUserID,
 		}
 		_ = api.SendEphemeralPost(mattermostUserId, reply)
 
--- a/server/issue.go
+++ b/server/issue.go
@@ -12,7 +12,6 @@ import (
 	"strings"
 
 	"github.com/andygrunwald/go-jira"
-	"github.com/google/go-querystring/query"
 	"github.com/pkg/errors"
 
 	"github.com/mattermost/mattermost-server/model"
@@ -83,16 +82,29 @@ func httpAPICreateIssue(ji Instance, w http.ResponseWriter, r *http.Request) (in
 		parentId = create.PostId
 	}
 
+	issue := &jira.Issue{
+		Fields: &create.Fields,
+	}
+
+	project, resp, err := jiraClient.Project.Get(issue.Fields.Project.Key)
+	if err != nil {
+		message := "failed to get the project, postId: " + create.PostId
+		if resp != nil {
+			bb, _ := ioutil.ReadAll(resp.Body)
+			resp.Body.Close()
+			message += ", details:" + string(bb)
+		}
+
+		return http.StatusInternalServerError, errors.WithMessage(err, message)
+	}
+
 	if len(create.RequiredFieldsNotCovered) > 0 {
 
-		v, _ := query.Values(create.Fields)
+		req := buildCreateQuery(ji, project, issue)
 
 		message := "The project you tried to create an issue for has **required fields** this plugin does not yet support. "
-		url := fmt.Sprintf("%v/secure/CreateIssueDetails!init.jspa", ji.GetURL())
-		query := "pid=10000&issuetype=10004"
-
 		reply := &model.Post{
-			Message:   fmt.Sprintf("%v Please [create your Jira issue manually](%v?%v&%v).", message, url, query, v.Encode()),
+			Message:   fmt.Sprintf("%v Please [create your Jira issue manually](%v).", message, req.URL.String()),
 			ChannelId: post.ChannelId,
 			RootId:    rootId,
 			ParentId:  parentId,
@@ -103,12 +115,13 @@ func httpAPICreateIssue(ji Instance, w http.ResponseWriter, r *http.Request) (in
 			return http.StatusInternalServerError,
 				errors.WithMessage(appErr, "failed to create notification post "+create.PostId)
 		}
+
+		w.Header().Set("Content-Type", "application/json")
+		fmt.Fprintf(w, "{}")
 		return http.StatusOK, nil
 	}
 
-	created, resp, err := jiraClient.Issue.Create(&jira.Issue{
-		Fields: &create.Fields,
-	})
+	created, resp, err := jiraClient.Issue.Create(issue)
 	if err != nil {
 		message := "failed to create the issue, postId: " + create.PostId
 		if resp != nil {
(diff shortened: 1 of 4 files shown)


server/issue.go
```diff
@@ -26,8 +27,9 @@ func httpAPICreateIssue(ji Instance, w http.ResponseWriter, r *http.Request) (in
 	api := ji.GetPlugin().API
 
 	create := &struct {
-		PostId string           `json:"post_id"`
-		Fields jira.IssueFields `json:"fields"`
+		RequiredFieldsNotCovered []string         `json:"required_fields_not_covered"`
```
jfrerich: Empty array if all required fields are covered by JiraFields.  Else we are going to respond with MM post with link to manually create Jira Ticket (prefilled with fields that were entered by user)

server/issue.go
```diff
@@ -73,6 +75,37 @@ func httpAPICreateIssue(ji Instance, w http.ResponseWriter, r *http.Request) (in
...
+		parentId = create.PostId
+	}
+
+	if len(create.RequiredFieldsNotCovered) > 0 {
+
+		v, _ := query.Values(create.Fields)
+
+		message := "The project you tried to create an issue for has **required fields** this plugin does not yet support. "
+		url := fmt.Sprintf("%v/secure/CreateIssueDetails!init.jspa", ji.GetURL())
+		query := "pid=10000&issuetype=10004"
```
jfrerich: hardcoded pid and issuetype, but will be sent from client.  I was really hoping to borrow this google/go-querystring module which will fill a query string from a struct, but there are issues with capitalization and other slight changes that will need to be handled.

server/issue.go
```diff
@@ -73,6 +75,37 @@ func httpAPICreateIssue(ji Instance, w http.ResponseWriter, r *http.Request) (in
...
+		// the original post was a reply
+		rootId = post.RootId
+		parentId = create.PostId
+	}
+
+	if len(create.RequiredFieldsNotCovered) > 0 {
+
+		v, _ := query.Values(create.Fields)
+
+		message := "The project you tried to create an issue for has **required fields** this plugin does not yet support. "
```
crspeller: Since we are sending the fields not covered up, why not tell them which fields where not covered?

jfrerich: I thought about doing that and sent the values instead of true/false for this reason.  Turns out the linked Jira page states at the top which fields are required (that weren't handled).  So I was debating where to add the fields to the post.   How about I cut the PR non-drafted PR and then add this feature.  I'm still working out some issues. 

Another good reason to add this enhancement would be so users have a traceable message listing fields we currently don't handle and they can propagate that info up to management and us.

jfrerich: @jasonblais This is a good request from @crspeller.  I'm curious what you would like the final message to look like?
> The project you tried to create an issue for has **required fields** this plugin does not yet support. 
> Please [create your Jira issue manually](permalink)
> Unsupported Fields:
>   - FieldName 1
>  - FieldName 2

jasonblais: If this is a low risk change, I'm all for it. In this case we can bring the action first, so the list of unsupported fields flows naturally from the second sentence.

```
[Please create your Jira issue manually](permalink). The project you tried to create an issue for has these **required fields** the plugin does not yet support:
- FieldName 1
- FieldName 2
- ...
```

/cc @asaadmahmood if you have feedback on the above proposed message

jfrerich:
@jasonblais The ephemeral message is a would have been a good idea for my testing.  I'm got >100 messages in my test channel :)

jasonblais:
@jfrerich By the way, for point 1, this may be out of scope for this PR so feel free to create a separate issue for it.

jfrerich:
@jasonblais I will move 1 to separate ticket to keep this one moving along.  Thanks for catching the issue!

server/issue.go
asaadmahmood: Looks fine for me, especially if its just a concern case, which we may in the future would further reduce (by supporting additional fields too).

server/issue.go
levb: I'd be wary of the client code relying on an empty array vs null. Will look there.

jasonblais:
> I will move 1 to separate ticket to keep this one moving along.

Certainly, ticket for 1 here https://mattermost.atlassian.net/browse/MM-16116

server/issue.go
jasonblais: @jfrerich This can also be filed as a separate improvement for later. The only blocking change from PM side is turning the message to be ephemeral

jfrerich: Thanks for the feedback, J.  I'm finishing my RC testing right now and will get this done right after.

server/issue.go
jfrerich: @lev Let me clarify and see if this answers your concern.  

`RequiredFieldsNotCovered` will be an empty array, in the event all Required fields were covered by our code.  If some required fields aren't covered, this array will be filled with the names of the fields and will be displayed back the user in an ephemeral post.

Does clear up your concern or am I still missing the issue?

levb: The question was, really, why {}, not `nil`?

jfrerich:
@jasonblais For the ephemeral message, how does this look?

image

jasonblais:
@jfrerich that looks great to me 👍

jfrerich:
@jasonblais one more question regarding the ephemeral post.  Should the `System` name and icon be replaced with Jira?

jasonblais:
@jfrerich Good catch, preferably we would use the same icon/username as for `/jira help` and other such commands

jfrerich:
@jasonblais How about this?

image

server/issue.go
```diff
@@ -111,7 +111,7 @@ func httpAPICreateIssue(ji Instance, w http.ResponseWriter, r *http.Request) (in
 			ChannelId: post.ChannelId,
 			RootId:    rootId,
 			ParentId:  parentId,
-			UserId:    mattermostUserId,
+			UserId:    ji.GetPlugin().conf.botUserID,
```
jfrerich: @levb Correct method to retrieve UserId for Plugin?

jasonblais:
@jfrerich That looks good! Assuming `jira_v2` is just a user selected in System Console > Jira page, or entered in config.json :) 

Minor nit: There's an extra period after the colon - `the plugin does not yet support:.`

jfrerich:
@jasonblais period removed.  One question left for @levb and we are ready to merge.

server/issue.go
levb: should use p.getConfig()

crspeller: No. You need to use the `getConfig` method.

jfrerich:
@crspeller getConfig() issue is addressed in latest commit.

levb:
build is failing @jfrerich

jfrerich:
Looking at it!

jfrerich:
> build is failing @jfrerich

@levb passing now.  I had a console.log() left in one of the .jsx files

PR #129, opened 2019-06-05
[MM-15932] In Jira Connect approval page, display a better name

cpoile:
#### Summary
- Instead of displaying nickname, display "fullname (@username)" or "@username" if user has no fullname set.
- This makes it easier for the user to understand what is being connected.

#### Links
[MM-15932](https://mattermost.atlassian.net/browse/MM-15932)

### DIFF
--- a/server/user_cloud.go
+++ b/server/user_cloud.go
@@ -4,13 +4,14 @@
 package main
 
 import (
-	"github.com/mattermost/mattermost-server/model"
 	"net/http"
 	"path"
 
 	"github.com/andygrunwald/go-jira"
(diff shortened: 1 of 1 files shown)


server/user_cloud.go
```diff
@@ -4,14 +4,13 @@
 package main
 
 import (
+	"github.com/mattermost/mattermost-server/model"
```
levb: nit: I think we have the convention of putting all MM imports at the bottom

cpoile: Thanks for pointing that out -- I think that was automatic. I've added go imports to my global file watchers now.

PR #134, opened 2019-06-06
[MM-15986] Check if user does not have permissions to create issues

jfrerich:
**Summary**
If a user does not have permissions to create issues, alert them in the create issue dialog window.  Previously, the dialog window hung in the "loading" state.

*New Dialog Window*
- Message added
- Cancel and Create buttons changed to singe Close button.

image

**Testing** 
Tested on sales-jira instance

**Repro**
*disable create issue permissions*
- Jira Administration > Issues > Permission schemes > Default software scheme > Permissions
- Create Issues > Edit > Remove 
- Removes Application access for Any logged in user

*reenable access*
- Permission > Add `Create Issues` label
- Granted to > Application access > Any logged in user

**Ticket**
Please view the following ticket for additional description
https://mattermost.atlassian.net/browse/MM-15986

### DIFF
--- a/server/issue.go
+++ b/server/issue.go
@@ -185,6 +185,13 @@ func httpAPIGetCreateIssueMetadata(ji Instance, w http.ResponseWriter, r *http.R
 	}
 
 	w.Header().Set("Content-Type", "application/json")
+
+	if len(cimd.Projects) == 0 {
+		fmt.Fprintf(w, "{\"error\": \"You do not have permission to create issues in any projects. Please contact your Jira admin.\"}")
+		return http.StatusInternalServerError,
+			errors.WithMessage(err, "failed to marshal response")
+	}
+
 	b, err := json.Marshal(cimd)
 	if err != nil {
 		return http.StatusInternalServerError,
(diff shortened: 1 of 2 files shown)


server/issue.go
```diff
@@ -185,6 +185,13 @@ func httpAPIGetCreateIssueMetadata(ji Instance, w http.ResponseWriter, r *http.R
 	}
 
 	w.Header().Set("Content-Type", "application/json")
+
+	if len(cimd.Projects) == 0 {
+		fmt.Fprintf(w, "{\"error\": \"You do not have permission to create issues in any projects. Please contact your Jira admin.\"}")
+		return http.StatusInternalServerError,
+			errors.WithMessage(err, "failed to marshal response")
```
crspeller: Copy paste error: `"failed to marshal response"`

levb: @jfrerich @crspeller I went ahead to fix the cut&paste but then noticed that as far as I can tell we should be returning a 200 OK here, not a 500, so I changed that too. Don't have an easy way of testing it though.

jfrerich: Thanks @levb. I had to research the usage of writing directly with Fprintf vs saving to var bb of type []byte.

PR #137, opened 2019-06-06
[MM-16094] Fix for:  Clicking post menu for a system message crashes the app

cpoile:
#### Summary
- Problem: When Jira plugin is enabled, clicking on "More Actions" (...) post menu for a system message crashes the app
- Fix: Certain types of system messages are bundled together, and we weren't testing for that. Trying to find the post of that bundle returned an undefined post, and then we tried to do stuff with that. = crash.  Thank you, Javascript. 

#### Links
[MM-16094](https://mattermost.atlassian.net/browse/MM-16094)


webapp/src/components/utils.jsx
```diff
@@ -64,3 +64,9 @@ export const getStyleForReactSelect = (theme) => {
         }),
     };
 };
+
+// TODO: change to use the isCombinedUserActivityPost in 'mattermost-redux/utils/post_list' when we upgrade to 5.12
```
levb: (non-blocking) can I get an explanation for this please?

webapp/src/components/utils.jsx
```diff
@@ -64,3 +64,9 @@ export const getStyleForReactSelect = (theme) => {
...
+export const isCombinedUserActivityPost = (id) => {
```
crspeller: I know it's not related to this PR, but this utils file is in a bad place. Can we move this to a utils folder and split these up into logically named files? I think we should avoid an explosion of stuff in a generic utils.jsx like the webapp.

webapp/src/components/utils.jsx
cpoile: @levb I just assumed we should use the original function from mattermost-redux (the one here is a cut and paste). And since it was added pretty recently in mattermost-redux, we probably can't assume an older version has it. That's all. Is there a better way to do this?

webapp/src/components/utils.jsx
crspeller: also then we can import `utils/bla.jsx` instead of `../../`

cpoile: @crspeller good points, done 👍

PR #140, opened 2019-06-07
No ticket: /jira command to open the URL instead of displaying the link

levb:
Found the way to open a browser window directly from CommandResponse


levb:
It's already in my big refactor for 2.1, feel free to close this. Having said that, I don't see much risk, the support for `GotoLocation` goes back a year+, can track it down more precisely. Feel free to close if you are not comfortable with this, @jasonblais

PR #142, opened 2019-06-07
[MM-16129] fixed CSRF issue for Desktop app

cpoile:
#### Summary
- Need to use the doFetch helper or else the Mattermost-User-Id will be stripped and we'll get an unauthorized response
- Unrelated: cleaned up a proptype that was failing

#### Links
- fixes: [MM-16129](https://mattermost.atlassian.net/browse/MM-16129)


server/issue.go
```diff
@@ -11,7 +11,7 @@ import (
 	"net/http"
 	"strings"
 
-	"github.com/andygrunwald/go-jira"
+	jira "github.com/andygrunwald/go-jira"
```
cpoile: For some reason gofmt is forcing this -- any idea why now all of a sudden? Will it break anything?

levb: 10/10 a nit, but why? `go-jira` resolves as `jira`, right? Intention-revealing?

cpoile: @levb gofmt (or go imports)'s choice. Why now, is my question...

webapp/src/client/index.js
```diff
@@ -9,6 +9,10 @@ export const doFetch = async (url, options) => {
     return data;
 };
 
+export const doFetchPromise = (url, options) => {
```
crspeller: What is this for?

server/issue.go
levb: not at all, I kinda like it

webapp/src/client/index.js
cpoile: It's so I can keep the debouncing logic in the component, instead of making a debounced version in the client helper. Of course I'm totally open to a way to improve it.

PR #143, opened 2019-06-07
[MM-16097] Fixing: Attach to Jira Issue fails silently when user lacks permissions

cpoile:
#### Summary
- When a user has access to browse a project but does not have access to add comments to the issue, Attach to Jira Issue fails silently
- Also some clean-up:
  - some proptypes were marked as required but weren't actually required causing red in the console
  - forgot to remove the project props in the attach-to-issue search (we're not filtering by project anymore)

UX:

image

#### Links
[MM-16097](https://mattermost.atlassian.net/browse/MM-16097)

### DIFF
--- a/server/issue.go
+++ b/server/issue.go
@@ -7,7 +7,6 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"net/http"
 	"strings"
@@ -372,21 +371,8 @@ func httpAPIAttachCommentToIssue(ji Instance, w http.ResponseWriter, r *http.Req
 	commentAdded, _, err := jiraClient.Issue.AddComment(attach.IssueKey, &jiraComment)
 	if err != nil {
 		if strings.Contains(err.Error(), "you do not have the permission to comment on this issue") {
-			reply := &model.Post{
-				Message:   "You do not have permission to create a comment in the selected Jira issue. Please choose another issue or contact your Jira admin.",
-				ChannelId: post.ChannelId,
-				UserId:    ji.GetPlugin().getConfig().botUserID,
-			}
-			_ = api.SendEphemeralPost(mattermostUserId, reply)
-
-			// No need to return an error to the client
-			w.Header().Set("Content-Type", "application/json")
-			_, err = io.WriteString(w, "{}")
-			if err != nil {
-				return http.StatusInternalServerError,
-					errors.WithMessage(err, "failed to write response")
-			}
-			return http.StatusOK, nil
+			return http.StatusNotFound,
+				errors.New("You do not have permission to create a comment in the selected Jira issue. Please choose another issue or contact your Jira admin.")
 		}
 
 		// The error was not a permissions error; it was unanticipated. Return it to the client.
--- a/server/issue.go
+++ b/server/issue.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"strings"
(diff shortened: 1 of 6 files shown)


webapp/src/components/jira_issue_selector/jira_issue_selector.jsx
```diff
@@ -27,11 +26,9 @@ export default class JiraIssueSelector extends Component {
     };
 
     searchIssues = (text) => {
-        const projectSearchTerm = this.props.currentProject ? 'project=' + this.props.currentProject : '';
         const textEncoded = encodeURIComponent(text.trim().replace(/"/g, '\\"'));
         const textSearchTerm = (textEncoded.length > 0) ? 'text ~ "' + textEncoded + '*"' : '';
-        const combinedTerms = (projectSearchTerm.length > 0 && textSearchTerm.length > 0) ? projectSearchTerm + ' AND ' + textSearchTerm : projectSearchTerm + textSearchTerm;
-        const finalQuery = combinedTerms + ' ' + searchDefaults;
+        const finalQuery = textSearchTerm + ' ' + searchDefaults;
```
levb: Not sure I fully understand this one, but if so... nice! (simplification)

server/issue.go
```diff
@@ -370,6 +371,25 @@ func httpAPIAttachCommentToIssue(ji Instance, w http.ResponseWriter, r *http.Req
 
 	commentAdded, _, err := jiraClient.Issue.AddComment(attach.IssueKey, &jiraComment)
 	if err != nil {
+		if strings.Contains(err.Error(), "you do not have the permission to comment on this issue") {
```
levb: This LGTM, but I wonder if there's a more "future-proof" way of doing this check by decoding the response' body, as JSON?

webapp/src/components/jira_issue_selector/jira_issue_selector.jsx
cpoile: @levb since we're no longer using project to narrow down the issue search, no need to include it in the jql query. = free simplification. :)

cpoile:
@jasonblais Shoot... I read the ticket too quickly, missed that you wanted it in the modal. Let me look...

server/issue.go
cpoile: checking...

cpoile: @levb It looks like go-jira only returns the error, it doesn't pass along the response... And the error looks to just be a string...  In the Jira API it's listed as `A schema has not been defined for this response code.`

cpoile:
@jasonblais How about this?
image

jasonblais:
@cpoile Sorry, in retrospect, I realize the placement and design of the error message should have been specified more clearly in the ticket. My mistake.

@asaadmahmood I would like your feedback on where and how this error message should be placed that follows [our UX guidelines](https://docs.mattermost.com/developer/fx-guidelines.html#feedback-error-messages) and is consistent with the placement of these messages elsewhere in the UI.

cpoile:
@jasonblais hah, yah, I had a suspicion that’s not really what we want. :) awaiting asaad’s design

asaadmahmood:
Previously when the error was triggered by the select, I intended it to be placed below the select box. But now that it is triggered by the attach button, we can place it above like this.
localhost_8065_mattermost_channels_town-square

I also thought about this, but this is not that prominent (so we will not use this).
localhost_8065_mattermost_channels_town-square (1)

cpoile:
@jasonblais @asaadmahmood I like the first one, but it's up to you guys which fits with our UI. Let me know and I'll make the change.

jasonblais:
@cpoile Let's go with the first option

cpoile:
@jasonblais @asaadmahmood  Does this look good? I'm using `help-text` and `error-text` classes, but they're different from the `input__help` and `error` classes that the webapp uses. I tried to make those work for 20 minutes, but they just aren't picked up in this modal for some reason (css mystifies me sometimes). We use `help-text` and `error-text` in this modal already, so the spacing and colors match.

image

asaadmahmood:
Looks good!

PR #144, opened 2019-06-07
[MM-16098] Add a plugin.json setting to allow disabling of all Jirav2 UI features

cpoile:
#### Summary
- added a system console setting
- opened up a new API endpoint (didn't think it belonged in the user info, and needed this settings info before the userInfo)
- it's quick and dirty -- the system console settings changes are not pushed to the clients, so if they had the ability to use the UI, they will still have it until they refresh. Open to any suggestions if you think it's too simplistic.

UX:
image

#### Links
[MM-16098](https://mattermost.atlassian.net/browse/MM-16098)

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -26,7 +26,7 @@
         "key": "EnableJiraUI",
         "display_name": "Allow users to connect their Mattermost accounts to Jira",
         "type": "bool",
-        "help_text": "When true, users can create and manage issues across Mattermost channels by connecting their Mattermost account to Jira with '/jira connect'. \n \n After changing this setting to true, install this plugin to your Jira instance with '/jira install'. See [documentation](https://about.mattermost.com/default-jira-plugin-link-application) to learn more.",
+        "help_text": "When false, disables all Jira-related user interactions in Mattermost. Does not affect Jira webhook notifications. After changing this setting to false, disable, then re-enable this plugin in **System Console > Plugins > Plugin Management** to reset the plugin state for all users. \n \n When true, install this plugin to your Jira instance with '/jira install’ to allow users to create and manage issues across Mattermost channels. See [documentation](https://about.mattermost.com/default-jira-plugin-link-application) to learn more.",
         "default": true
       },
       {
--- a/plugin.json
+++ b/plugin.json
@@ -23,6 +23,13 @@
         "help_text": "Select the username that this integration is attached to."
       },
       {
+        "key": "EnableJiraUI",
+        "display_name": "Allow users to connect their Mattermost accounts to Jira",
+        "type": "bool",
+        "help_text": "When true, users can create and manage issues across Mattermost channels by connecting their Mattermost account to Jira with '/jira connect'. \n \n After changing this setting to true, install this plugin to your Jira instance with '/jira install'. See [documentation](https://about.mattermost.com/default-jira-plugin-link-application) to learn more.",
+        "default": true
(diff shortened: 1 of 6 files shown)


cpoile:
@jasonblais you’re right, if we ask them to restart the plugin, that will force everyone off or on. Nice thinking!

webapp/src/plugin.jsx
```diff
@@ -10,10 +10,15 @@ import AttachCommentToIssueModal from 'components/modals/attach_comment_to_issue
 import PluginId from 'plugin_id';
 
 import reducers from './reducers';
-import {handleConnectChange, getConnected, handleInstanceStatusChange} from './actions';
+import {handleConnectChange, getConnected, handleInstanceStatusChange, getSettings} from './actions';
 
 export default class Plugin {
     async initialize(registry, store) {
+        const settings = await getSettings(store.getState);
```
crspeller: Initialize blocks the webapp from loading until it's complete. Not sure it's a great idea to make a blocking request here. Maybe we can move the logic of visibility to the components?

cpoile: I thought that too, but here's my thinking:
1. There is an await already just a couple lines below, so we're already blocking the initialization.
1. It should be an extremely quick request.
1. Even if it blocks for 10 seconds, nothing blocks on the webapp, you just don't get the dropdown components until the request completes. (I slept the request on the server for 10 seconds to test it.) 

With that, I thought this was by far the simplest way of solving the problem. Yes, I could put another boolean in the state and do a test before returning the dropdown components (just like we do for instanceInstalled or userConnected), but then I'd have to add all that reducer code and selector code, and that's a lot of code for something simple. 

But, I'm open to doing it that way if you're set against this approach. :)

cpoile: @crspeller ^^

plugin.json
```diff
@@ -22,6 +22,13 @@
         "type": "username",
         "help_text": "Select the username that this integration is attached to."
       },
+      {
+        "key": "EnableJiraUI",
+        "display_name": "Allow users to connect their Mattermost accounts to Jira",
+        "type": "bool",
+        "help_text": "When true, users can create and manage issues across Mattermost channels by connecting their Mattermost account to Jira with '/jira connect'. \n \n After changing this setting to true, install this plugin to your Jira instance with '/jira install'. See [documentation](https://about.mattermost.com/default-jira-plugin-link-application) to learn more.",
```
jasonblais:
```suggestion
        "help_text": "When false, disables all Jira-related user interactions in Mattermost. Does not affect Jira webhook notifications. After changing this setting to false, disable, then re-enable this plugin in **System Console > Plugins > Plugin Management** to reset the plugin state for all users. \n \n When true, install this plugin to your Jira instance with '/jira install’ to allow users to create and manage issues across Mattermost channels. See [documentation](https://about.mattermost.com/default-jira-plugin-link-application) to learn more.",
```

jasonblais: @cpoile How's this?

webapp/src/plugin.jsx
crspeller: If it only blocks the menus from appearing then it's fine. I thought it would block the whole webapp from loading. If it becomes an issue we can do the more complicated method.

cpoile: Nope, 100% doesn't block the mattermost webapp.

PR #150, opened 2019-06-10
MM-16110: Only use slack attachments when there is something to show …

levb:
Ticket: https://mattermost.atlassian.net/browse/MM-16110

@jasonblais if there are fields (Assignee or Priority) and no description, there will be an empty line in the slack attachment prior to the fields. It's not a huge amount of space though. Otherwise should work.

image


jasonblais:
> if there are fields (Assignee or Priority) and no description, there will be an empty line in the slack attachment prior to the fields. It's not a huge amount of space though. Otherwise should work.

Thanks Lev, that should be fine

PR #154, opened 2019-06-11
 [MM-16229] Implement proper client-side error handling for create/attach Jira issue dialogs

cpoile:
#### Summary
- Display a client-side error when the user clicks submit and there are empty fields marked `required`.
- Removed the built in validation for input fields (bootstrap?) because @jasonblais found that these errors did not display in the desktop app. Replaced with this method. 
- For now, the validation is simply whether or not it is required, and if so make sure it's not empty. Can be improved as needed.

Unrelated fixes:
- Found a few problems with required props, fixed those.
- Should be using find instead of filter in react-select value fields.

Looks like:
image
image

#### Links
[MM-16229](https://mattermost.atlassian.net/browse/MM-16229)

### DIFF
--- a/webapp/src/components/input.jsx
+++ b/webapp/src/components/input.jsx
@@ -66,9 +66,9 @@ export default class Input extends PureComponent {
 
     render() {
         const requiredMsg = 'This field is required.';
-        let error = null;
+        let validationError = null;
         if (this.props.required && this.state.invalidRequired) {
-            error = (
+            validationError = (
                 <p className='help-text error-text'>
                     <span>{requiredMsg}</span>
                 </p>
@@ -131,7 +131,7 @@ export default class Input extends PureComponent {
                 required={this.props.required}
             >
                 {input}
-                {error}
+                {validationError}
             </Setting>
         );
     }
--- a/webapp/src/components/input.jsx
+++ b/webapp/src/components/input.jsx
@@ -38,12 +38,12 @@ export default class Input extends PureComponent {
     constructor(props) {
         super(props);
 
-        this.state = {invalidRequired: false};
+        this.state = {invalid: false};
     }
 
     componentDidUpdate(prevProps, prevState) {
-        if (prevState.invalidRequired && this.props.value !== prevProps.value) {
-            this.setState({invalidRequired: false}); //eslint-disable-line react/no-did-update-set-state
+        if (prevState.invalid && this.props.value !== prevProps.value) {
+            this.setState({invalid: false}); //eslint-disable-line react/no-did-update-set-state
         }
     }
 
@@ -60,14 +60,14 @@ export default class Input extends PureComponent {
             return true;
         }
         const valid = this.props.value && this.props.value.toString().length !== 0;
-        this.setState({invalidRequired: !valid});
+        this.setState({invalid: !valid});
         return valid;
     };
 
     render() {
         const requiredMsg = 'This field is required.';
         let validationError = null;
-        if (this.props.required && this.state.invalidRequired) {
+        if (this.props.required && this.state.invalid) {
             validationError = (
                 <p className='help-text error-text'>
                     <span>{requiredMsg}</span>
--- a/webapp/src/components/input.jsx
+++ b/webapp/src/components/input.jsx
@@ -35,6 +35,18 @@ export default class Input extends PureComponent {
         readOnly: false,
     };
 
+    constructor(props) {
+        super(props);
+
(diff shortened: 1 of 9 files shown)


cpoile:
@jasonblais It's the error-text class, so I'm assuming yes. Here's with the default theme:
image

crspeller:
> because @jasonblais found that these errors did not display in the desktop app

 That doesn't make sense. it's Electron which is just Chrome. Any explanation for this?

This overall seems over complicating what should be simple. Can't we just validate on submit?

cpoile:
@crspeller
1. Weird, I know, but I didn't see the errors either on desktop. Try it out on community -- any input field will have the validation working (the react-select fields don't validate).
1. What do you mean by validating on submit? As in, through the form fields? The problem was the validation didn't show (point 1), and the react-select components don't have form validation.

crspeller:
@cpoile 
1. Are you saying there are existing fields that have this issue? Can you point out one for me?
2. I mean when the form is submitted run the validation and display the error like we do pretty much everywhere in MM. Maybe I am going overboard in the name of simplicity.

webapp/src/components/input.jsx
```diff
@@ -43,11 +55,31 @@ export default class Input extends PureComponent {
         }
     };
 
+    isValid = () => {
+        if (!this.props.required) {
+            return true;
+        }
+        const valid = this.props.value && this.props.value.toString().length !== 0;
+        this.setState({invalidRequired: !valid});
```
levb: nit: This line cracks me up (naming). No suggestions, though.

cpoile:
@crspeller No, they're good questions -- I just hope I'm understanding you.
1. Yes, everyone that is a field that is required. In desktop, for me and Jason, these don't show the floating tooltip when you press create (create an issue, select a project like Customer Reports, don't fill in summary, press create -- no tooltip, but it should show one).
1. Here I may not be understanding you -- this is what the PR does -- on create, it runs validation and the components that aren't valid display an error. We wanted the error to be below the field that isn't valid (as it should be).

webapp/src/components/input.jsx
cpoile: Totally agree. Not sure either, maybe just `invalid`?

levb:
@crspeller I asked basically the same even if semantically different question talking to @cpoile offline, "why don't we do it on the server?" I suppose the answer is, we _value_ displaying these errors in line, enough to have this code. I really welcome this conversation - I'm server-side (and simplicity) biased.

I'd say we have a separate conversation about it - input validation philosophy for interactive plugins? And for now merge this PR if it's the only outstanding question?

crspeller:
1. I actually thought you where meaning the other places we use react-select. Since that issue would be a bug of react-select no? Wondering if just solving this would be enough since we only validate empty.

2. What I really meant was what Lev was taking about which is centralizing the validation and not worrying about putting it inline. It's what MM does all over the interface.

webapp/src/components/input.jsx
cpoile: done

cpoile:
@crspeller 
1. Not a bug, it's on their roadmap (not implemented yet because it's complicated) https://github.com/JedWatson/react-select/issues/3140
1. @jasonblais what do you think, just have a single error message saying something like `One of the form values was marked as required but was not filled out`?

jasonblais:
This approach is more consistent with Mattermost modals as far as I know

image
image
image

@crspeller @cpoile

crspeller:
@jasonblais Thanks for pointing out those modals.
@cpoile Is it possible to do a simpler implementation like those modals do?

cpoile:
@crspeller 

What new_channel_modal does:
1. in handleSubmit: it validates the displayName element, and sets state (displayNameError) if it's not valid, then returns immediately (cancelling the submit)
2. that state change triggers the render, which displays the error

What we're doing here:
1. the same thing
1. The added difference is that we don't know how many JiraFields we have to validate. So we're putting them in a map, and going through that when we validate. If any of the components in the map are not valid, then we don't submit.
1. If we knew what fields we had ahead of time, it would be the same as what the new_channel_modal does. And for the project and issue fields (which aren't dynamic) it basically is.

We have to add the components to the map, and remove them, because people can switch projects and new jira_field components are created and destroyed. So we have to keep track of that.

I'm totally open to any other solutions. :) (Please!) But I don't think this is overly complicated. I tried a couple solutions before settling on this, but I can't think of any other way to do it when we don't know what fields we have to validate ahead of time.

PR #155, opened 2019-06-11
MM-16266: always add Reporter creating issues on Jira Server

levb:
Ticket: https://mattermost.atlassian.net/browse/MM-16266

Also updated to go-jira 1.7 that supports `Reporter`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/mattermost/mattermost-plugin-jira/server
 go 1.12
 
 require (
-	github.com/andygrunwald/go-jira v1.6.0
+	github.com/andygrunwald/go-jira v1.7.0
 	github.com/blang/semver v3.5.1+incompatible // indirect
 	github.com/dghubble/oauth1 v0.5.0
 	github.com/dgrijalva/jwt-go v3.2.0+incompatible
--- a/server/issue.go
+++ b/server/issue.go
@@ -83,6 +83,23 @@ func httpAPICreateIssue(ji Instance, w http.ResponseWriter, r *http.Request) (in
 		Fields: &create.Fields,
 	}
 
+	for i, notCovered := range create.RequiredFieldsNotCovered {
+
+		if notCovered == "reporter" || notCovered == "Reporter" {
+			requiredFieldsNotCovered := create.RequiredFieldsNotCovered[:i]
+			if i < len(create.RequiredFieldsNotCovered) {
+				requiredFieldsNotCovered = append(requiredFieldsNotCovered,
+					create.RequiredFieldsNotCovered[i+1:]...)
+			}
+			create.RequiredFieldsNotCovered = requiredFieldsNotCovered
+
+			if ji.GetType() == JIRATypeServer {
+				issue.Fields.Reporter = &jiraUser.User
+			}
+			break
+		}
+	}
+
 	project, resp, err := jiraClient.Project.Get(issue.Fields.Project.Key)
 	if err != nil {
 		message := "failed to get the project, postId: " + create.PostId
@@ -118,10 +135,6 @@ func httpAPICreateIssue(ji Instance, w http.ResponseWriter, r *http.Request) (in
 		w.Header().Set("Content-Type", "application/json")
 		fmt.Fprintf(w, "{}")
 		return http.StatusOK, nil
-	}
-
-	if issue.Fields.Reporter == nil && ji.GetType() == JIRATypeServer {
-		issue.Fields.Reporter = &jiraUser.User
 	}
 
 	created, resp, err := jiraClient.Issue.Create(issue)
--- a/server/issue.go
+++ b/server/issue.go
@@ -87,7 +87,7 @@ func httpAPICreateIssue(ji Instance, w http.ResponseWriter, r *http.Request) (in
 
 		if notCovered == "reporter" || notCovered == "Reporter" {
 			requiredFieldsNotCovered := create.RequiredFieldsNotCovered[:i]
-			if i < len(create.RequiredFieldsNotCovered) {
+			if i+1 < len(create.RequiredFieldsNotCovered) {
 				requiredFieldsNotCovered = append(requiredFieldsNotCovered,
 					create.RequiredFieldsNotCovered[i+1:]...)
 			}
--- a/server/issue.go
+++ b/server/issue.go
@@ -85,7 +85,7 @@ func httpAPICreateIssue(ji Instance, w http.ResponseWriter, r *http.Request) (in
 
 	for i, notCovered := range create.RequiredFieldsNotCovered {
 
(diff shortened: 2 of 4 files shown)


go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/mattermost/mattermost-plugin-jira/server
 go 1.12
 
 require (
-	github.com/andygrunwald/go-jira v1.6.0
+	github.com/andygrunwald/go-jira v1.7.0
```
levb: A trivial change at the last moment

server/issue.go
```diff
@@ -120,6 +120,10 @@ func httpAPICreateIssue(ji Instance, w http.ResponseWriter, r *http.Request) (in
 		return http.StatusOK, nil
 	}
 
+	if issue.Fields.Reporter == nil && ji.GetType() == JIRATypeServer {
```
crspeller: Why not both cloud and server?

go.mod
crspeller: 🤕

webapp/src/components/modals/create_issue/create_issue.jsx
```diff
@@ -61,6 +61,7 @@ export default class CreateIssueModal extends PureComponent {
         'priority',
         'description',
         'summary',
+        'reporter',
```
jfrerich: For clarification of how this change will affect the flow:

The user will now be able to fill in the reporter field in the dialog window.
* If they leave it empty, it will be handled server side and set to themselves.
* If they enter a value `jiraClient.Issue.Create(issue)` will attempt to run and return an error if reporter not found.
  * Err will be displayed at top of dialog window.

crspeller: I thought we were not going to show it?

server/issue.go
levb: The cloud (mmtest) error-ed out when Reporter was set, I lost the copy-pasted record of it, but basically was not supposed to be set. I tested on both Serveer 7 and 8, worked on both.

webapp/src/components/modals/create_issue/create_issue.jsx
levb: So I am still confused about it. I did not see it shown, anywhere. If I did not add it to the list, the webapp wouldn't even submit without it.

go.mod
levb: For `IssueFields.Reporter`

webapp/src/components/modals/create_issue/create_issue.jsx
levb: Hmmm, @jasonblais this may not be what you'd want.

jasonblais:
@levb UX-wise there should be no change, preferably.

1. We don't handle the ability to select users, therefore someone who tries to select a Reporter won't work. It'll likely fail as @jfrerich reported above.
2. Any UX changes are risky this close to release date.

When the `Reporter` field is optional, we assign the issue creator as the `Reporter`. Can we do the same when the field is required?

webapp/src/components/modals/create_issue/create_issue.jsx
jasonblais: @levb UX-wise there should be no change, preferably.

1. We don't handle the ability to select users, therefore someone who tries to select a Reporter won't work. It'll likely fail as @jfrerich reported above.
2. Any UX changes are risky this close to release date.

When the `Reporter` field is optional, we assign the issue creator as the `Reporter`. Can we do the same when the field is required?

server/issue.go
jasonblais: Is there harm in making this consistent across Cloud and Server, instead of special casing it? Or are we certain that this doesn't affect Server and thus don't want to make any changes to it?

webapp/src/components/modals/create_issue/create_issue.jsx
levb: The short answer is, no. Let me see if I can do this all in the client instead.

jasonblais: Cool. I'm also open to discussing other options

server/issue.go
levb: Setting the Reporter field seems to be erroring out on the Cloud, it appears tto want to use the session user. I kept the "server-only" logic.

server/issue.go
```diff
@@ -83,6 +83,23 @@ func httpAPICreateIssue(ji Instance, w http.ResponseWriter, r *http.Request) (in
 		Fields: &create.Fields,
 	}
 
+	for i, notCovered := range create.RequiredFieldsNotCovered {
+
+		if notCovered == "reporter" || notCovered == "Reporter" {
```
crspeller:
```suggestion
		if strings.ToLower(notCovered) == "reporter" {
```

server/issue.go
```diff
@@ -83,6 +83,23 @@ func httpAPICreateIssue(ji Instance, w http.ResponseWriter, r *http.Request) (in
...
+			requiredFieldsNotCovered := create.RequiredFieldsNotCovered[:i]
+			if i < len(create.RequiredFieldsNotCovered) {
```
crspeller: Isn't this always true?

server/issue.go
```diff
@@ -83,6 +83,23 @@ func httpAPICreateIssue(ji Instance, w http.ResponseWriter, r *http.Request) (in
...
+				requiredFieldsNotCovered = append(requiredFieldsNotCovered,
+					create.RequiredFieldsNotCovered[i+1:]...)
+			}
+			create.RequiredFieldsNotCovered = requiredFieldsNotCovered
+
+			if ji.GetType() == JIRATypeServer {
+				issue.Fields.Reporter = &jiraUser.User
```
crspeller: So we don't just want this line because:
- We want to support changing the reporter in some circumstances.
- Jira cloud barfs if we set it?

Correct?

levb: Yes. Jira Cloud barfs at it. And on the Server we end up only setting it if it is listed as "required", otherwise it's empty (and defaults to the user). I am hesitant to just force it atter seeing it fail on the Cloud.

server/issue.go
levb: Whoops. Thanks for catching!

server/issue.go
jasonblais: @levb what happens on Jira Cloud then?

levb: @jasonblais I'll test some more, but so far everything appeared to work. mmtest has some weird field setup, so I couldn't really test there, worked for me on `levbrouk`. I suggest that any cloud-specific issues that arise we'd have to address later as bugs. This is still an improvement for the server.

jasonblais: Thanks @levb, that sounds good then. We can also have Dylan smoke test post-merge.

PR #156, opened 2019-06-12
[MM-16116] Fix searching for text in single-list dropdowns

jfrerich:
**Summary** 

See commit message for a description of code changes

This PR solves an issue where a field has a drop-down selection, but the user is unable to type text to filter the drop down options.

**Ticket**
https://mattermost.atlassian.net/browse/MM-16116

### DIFF
--- a/webapp/src/components/jira_field.jsx
+++ b/webapp/src/components/jira_field.jsx
@@ -25,7 +25,7 @@ export default class JiraField extends React.PureComponent {
     // includes both .value and .name because allowedValue test cases have used .value or .name
     // and are mutually exclusive.
     // should wrap this with some if else logic
-    makeReactSelectValue = (allowedValue) => {
+    makeReactSelectIconValue = (allowedValue) => {
         const iconLabel = (
             <React.Fragment>
                 <img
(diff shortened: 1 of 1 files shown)


webapp/src/components/jira_field.jsx
```diff
@@ -89,7 +98,10 @@ export default class JiraField extends React.PureComponent {
 
         // if this.props.field has allowedValues, then props.value will be an object
         if (field.allowedValues && field.allowedValues.length && field.schema.type !== 'array') {
-            const options = field.allowedValues.map(this.makeReactSelectValue);
+            let options = field.allowedValues.map(this.makeReactSelectValue);
+            if (field.name === 'Priority') {
```
crspeller: Why are we special casing Priority? 
Why can't we follow the example here? https://github.com/JedWatson/react-select/blob/master/docs/examples/CustomOption.js

jasonblais: @crspeller FYI from the commit message (if relevant)

> Note : priority is the only dropdown field that does not currently support searching by text. This is because options.label expects a string, not a fragment.

jfrerich: I'm not seeing how to remove the `priority` special casing, even with that example.  (I want to, but don't see it).

PR #158, opened 2019-06-12
MM-16116 Fix serching in single select dropdowns.

crspeller:
Uses a custom component to render options that has icon if available. Avoids the issue of component labels not being searchable without special casing priority.

Replaces: https://github.com/mattermost/mattermost-plugin-jira/pull/156


jfrerich:
Thanks for the walkthrough, @crspeller.  Makes sense after the fact and avoids the special casing in my original PR.

PR #160, opened 2019-06-12
[MM-15229] Add "/jira view <issuekey>" command

cpoile:
#### Summary
- Adds a `/jira view` command
- As described in the ticket: Makes the default command point to view, e.g. `/jira issue`.
- Not reusing the webhook parser; wrote a new one. Correctly handles jira messages now.

UX looks like:
image

#### Links
[MM-15229](https://mattermost.atlassian.net/browse/MM-15229)

### DIFF
--- a/server/issue_parser.go
+++ b/server/issue_parser.go
@@ -1,0 +1,62 @@
+// See License for license information.
+// Copyright (c) 2017-present Mattermost, Inc. All Rights Reserved.
+
+package main
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+
+	jira "github.com/andygrunwald/go-jira"
+
+	"github.com/mattermost/mattermost-server/model"
+)
+
+var jiraLinkWithTextRegex = regexp.MustCompile(`\[([^\[]+)\|([^\]]+)\]`)
+
+func parseJiraLinksToMarkdown(text string) string {
+	return jiraLinkWithTextRegex.ReplaceAllString(text, "[${1}](${2})")
+}
+
+func mdKeySummaryLink(issue *jira.Issue) string {
+	pos := strings.LastIndex(issue.Self, "/rest/api")
+	if pos < 0 {
+		return ""
+	}
+	return fmt.Sprintf("[%s](%s%s)", issue.Key+": "+issue.Fields.Summary, issue.Self[:pos], "/browse/"+issue.Key)
+}
+
+func parseIssue(issue *jira.Issue) []*model.SlackAttachment {
+	text := mdKeySummaryLink(issue)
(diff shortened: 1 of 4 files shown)


cpoile:
@jasonblais Thanks!

1. Ok, I made that change:
image
The only problem is that the way the webhook parser works, it puts in issue type + link at the top of the attachment, which is duplicating information. If you want to remove that, I will have to either:
1. remove the first line of the attachment (now we're getting /really/ hacky); or
1. write a new parser specific to get issue

I have no problem doing 2, but it will duplicate a lot of code we already have.

2. Yes.

jasonblais:
> it puts in issue type + link at the top of the attachment, which is duplicating information

Oh..that was not obvious, I actually thought that was part of the issue description itself.. :) 

In addition to the two options proposed, are there other options we have, e.g. format how the type and link are rendered within the attachment itself?

cpoile:
@jasonblais Okay, updated. Now looks like:
image
Links are parsed correctly now (need someone to review my regex for a second opinion -- @levb ?)

server/issue_parser.go
```diff
@@ -0,0 +1,62 @@
...
+	pos := strings.LastIndex(issue.Self, "/rest/api")
+	if pos < 0 {
+		return ""
+	}
+	return fmt.Sprintf("[%s](%s%s)", issue.Key+": "+issue.Fields.Summary, issue.Self[:pos], "/browse/"+issue.Key)
+}
+
+func parseIssue(issue *jira.Issue) []*model.SlackAttachment {
+	text := mdKeySummaryLink(issue)
+	fmt.Printf("%q", issue.Fields.Description)
```
jfrerich: Can you remove this print?

jfrerich: Can you remove this print?

server/issue_parser.go
```diff
@@ -0,0 +1,62 @@
...
+		fields = append(fields, &model.SlackAttachmentField{
+			Title: "Priority",
+			Value: issue.Fields.Priority.Name,
+			Short: true,
+		})
+	}
+	return []*model.SlackAttachment{
+		{
+			// TODO is this supposed to be themed?
+			Color:  "#95b7d0",
```
jfrerich: I don't have the answer for this TODO, but I assume this hardcoded value can be pulled from somewhere.

server/issue_parser.go
cpoile: whoops, missed that, thanks!

server/issue_parser.go
cpoile: I bet, but I don't know where, and the original webhook code also has it hardcoded. I think it's probably a different PR -- to pull in the client's theme somehow.

PR #161, opened 2019-06-13
MM-16098 Clarify what exactly EnableJiraUI enables

jasonblais:
State that EnableJiraUI allows users to attach and create Jira issues in Mattermost. When disabled, users aren't able to take these actions.

Fixes https://mattermost.atlassian.net/browse/MM-16098

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -24,9 +24,9 @@
       },
       {
         "key": "EnableJiraUI",
-        "display_name": "Allow users to connect their Mattermost accounts to Jira",
+        "display_name": "Allow users to attach and create Jira issues in Mattermost",
         "type": "bool",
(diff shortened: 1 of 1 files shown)


cpoile:
@jasonblais There might be a better way to word the help text as well. Maybe start with the "When true...", and then finish with the "When false."  
Also, consider removing the text re: /jira install -- because we can still install and everything, regardless of this setting.

jasonblais:
@cpoile Given the setting is intended for the use case when the admin wants to disable the setting, I started it with `When false..`. That said, I'm open to switching the order.

As for removing the text about `/jira install` - I'd lean towards leaving it so that it's more obvious the admin has to take an action before users can create and attach issues. While it's true they can execute it even if the setting is disabled, it won't really be useful, at least not currently.

jasonblais:
cc @levb for this PR, queued for Jira v2.0.4

PR #162, opened 2019-06-13
[MM-16308] Single Select validation issue

cpoile:
#### Summary
- essentially I was trying to compare objects, which doesn't work.  (facepalm)

#### Links
Fixes: [MM-16308](https://mattermost.atlassian.net/browse/MM-16308)

### DIFF
--- a/webapp/src/components/react_select_setting.jsx
+++ b/webapp/src/components/react_select_setting.jsx
@@ -27,7 +27,7 @@ export default class ReactSelectSetting extends React.PureComponent {
     }
 
     componentDidUpdate(prevProps, prevState) {
-        if (prevState.invalid && this.props.value !== prevProps.value) {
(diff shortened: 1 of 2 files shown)


webapp/src/components/react_select_setting.jsx
```diff
@@ -27,7 +27,7 @@ export default class ReactSelectSetting extends React.PureComponent {
     }
 
     componentDidUpdate(prevProps, prevState) {
-        if (prevState.invalid && this.props.value !== prevProps.value) {
+        if (prevState.invalid && (this.props.value && this.props.value.value) !== (prevProps.value && prevProps.value.value)) {
```
crspeller: `.value.value` lovely :P

cpoile: very valuable value value

PR #164, opened 2019-06-13
[MM-16309] use a dummy root component to detect when user is logged in

cpoile:
#### Summary
- use a dummy root component that will set up the UI once the user is logged in

#### Links
- Fixes [MM-16309](https://mattermost.atlassian.net/browse/MM-16309)
- Requires https://github.com/mattermost/mattermost-webapp/pull/2966

### DIFF
--- a/webapp/src/plugin.jsx
+++ b/webapp/src/plugin.jsx
@@ -6,35 +6,45 @@ import CreateIssueModal from 'components/modals/create_issue';
 
 import AttachCommentToIssuePostMenuAction from 'components/post_menu_actions/attach_comment_to_issue';
 import AttachCommentToIssueModal from 'components/modals/attach_comment_to_issue';
+import SetupUI from 'components/setup_ui';
 
 import PluginId from 'plugin_id';
 
 import reducers from './reducers';
 import {handleConnectChange, getConnected, handleInstanceStatusChange, getSettings} from './actions';
 
+export let setupUI;
+
+const setupUILater = (registry, store) => async () => {
+    const settings = await getSettings(store.getState);
+    if (!settings.ui_enabled) {
+        return;
+    }
+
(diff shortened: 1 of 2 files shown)


webapp/src/plugin.jsx
```diff
@@ -6,35 +6,45 @@ import CreateIssueModal from 'components/modals/create_issue';
...
+import SetupUI from 'components/setup_ui';
 
 import PluginId from 'plugin_id';
 
 import reducers from './reducers';
 import {handleConnectChange, getConnected, handleInstanceStatusChange, getSettings} from './actions';
 
+export let setupUI;
+
+const setupUILater = (registry, store) => async () => {
```
crspeller: Instead of deferring everything. Could we just do `await getConnected()(store.dispatch, store.getState);` in the component? Then all the other components should update with the new info.

cpoile: @crspeller You mean call that in the dummy component? Off the top of my head, I'd rather keep all setup logic here, and let that dummy component just trigger the setup. But also, I'm not sure that would work because we don't want to show the UI components at all if the UI is turned off, and right now that only happens by not registering them in the first place.

cpoile: @crspeller You mean call that in the dummy component? Off the top of my head, I'd rather keep all setup logic here, and let that dummy component just trigger the setup. But also, I'm not sure that would work because we don't want to show the UI components at all if the UI is turned off, and right now that only happens by not registering them in the first place.

crspeller: Alright. I don't think this is worth worrying about at the 11th hour. It touches more broad architectural issues. If this works let's just merge it.

cpoile: @crspeller good point. This is just a fix for now. :)

PR #166, opened 2019-06-14
MM-16364 Use AccountID as applicable rather than Name

levb:
- Tried to reduce the use of `.Name` to the minimum outside of Jira
Server-specific code.
- removed rather than change `LoadMattermostUserId()` since it is not used
in this version
- Also patched `make deploy` to the current Token header format

Ticket: https://mattermost.atlassian.net/browse/MM-16364

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -123,7 +123,7 @@ deploy: dist
 ## or copying the files directly to a sibling mattermost-server directory.
 ifneq ($(and $(MM_SERVICESETTINGS_SITEURL),$(MM_ADMIN_USERNAME),$(MM_ADMIN_PASSWORD),$(CURL)),)
 	@echo "Installing plugin via API"
-	$(eval TOKEN := $(shell curl -i -X POST $(MM_SERVICESETTINGS_SITEURL)/api/v4/users/login -d '{"login_id": "$(MM_ADMIN_USERNAME)", "password": "$(MM_ADMIN_PASSWORD)"}' | grep -o MMAUTHTOKEN=[0-9a-z]\* | cut -f2 -d'=' 2> /dev/null))
+	$(eval TOKEN := $(shell curl -i -X POST $(MM_SERVICESETTINGS_SITEURL)/api/v4/users/login -d '{"login_id": "$(MM_ADMIN_USERNAME)", "password": "$(MM_ADMIN_PASSWORD)"}' | grep -o "Token: [0-9a-z]*" | cut -f2 -d' ' 2> /dev/null))
 	@curl -s -H "Authorization: Bearer $(TOKEN)" -X POST $(MM_SERVICESETTINGS_SITEURL)/api/v4/plugins -F "plugin=@dist/$(BUNDLE_NAME)" -F "force=true" > /dev/null && \
 		curl -s -H "Authorization: Bearer $(TOKEN)" -X POST $(MM_SERVICESETTINGS_SITEURL)/api/v4/plugins/$(PLUGIN_ID)/enable > /dev/null && \
 		echo "OK." || echo "Sorry, something went wrong."
--- a/server/instance_cloud.go
+++ b/server/instance_cloud.go
@@ -14,6 +14,7 @@ import (
 	"github.com/dgrijalva/jwt-go"
 	"github.com/pkg/errors"
 	ajwt "github.com/rbriski/atlassian-jwt"
+	"golang.org/x/oauth2"
 	oauth2_jira "golang.org/x/oauth2/jira"
 )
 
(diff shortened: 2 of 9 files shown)


server/instance_cloud.go
```diff
@@ -133,14 +134,17 @@ func (jci jiraCloudInstance) GetJIRAClient(jiraUser JIRAUser) (*jira.Client, err
 func (jci jiraCloudInstance) getJIRAClientForUser(jiraUser JIRAUser) (*jira.Client, *http.Client, error) {
 	oauth2Conf := oauth2_jira.Config{
 		BaseURL: jci.GetURL(),
-		Subject: jiraUser.Name,
+		Subject: jiraUser.AccountID,
```
levb: Apologies for the non-surgical change here, but I was going through all fields in the structs, and this just seemed much more intention-revealing.

server/kv.go
```diff
@@ -320,22 +320,22 @@ func (p *Plugin) StoreUserInfo(ji Instance, mattermostUserId string, jiraUser JI
...
+			fmt.Sprintf("failed to store user, mattermostUserId:%s, Jira user:%q", mattermostUserId, jiraUser.DisplayName))
 	}()
 
 	err := p.kvSet(keyWithInstance(ji, mattermostUserId), jiraUser)
 	if err != nil {
 		return err
 	}
 
-	err = p.kvSet(keyWithInstance(ji, jiraUser.Name), mattermostUserId)
+	err = p.kvSet(keyWithInstance(ji, jiraUser.UserKey), mattermostUserId)
```
levb: This (`UserKey`) is a temporary hack, need to express in a better way in 2.x. Probably as a method on the `Instance`

server/kv.go
```diff
@@ -387,14 +387,14 @@ func (p *Plugin) DeleteUserInfo(ji Instance, mattermostUserId string) (returnErr
 		return appErr
 	}
 
-	appErr = p.API.KVDelete(keyWithInstance(ji, jiraUser.Name))
+	appErr = p.API.KVDelete(keyWithInstance(ji, jiraUser.UserKey))
```
levb: ditto

server/user.go
```diff
@@ -20,6 +20,8 @@ const (
 
 type JIRAUser struct {
 	jira.User
+	// TODO replace with jira.User.Key if it can be relied upon
+	UserKey            string
```
levb: ditto

server/user.go
```diff
@@ -158,9 +160,8 @@ func (p *Plugin) StoreUserInfoNotify(ji Instance, mattermostUserId string, jiraU
 	p.API.PublishWebSocketEvent(
 		WS_EVENT_CONNECT,
 		map[string]interface{}{
-			"is_connected":  true,
-			"jira_username": jiraUser.Name,
```
levb: Does not appear to be used anywhere in webapp

Makefile
```diff
@@ -123,7 +123,7 @@ deploy: dist
 ## or copying the files directly to a sibling mattermost-server directory.
 ifneq ($(and $(MM_SERVICESETTINGS_SITEURL),$(MM_ADMIN_USERNAME),$(MM_ADMIN_PASSWORD),$(CURL)),)
 	@echo "Installing plugin via API"
-	$(eval TOKEN := $(shell curl -i -X POST $(MM_SERVICESETTINGS_SITEURL)/api/v4/users/login -d '{"login_id": "$(MM_ADMIN_USERNAME)", "password": "$(MM_ADMIN_PASSWORD)"}' | grep -o MMAUTHTOKEN=[0-9a-z]\* | cut -f2 -d'=' 2> /dev/null))
+	$(eval TOKEN := $(shell curl -i -X POST $(MM_SERVICESETTINGS_SITEURL)/api/v4/users/login -d '{"login_id": "$(MM_ADMIN_USERNAME)", "password": "$(MM_ADMIN_PASSWORD)"}' | grep -o "Token: [0-9a-z]*" | cut -f2 -d' ' 2> /dev/null))
```
hanzei: Should this change also be applied to the sample-plugin?

levb: Yes, looks like that's the new way

hanzei: Do you know when this change was introduced? Does the "old" command not work any longer?

I'm trying to figure out if we have to communicate this change.

levb: Frankly I was caught by surprise a little, and learned about this because my make stopped working. Definitely looks like a breaking change. @crspeller Can you answer the ^^ please?

crspeller: I think I chaned this a some point because whatever commands where being used here where not Linux compatible.

jasonblais: Are the changes documented and/or reflected in the sample and demo plugins? @crspeller

hanzei: > I think I chaned this a some point because whatever commands where being used here where not Linux compatible.

@crspeller I'm not sure I get what you are saying. I thought the "old" command was compatible with Linux. I did had problems on various systems with it. What exactly was/is the issue?

PR #170, opened 2019-06-14
Don't setupUI more than once

cpoile:
### Summary
Multiple dropdown items were being registered. This prevents that. 
If you can think of a better way, would be nice.

### DIFF
--- a/webapp/src/plugin.jsx
+++ b/webapp/src/plugin.jsx
@@ -14,8 +14,12 @@ import reducers from './reducers';
 import {handleConnectChange, getConnected, handleInstanceStatusChange, getSettings} from './actions';
 
 export let setupUI;
+let haveSetUpUI = false;
 
 const setupUILater = (registry, store) => async () => {
+    if (haveSetUpUI) {
+        return;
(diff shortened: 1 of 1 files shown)


webapp/src/plugin.jsx
```diff
@@ -37,6 +41,7 @@ const setupUILater = (registry, store) => async () => {
         registry.registerWebSocketEventHandler(`custom_${PluginId}_connect`, handleConnectChange(store));
         registry.registerWebSocketEventHandler(`custom_${PluginId}_disconnect`, handleConnectChange(store));
         registry.registerWebSocketEventHandler(`custom_${PluginId}_instance_status`, handleInstanceStatusChange(store));
+        haveSetUpUI = true;
```
levb: This seems to still be theoretically race-prone. Should the flag semantics be `setupUIStarted`, and set immediately after checking it?

cpoile: @levb Yah, that makes sense. I just didn't want it to fail and then have nothing. But your worry is more important.

PR #175, opened 2019-06-19
[MM-16124] Jira Connect slash command will generate a link when no instance is installed

cpoile:
#### Summary
- Check if there is an instance installed before creating a connect link
- PM review for the returned message

#### Links
- Fixes: [MM-16124](https://mattermost.atlassian.net/browse/MM-16124)

### DIFF
--- a/server/command.go
+++ b/server/command.go
@@ -90,6 +90,12 @@ func executeConnect(p *Plugin, c *plugin.Context, header *model.CommandArgs, arg
 	if len(args) != 0 {
 		return help()
 	}
+
+	_, err := p.currentInstanceStore.LoadCurrentJIRAInstance()
+	if err != nil {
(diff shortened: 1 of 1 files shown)


cpoile:
@jasonblais We could add the permissions really easily with @levb 's refactoring, so keep it in mind.

server/command.go
```diff
@@ -90,6 +90,12 @@ func executeConnect(p *Plugin, c *plugin.Context, header *model.CommandArgs, arg
 	if len(args) != 0 {
 		return help()
 	}
+
+	_, err := p.currentInstanceStore.LoadCurrentJIRAInstance()
```
levb: 0/5 would suggest changing https://github.com/mattermost/mattermost-plugin-jira/blob/master/server/kv.go#L280 to return an ErrNotFound, then check for it and not override other error messages (can wrap them, though)

PR #176, opened 2019-06-19
[MM-15786] /jira create without an installed Jira instance throws a 500 and hangs the create modal

cpoile:
#### Summary
- Check on the client-side whether the user is connected. If they aren't, send an ephemeral message.
- Also checks if the instance is connected (because if the instance is deleted while a user is connected, then the user tries to create, it would just hang).

PM review: 
The messages are: 
- `There is no Jira instance installed. Please contact your system administrator.`
- `Your username is not connected to Jira. Please type '/jira connect'.`

Links:
- Fixes: [MM-15786](https://mattermost.atlassian.net/browse/MM-15786)

### DIFF
--- a/webapp/src/actions/index.js
+++ b/webapp/src/actions/index.js
@@ -1,6 +1,9 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+import {PostTypes} from 'mattermost-redux/action_types';
+import {getCurrentChannelId} from 'mattermost-redux/selectors/entities/common';
+
 import ActionTypes from 'action_types';
 import {doFetch} from 'client';
 import {getPluginServerRoute} from 'selectors';
@@ -229,4 +232,26 @@ export function handleInstanceStatusChange(store) {
             data: msg.data,
         });
     };
+}
+
+export function sendEphemeralPost(store, message, channelId) {
+    const timestamp = Date.now();
+    const post = {
+        id: 'jiraPlugin' + Date.now(),
+        user_id: store.getState().entities.users.currentUserId,
+        channel_id: channelId || getCurrentChannelId(store.getState()),
+        message,
+        type: 'system_ephemeral',
+        create_at: timestamp,
+        update_at: timestamp,
+        root_id: '',
+        parent_id: '',
+        props: {},
+    };
+
+    store.dispatch({
+        type: PostTypes.RECEIVED_NEW_POST,
+        data: post,
+        channelId,
+    });
 }
--- a/webapp/src/hooks/hooks.js
+++ b/webapp/src/hooks/hooks.js
@@ -12,7 +12,7 @@ export default class Hooks {
     slashCommandWillBePostedHook = (message, contextArgs) => {
         if (message && (message.startsWith('/jira create ') || message === '/jira create')) {
             if (!isInstanceInstalled(this.store.getState())) {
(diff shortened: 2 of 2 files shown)


webapp/src/actions/index.js
```diff
@@ -230,3 +233,25 @@ export function handleInstanceStatusChange(store) {
         });
     };
 }
+
+export function sendEphemeralPost(store, message, channelId) {
```
levb: nit: do we have to write this? nothing in the mattermost webapp and/or redux that we can (should?) use? Also, does this come from the right user?

cpoile: @levb That's a good question. There is a sendEphemeralPost in webapp's actions/global_actions.jsx. I tried briefly but couldn't get it to work. The zoom plugin had to repeat this code, so I followed its lead.
@crspeller Do you know of a way to use the webapp function?

webapp/src/hooks/hooks.js
```diff
@@ -10,6 +11,14 @@ export default class Hooks {
 
     slashCommandWillBePostedHook = (message, contextArgs) => {
         if (message && (message.startsWith('/jira create ') || message === '/jira create')) {
+            if (!isInstanceInstalled(this.store.getState())) {
+                sendEphemeralPost(this.store, 'There is no Jira instance installed. Please contact your system administrator.');
+                return Promise.resolve({});
+            }
+            if (!isUserConnected(this.store.getState())) {
+                sendEphemeralPost(this.store, 'Your username is not connected to Jira. Please type `/jira connect`.');
```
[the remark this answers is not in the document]
jasonblais: For text, propose

```
Your Mattermost account is not connected to Jira. Please use `/jira connect` to connect your account, then try again.
```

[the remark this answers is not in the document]
cpoile: @jasonblais Yes, but also for the case where there is no instance installed.

This only really applies to the `/jira create` case, because it has to be handled on the client side. All the other slash commands are handled on the server, and Lev's refactoring will include easy-to-apply permissions filters for all commands. But sure, we should open tickets to make sure they're in place after the refactor.

webapp/src/actions/index.js
crspeller: The way to reuse that at the moment would be to move the function to mattermost-redux then import it.

cpoile: eww... I don't want to do that...  :)

cpoile: Maybe it's a good ticket, but probably not for this one.

cpoile: @levb re: right userId -- I'm not sure that it matters for ephemeral posts. Looks like the webapp function doesn't even set a userId: https://github.com/mattermost/mattermost-webapp/blob/9488f4fa2b92500e7c1254c58443b94a38ceda47/actions/global_actions.jsx#L181

The zoom plugin uses the `state.entities.users.currentUserId` like I do here.

webapp/src/hooks/hooks.js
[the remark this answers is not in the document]
jasonblais: > This only really applies to the /jira create case, because it has to be handled on the client side. All the other slash commands are handled on the server, and Lev's refactoring will include easy-to-apply permissions filters for all commands. But sure, we should open tickets to make sure they're in place after the refactor.

Okay, good to know. This is fine to special case for `/jira create` in that case and consider improving the server-side work later if/as needed 👍

PR #181, opened 2019-06-20
Cherry-picked PR #134

levb:


webapp/src/components/modals/create_issue/create_issue.jsx
```diff
@@ -154,12 +154,47 @@ export default class CreateIssueModal extends PureComponent {
...
+                <React.Fragment>
+                    <FormButton
+                        type='submit'
+                        btnClass='btn btn-primary'
+                        defaultMessage='Close'
+                        onClick={this.handleClose}
+                    />
+                </React.Fragment>
+            );
+        } else if (!jiraIssueMetadata || !jiraIssueMetadata.projects) {
```
levb: @jfrerich please note that the original PR also checked `post` here, but it's undefined in `master`. I am not sure if I missed something, or if it is correct that we no longer need it. Please confirm.

jfrerich: I don't remember deleting checking for `post` and still have in my latest `jira2` branch.

PR #182, opened 2019-06-20
[MM-15913] Update UX when creating issues for projects that require a…

levb:
…dditional fields (#119)

* - new prop in json object tells server to post link to create issue page
  in jira.
- getFieldsNotCovered() determines if any required Jira fields are
  outside realm of fields and returns array of fieldnames not
  supported
  - compares jirametadata fields to hardcoded list of fieldnames
    currently supported
  - currently passes arrays as props to jiraFields. will remove
- Issue with adding jiraIssue Struct as encoded query in link
  - query attributes are Capitalized and link requires lower case.
    (mostly). example Priority needs to be priority
  - write / find something to convert all struct keys to
    lowercase
  - can hardcode to get current set of supported fields working and fix
    in Jira 2.1
  - Link is posting to MM, but query params need to be automated
    - trying to use populated jiraIssue struct

* fill query string from allowed fields for now
  - refactor to its own function

* Small clean up and nit fix from review
Break up if statement for custom fields and allowed field names. Was
easier for debugging

* - convert message to ephemeral post when user non handled required fields
  are found
- add non-handled required fields as list in message

* for ephemeeral message, bost as Jira Bot, not system

* Remove extra period after colon

* Get botUserId through getConfig Method, not actual field name (conf)

* remove console.log.  Was causing make test / build failure

### DIFF
--- a/webapp/src/components/modals/channel_settings/channel_settings_internal.jsx
+++ b/webapp/src/components/modals/channel_settings/channel_settings_internal.jsx
@@ -8,7 +8,7 @@ import {Modal} from 'react-bootstrap';
 import ReactSelectSetting from 'components/react_select_setting';
 import FormButton from 'components/form_button';
 import Loading from 'components/loading';
-import {getProjectValues, getIssueValuesForMultipleProjects} from 'jira_issue_metadata';
+import {getProjectValues, getIssueValuesForMultipleProjects} from 'utils/jira_issue_metadata';
 
 const JiraEventOptions = [
     {value: 'jira:issue_created', label: 'Issue Created'},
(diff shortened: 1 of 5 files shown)


webapp/src/components/modals/channel_settings/channel_settings_internal.jsx
```diff
@@ -8,7 +8,7 @@ import {Modal} from 'react-bootstrap';
 import ReactSelectSetting from 'components/react_select_setting';
 import FormButton from 'components/form_button';
 import Loading from 'components/loading';
-import {getProjectValues, getIssueValuesForMultipleProjects} from 'jira_issue_metadata';
+import {getProjectValues, getIssueValuesForMultipleProjects} from 'utils/jira_issue_metadata';
```
levb: @cpoile I overlooked this in the previous merge, not sure how it passed `make all`

cpoile: oh yah, interesting...

levb:
Merged with https://github.com/mattermost/mattermost-plugin-jira/pull/183

PR #184, opened 2019-06-21
MM-16265 MM-16507 - Migrate to bot accounts

### LINKED ISSUE #90: Migrate Jira plugin to use bot accounts

----

There are two stages for the issue:

1 - Migrate Jira plugin to use bot accounts.

For an example, here is a PR that migrated the Demo Plugin to use bot accounts with a helper function https://github.com/mattermost/mattermost-plugin-demo/pull/29

2 - Remove `User` Mattermost System Console setting in plugins.json.

cpoile:
#### Summary
- Migrate to bot accounts
- Moved command responses to ephemeral posts. To make this as painless as possible, refactored `responsef(...)` -> `p.responsef(header, ...)`. Was suggested by ChristopherS -- we don't need an iconURL if we send ephemeral posts.
- Removed UserName from the config. This eliminated a couple tests (if the EnsureBot fails, the plugin won't be running).
- Moved to 5.12
- Fixed the token issue (I know it's in the backlog of merges, but it was annoying :)

Looks like:
image

image

#### Links
- Fixes [MM-16265](https://mattermost.atlassian.net/browse/MM-16265)
- [MM-16507](https://mattermost.atlassian.net/browse/MM-16507)
- Fixes #90


server/command.go
```diff
@@ -489,14 +492,18 @@ func getCommand() *model.Command {
...
-		Username:     PluginMattermostUsername,
-		IconURL:      PluginIconURL,
-		Type:         model.POST_DEFAULT,
+func (p *Plugin) postCommandResponse(args *model.CommandArgs, text string) {
+	post := &model.Post{
+		UserId:    p.getUserID(),
+		ChannelId: args.ChannelId,
+		Message:   text,
 	}
+	_ = p.API.SendEphemeralPost(args.UserId, post)
```
levb: I always thought that CommandResponse was a weird layer. I 👍 this.

server/plugin.go
```diff
@@ -99,10 +100,31 @@ func (p *Plugin) OnConfigurationChange() error {
...
+	p.updateConfig(func(conf *config) {
+		conf.botUserID = botUserID
+	})
+
+	bundlePath, err := p.API.GetBundlePath()
+	if err != nil {
+		return errors.Wrap(err, "couldn't get bundle path")
+	}
+
+	profileImage, err := ioutil.ReadFile(filepath.Join(bundlePath, "assets", "profile.png"))
```
levb: Should put this in the `/public`? @hanzei would want it I think, and would be right. We didn't do it before only to be 5.10-compatible.

jasonblais:
Tagging `Docs: Needed` label as we'll want to update docs prior to the release

server/plugin.go
hanzei: No, this should not be in `/public`. `/public` is used to serve files via http. This file just gets read and then send via RPC. There is not need to expose this to the public.

This is fine the way it is.

PR #185, opened 2019-06-21
[MM-16404] Fix Error Handling for Attach and Create issue Modals

jfrerich:
#### Summary 

This PR address error handling for create and attach issue modals

### Create Issue Modal
In the event the `/api/v2/get-create-issue-metadata` endpoint returns an error, the create issue modal will only show the error and a close button.
image

In the event createIssue API call to Atlassian fails, show the error, but keep the modal open and allow user to resolve the issue.
image

### Attach Issue Modal
Server-side errors are now shown above the `Jira Issue` drop down.  These are errors which would return from the '/api/v2/get-search-issues` endpoint.

image

#### Testing

Testing a an error from create issue API call is easy to test.  Using the summary field, add text >255 characters.

For the failed endpoints below, I am not sure how to test these or force these errors.  I manually added a return statement with a failure message for each.
 '/api/v2/get-search-issues`
`/api/v2/get-create-issue-metadata`

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-16404

### DIFF
--- a/webapp/src/components/jira_issue_selector/jira_issue_selector.jsx
+++ b/webapp/src/components/jira_issue_selector/jira_issue_selector.jsx
@@ -44,10 +44,11 @@ export default class JiraIssueSelector extends Component {
         const textSearchTerm = (textEncoded.length > 0) ? 'text ~ "' + textEncoded + '*"' : '';
         const finalQuery = textSearchTerm + ' ' + searchDefaults;
 
-        return doFetchWithResponse(this.props.fetchIssuesEndpoint + `?jql=${finalQuery}`).then(
-            ({data}) => {
-                return data;
-            });
+        doFetchWithResponse(this.props.fetchIssuesEndpoint + `?jql=${finalQuery}`).then(({data}) => {
+            return data;
+        }).catch((e) => {
+            this.setState({error: e});
+        });
     };
 
     debouncedSearchIssues = debounce(this.searchIssues, searchDebounceDelay);
@@ -64,7 +65,6 @@ export default class JiraIssueSelector extends Component {
 
     render = () => {
         const {error} = this.props;
-
         const requiredStar = (
             <span
                 className={'error-text'}
@@ -83,6 +83,20 @@ export default class JiraIssueSelector extends Component {
             );
         }
 
+        const serverError = this.state.error;
+        let errComponent;
+        if (this.state.error) {
+            errComponent = (
+                <p className='alert alert-danger'>
+                    <i
+                        className='fa fa-warning'
+                        title='Warning Icon'
+                    />
+                    <span> {serverError.toString()}</span>
+                </p>
+            );
+        }
+
         const requiredMsg = 'This field is required.';
         let validationError = null;
         if (this.props.required && this.state.invalid) {
@@ -95,6 +109,7 @@ export default class JiraIssueSelector extends Component {
 
         return (
             <div className={'form-group margin-bottom x3'}>
+                {errComponent}
                 <label
                     className={'control-label'}
                     htmlFor={'issue'}
--- a/webapp/src/components/modals/create_issue/create_issue.jsx
+++ b/webapp/src/components/modals/create_issue/create_issue.jsx
@@ -27,6 +27,7 @@ const initialState = {
         },
     },
     error: null,
+    getMetaDataError: null,
 };
 
 export default class CreateIssueModal extends PureComponent {
@@ -34,8 +35,6 @@ export default class CreateIssueModal extends PureComponent {
         close: PropTypes.func.isRequired,
         create: PropTypes.func.isRequired,
         post: PropTypes.object,
-        description: PropTypes.string,
-        channelId: PropTypes.string,
         currentTeam: PropTypes.object.isRequired,
         theme: PropTypes.object.isRequired,
         visible: PropTypes.bool.isRequired,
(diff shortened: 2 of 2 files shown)


jfrerich:
@jasonblais, agreed that we should adjust the display message. Looking at the message posted, the `errors` field seems to be just `field' with an error message.  Not sure why errorMessages is an empty array.

jfrerich:
Also, I'd like to see this working on community-release and see what was causing the errors in the Jira ticket.

jasonblais:
> agreed that we should adjust the display message. Looking at the message posted, the errors field seems to be just `field' with an error message. Not sure why errorMessages is an empty array.

Thanks @jfrerich. Let me know if this is a separate improvement ticket?

webapp/src/components/jira_issue_selector/jira_issue_selector.jsx
```diff
@@ -83,6 +83,24 @@ export default class JiraIssueSelector extends Component {
             );
         }
 
+        const serverError = this.state.error;
+        let errComponent;
+        if (this.state.error) {
+            errComponent = (
+
+                <React.Fragment>
```
cpoile: I'm not sure if you need this to be a fragment -- I think you can just return a p

webapp/src/components/modals/create_issue/create_issue.jsx
```diff
@@ -240,23 +241,55 @@ export default class CreateIssueModal extends PureComponent {
...
+
+        // if no getMetaDataError, show fields and allow user to input
+        // fields. An error at this point is from a server-side create
+        // issue submission and should be displayed (in addition to fields)
+        // to the user after clicking the create button.
+        if (error) {
+            issueError = (
                 <div style={style.modal}>
-                    {jiraIssueMetadata.error}
+                    <React.Fragment>
```
cpoile: Also here, not sure if we need it...?
Maybe these were here from sometime before, but may as well clean them up.

webapp/src/components/modals/create_issue/create_issue.jsx
```diff
@@ -240,23 +241,55 @@ export default class CreateIssueModal extends PureComponent {
...
-                        onClick={this.handleClose}
-                    />
-                </React.Fragment>
+        // if getmetadata fails, only display the error and a close button.
+        // user is not going to be able to create a ticket or even a partial
+        // ticket. likely a permissions error.
+        if (getMetaDataError) {
+            const metaDataError = (
+                <div style={style.modal}>
+                    <React.Fragment>
```
cpoile: Also here.

webapp/src/components/jira_issue_selector/jira_issue_selector.jsx
jfrerich: Verified this is not needed.  Removing

webapp/src/components/modals/create_issue/create_issue.jsx
```diff
@@ -200,28 +205,24 @@ export default class CreateIssueModal extends PureComponent {
...
-                            title='Warning Icon'
-                        />
-                        <span> {error}</span>
-                    </p>
-                </React.Fragment>
-            );
-        }
-        let component;
+        const footerClose = (
+            <React.Fragment>
```
crspeller: Don't need the fragment here.

jfrerich:
> t

jfrerich:
> > agreed that we should adjust the display message. Looking at the message posted, the errors field seems to be just `field' with an error message. Not sure why errorMessages is an empty array.
> 
> Thanks @jfrerich. Let me know if this is a separate improvement ticket?

I'd prefer to have a separate ticket so this can get merged and tested on community-release.  We can adjust the warnings later, but at this point, folks don't get any feedback.

webapp/src/components/modals/create_issue/create_issue.jsx
```diff
@@ -240,23 +239,47 @@ export default class CreateIssueModal extends PureComponent {
...
+                    <span> {error}</span>
+                </p>
+            );
+        }
+
+        // if getmetadata fails, only display the error and a close button.
+        // user is not going to be able to create a ticket or even a partial
+        // ticket. likely a permissions error.
+        if (getMetaDataError) {
+            const metaDataError = (
```
cpoile: Can just assign to component directly here, right?

webapp/src/components/modals/create_issue/create_issue.jsx
```diff
@@ -240,23 +239,47 @@ export default class CreateIssueModal extends PureComponent {
...
+                    <i
+                        className='fa fa-warning'
+                        title='Warning Icon'
+                    />
+                    <span> {getMetaDataError}</span>
+                </p>
+            );
+            component = metaDataError;
+            footer = footerClose;
+        } else if (jiraIssueMetadata && jiraIssueMetadata.error) {
```
cpoile: Probably nothing you touched here, but I'm confused about this -- is it possible to have a null getMetaDataError, but a jiraIssueMetadata.error?

jasonblais:
Sure thing, ticket created here https://mattermost.atlassian.net/browse/MM-16540

levb:
Do we really intend to merge this into the jira2 branch? The PR is marked as 2.1, so should probably be based off `master`?

jfrerich:
> Do we really intend to merge this into the jira2 branch? The PR is marked as 2.1, so should probably be based off `master`?

Sorry @levb.  I think you are correct and I compared against Jira2 out of habit.

webapp/src/components/modals/create_issue/create_issue.jsx
jfrerich: Yup. Fixing in new commit!

webapp/src/components/modals/create_issue/create_issue.jsx
jfrerich: @cpoile Actually, there is a case where jiraIssueMetadata.error is returned from the server. This occurs when getting Jira metadata returns without any projects. This was introduced from a ticket a while back.    

We could probably refactor this use one method for returning all errors.  

For reference, `getMetadataError` catches the following returned errors `issue.go`

```go
			errors.WithMessage(err, message)
```

A `jiraIssueMetadata.error` is returned from the following in `issue.go`
```go
	var bb []byte
	if len(cimd.Projects) != 0 {
		bb = []byte(`{"error": "You do not have permission to create issues in any projects. Please contact your Jira admin."}`)
	} else {
		bb, err = json.Marshal(cimd)
		if err != nil {
				errors.WithMessage(err, "failed to marshal response")
		}
	}
	_, err = w.Write(bb)

```

jfrerich: I don't have a preference on which method use, but perhaps we should only use one?  Also not sure if I should address in this ticket?

PR #189, opened 2019-06-26
MM-16588 Normalizing the "jira instance not installed" error responses; fixed help

cpoile:
#### Summary
- Fixing the different ways we were reporting "no jira instance installed"
- Hooked up `/jira help` which somehow got disconnected

PM review:
- Are these good? For now when you type `/jira connect` you get:
image
- All others are treated like an error would be, when there is no jira instance installed and you type `/jira disconnect` `/jira transition x y` `/jira view xxx:
image

#### Links
- Fixes [MM-16588](https://mattermost.atlassian.net/browse/MM-16588)


cpoile:
@levb Good point. I'm recording the errors to the log now. 

I like the idea of have typed errors, but I think that would require us to distinguish between an expected "can't find current instance" error (the admin hasn't set one up) and an unexpected error (one is set up, but we couldn't load it for some reason). And we're not making that distinction yet, right?

PR #195, opened 2019-06-27
[MM-16503] Special handling of Reporter is broken in localized Jira

cpoile:
#### Summary
- Send the key and the localized name to the server (for fields we don't support)
- Check against the key (which shouldn't change). Give feedback with the localized name.

#### Fixes
- [MM-16503](https://mattermost.atlassian.net/browse/MM-16503)

### DIFF
--- a/server/issue.go
+++ b/server/issue.go
@@ -26,7 +26,7 @@ func httpAPICreateIssue(ji Instance, w http.ResponseWriter, r *http.Request) (in
 	api := ji.GetPlugin().API
 
 	create := &struct {
-		RequiredFieldsNotCovered []string         `json:"required_fields_not_covered"`
+		RequiredFieldsNotCovered [][]string       `json:"required_fields_not_covered"`
 		PostId                   string           `json:"post_id"`
 		CurrentTeam              string           `json:"current_team"`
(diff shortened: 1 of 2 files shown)


cpoile:
I know Lev is away at the moment, but he touched this code last (and looks like he commented out the part I just put back in?) so I would like him to review -- in case I'm changing something that was done on purpose.

server/issue.go
```diff
@@ -26,7 +26,7 @@ func httpAPICreateIssue(ji Instance, w http.ResponseWriter, r *http.Request) (in
 	api := ji.GetPlugin().API
 
 	create := &struct {
-		RequiredFieldsNotCovered []string         `json:"required_fields_not_covered"`
+		RequiredFieldsNotCovered [][]string       `json:"required_fields_not_covered"`
```
levb: Nice - proper and simple fix! Thanks for figuring this out.

Why a slice and not a map?

cpoile: @levb Well, I would have preferred a tuple, but we don't have those. I thought since we're controlling what goes into that slice, it's okay to assume it's a tuple slice. I thought I map would be overkill. But 0/5 on it, your choice.

PR #196, opened 2019-06-27
[MM-16528] Currently selected options in single select list are not highlighted

cpoile:
#### Summary
- Tracked the problem down to CSS. 
- One word change fixes it.

Asaad: 
- Would this break how you expect the UX to work? We're not highlighting the item that's already selected (the one already in the text area of the select box) we're highlighting the item that they are hovering over or currently cursoring through.

#### Links
- Fixes [MM-16528](https://mattermost.atlassian.net/browse/MM-16528)

### DIFF
--- a/webapp/src/utils/styles.js
+++ b/webapp/src/utils/styles.js
@@ -31,7 +31,7 @@ export const getStyleForReactSelect = (theme) => {
         }),
         option: (provided, state) => ({
             ...provided,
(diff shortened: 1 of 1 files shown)


asaadmahmood:
@cpoile Can you send me a screenshot of what you're talking about? I don't know which component we're even talking about (specifically).

cpoile:
@asaadmahmood Here you go:
image

PR #210, opened 2019-07-09
[MM-16107] Post webhook events to channel even if no instance installed

cpoile:
#### Summary
- Problem: We were returning a 500 if there was no instance installed. This prevented webhooks from being sent to channels.
- Solution: ...Don't do that.

Also:
- Here's another example of where we might want to differentiate between an expected error ("there isn't an instance installed") and an unexpected error ("there /should/ be an instance installed, but we couldn't get it"). For now we're assuming it's an expected error, because we don’t want to be sending 500's for every webhook event -- we want to keep receiving webhooks, and returning a 500 for every webhook event might get us cut off.

#### Links
[MM-16107](https://mattermost.atlassian.net/browse/MM-16107)


server/webhook.go
```diff
@@ -75,7 +75,14 @@ func (wh webhook) PostToChannel(p *Plugin, channelId, fromUserId string) (*model
 	return post, http.StatusOK, nil
 }
 
-func (wh *webhook) PostNotifications(p *Plugin, ji Instance) ([]*model.Post, int, error) {
+func (wh *webhook) PostNotifications(p *Plugin) ([]*model.Post, int, error) {
+	// We will only send webhook events if we have a connected instance.
+	ji, err := p.currentInstanceStore.LoadCurrentJIRAInstance()
```
levb: Maybe move this after the check for len(notifications)? Also, we should probably return http.StatusOK there too...

server/webhook_http_test.go
```diff
@@ -65,6 +65,7 @@ func TestWebhookHTTP(t *testing.T) {
 		ExpectedSlackAttachment bool
 		ExpectedText            string
 		ExpectedFields          []*model.SlackAttachmentField
+		CurrentInstance         bool
```
levb: Is it worth just making this a sub-test and running it twice, with the 2 CurrrentInstance settings?

server/webhook.go
cpoile: Done! And we already did return a 200 there.

server/webhook_http_test.go
cpoile: I've spent a bit too much time trying to make this work. I looped through currentInstance true and false, but I kept running into problems with the `webhookWrapperFunc` and the `httptest recorder` -- I think they aren't being reset, even if I wrapped the entire `t.Run` function in the loop. 

Can you take a quick look at the current commit? I'm looping at line 185. It seems like it should work, but it doesn't.

PR #211, opened 2019-07-09
MM-16438: fixed and slightly improved `/jira transition`

levb:
ticket: https://mattermost.atlassian.net/browse/MM-16438

- Fixed parameter validation
- Improved transition disambiguation
- Also fixed a bug in a prior commit for unmarshalling ASC

### DIFF
--- a/server/command.go
+++ b/server/command.go
@@ -3,6 +3,7 @@ package main
 import (
 	"fmt"
 	"sort"
+	"strconv"
 	"strings"
 
 	"github.com/mattermost/mattermost-server/model"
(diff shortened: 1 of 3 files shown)


server/command.go
```diff
@@ -529,80 +530,78 @@ func (p *Plugin) responsef(commandArgs *model.CommandArgs, format string, args .
...
-// 		return p.responsef(header, "failed to delete Jira instance %s: %v", instanceKey, err)
-// 	}
-
-// 	// if that was our only instance, just respond with an empty list.
-// 	if len(known) == 1 {
-// 		return executeList(p, c, header)
-// 	}
-// 	return executeInstanceSelect(p, c, header, "1")
-// }
+func executeInstanceSelect(p *Plugin, c *plugin.Context, header *model.CommandArgs, args ...string) *model.CommandResponse {
```
cpoile: Are we exposing these now? (same for instance delete below)

levb:
I think this got merged before @aaronrothschild review, can you please review it post-factum?

aaronrothschild:
Seems to make sense.  Peronsally, I would expect both "inprogress" and "progress" to work to transition an issue to "in progress" status

server/command.go
levb: They are still commented out in the table

PR #214, opened 2019-07-11
MM-16925: Fixed leaks of message bodies with jiraClient

levb:
Ticket: https://mattermost.atlassian.net/browse/MM-16925

Many Jira client APIs return a response object that we were throwing away or not Closing otherwise. That was likely a mistake. I am not sure that this was the cause of the customer-reported issue, but it appears to be a (significant) leak nonetheless.

cc @amyblais - I am leaving the decisions to the PMs, but we may want to put this in 5.13

### DIFF
--- a/server/issue.go
+++ b/server/issue.go
@@ -101,11 +101,11 @@ func httpAPICreateIssue(ji Instance, w http.ResponseWriter, r *http.Request) (in
 	}
 
 	project, resp, err := jiraClient.Project.Get(issue.Fields.Project.Key)
+	defer CloseJiraResponse(resp)
 	if err != nil {
 		message := "failed to get the project, postId: " + create.PostId
 		if resp != nil {
 			bb, _ := ioutil.ReadAll(resp.Body)
-			resp.Body.Close()
 			message += ", details:" + string(bb)
 		}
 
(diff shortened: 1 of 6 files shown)


levb:
> 1 - Are we aware of potential risk of the changes here? There won't be much time to test before 5.13 ships.

This is a low-risk change, a thorough code review and a cursory run-through should be enough for testing.

> 2 - Is there a workaround to fix the memory leak in 5.12?
I am not 100% sure what the leak is. This PR may be one explanation, but they managed to leak a lot of megabytes... I definitely do not know how to work around it short of restarting the plugin process periodically.

server/issue.go
```diff
@@ -101,11 +101,11 @@ func httpAPICreateIssue(ji Instance, w http.ResponseWriter, r *http.Request) (in
 	}
 
 	project, resp, err := jiraClient.Project.Get(issue.Fields.Project.Key)
+	defer CloseJiraResponse(resp)
```
crspeller: Closing the response body is usually done after the error check. (as it's not necessary to close the response body if an error occurs.) Not sure if this is different for the jira client library though.

levb: It's not necessary because it (the _http_ Request.Body) is a nil on error. I wasn't 100% sure of the Jira Request, so this seemed safest, with a nil check.

crspeller: I wasn't saying that this would not work. It's just not the idiomatic way to do it. Generally it's request -> error check -> defer close without a nil check. 1/5

levb: The problem is, I don't know if there may be other situations when the Jira client itself raises an error but there is a response to clean up. This seems safer than checking the error first. We even have cases where we read the response in case of an error.

cpoile: Ohh, yah, I think the jiraClient returns an "error" for anything > 200 range, doesn't it? And it still returns the response.

levb:
Let's not merge this yet, let me investigate more. Same for master/2.1

levb:
This indeed solves nothing. Most Jira APIs already return a `NewJiraError(resp, err)` which populates the details and closes the Body. The 2 that don't are `Issue.Create` and `Issue.GetCreateIssueMetadata` but we are already handling them. I'll do a little cleanup in master but closing this one.

PR #215, opened 2019-07-11
MM-16925: fixed resp.Body leaks

levb:
- Cherry-picked https://github.com/mattermost/mattermost-plugin-jira/pull/214
- Fixed the new Jira API invocations in master
- Bumped release to 2.1.0!!!!

### DIFF
--- a/server/issue.go
+++ b/server/issue.go
@@ -216,7 +216,7 @@ func httpAPICreateIssue(ji Instance, w http.ResponseWriter, r *http.Request) (in
 					return
 				}
 				_, resp, e := jiraClient.Issue.PostAttachment(created.ID, bytes.NewReader(byteData), info.Name)
-				defer CloseJiraResponse(resp)
+				CloseJiraResponse(resp)
 				if e != nil {
 					// TODO report errors, as DMs from JIRA bot?
 					api.LogError("failed to attach file to issue: "+e.Error(), "file", info.Path, "issue", created.Key)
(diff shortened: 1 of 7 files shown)


server/issue.go
```diff
@@ -415,12 +416,12 @@ func httpAPIGetSearchIssues(ji Instance, w http.ResponseWriter, r *http.Request)
 		MaxResults: 50,
 		Fields:     []string{"key", "summary"},
 	})
+	defer CloseJiraResponse(resp)
```
jfrerich: Doesn't the continue only affect the for loop?  If so, I think you can refactor and share some of code in the if/else blocks.

https://github.com/mattermost/mattermost-plugin-jira/blob/d42602692e1602d64a40b75fcc529ffe77cd9c59/server/webhook.go#L120-L130

server/webhook.go
```diff
@@ -123,6 +123,7 @@ func (wh *webhook) PostNotifications(p *Plugin) ([]*model.Post, int, error) {
 			}
 
 			resp, _ := jiraClient.Do(req, nil)
```
jfrerich: Doesn't the continue only affect the for loop?  If so, I think you can refactor and share some of code in the if/else blocks.  You could pull out he 

https://github.com/mattermost/mattermost-plugin-jira/blob/d42602692e1602d64a40b75fcc529ffe77cd9c59/server/webhook.go#L120-L130

cpoile: That's my code, I'll make the change and commit. Good catch @jfrerich!

levb: Thanks for catching! I was search/replacing and it's really tricky with defer.

levb:
Replaced with #218

PR #216, opened 2019-07-12
MM-15927: ChangeLog parsing should scan and post all supported events, not just the first one

cpoile:
#### Summary
- Parsing all changelog events now
- Changed the wording of a couple events to show the change
- First step toward normalizing the event information (for moving towards a custom-post type for webhook events)
- Can't change background colour easily in markdown; working with @asaadmahmood to figure out a way that doesn't require changes to webapp
- Added missing Issuetype event

UX looks like:
Resolution&status change:  
image

That's the only webhook with multiple changelog events other than `issue_created`. But `issue_created` is handled specially, not using the changelog because Jira Server doesn't send a changelog for that event.

Slight changes to wording so that I could refactor some code:
Summary:
image
And Sprint and Rank (don't see those in the server UI, but the updated tests show the difference.

#### Links
[MM-15927](https://mattermost.atlassian.net/browse/MM-15927)


server/webhook.go
```diff
@@ -19,13 +19,20 @@ type Webhook interface {
 	PostNotifications(p *Plugin) ([]*model.Post, int, error)
 }
 
+type webhookEvent struct {
```
levb: 0/5 how about `type field`? Or if it's confusing, maybe `type webhookField`

server/webhook.go
```diff
@@ -19,13 +19,20 @@ type Webhook interface {
...
+	from  string
+	to    string
+}
+
 type webhook struct {
 	*JiraWebhook
 	eventMask     uint64
 	headline      string
 	text          string
 	fields        []*model.SlackAttachmentField
```
levb: are these even used now?

server/webhook.go
```diff
@@ -19,13 +19,20 @@ type Webhook interface {
 	PostNotifications(p *Plugin) ([]*model.Post, int, error)
 }
 
+type webhookEvent struct {
+	field string
```
levb: `name`?

server/webhook.go
cpoile: fields? Yes, for the create_issue event

levb: We should probably consolidate... why have 2 collections of fields, and there are more fields to display in Create. Just no past values. But not in this PR.

jasonblais:
I left a note in the Jira plugin channel, as this got merged with quick reviews https://community.mattermost.com/core/pl/zrxk7c7ra7fjfxrg9ina8xfzmw

> That's the only webhook with multiple changelog events

I'm thinking of another scenario where it could happen - if the user is on the Edit Issue dialog in Jira, they could end up changing multiple fields (e.g. sprint, rank, description - and add a comment). Curious whether this PR handles that case.

PR #219, opened 2019-07-14
MM-16870: Added PluginVersion to User, Instance, and Subscriptions KV

levb:
Ticket: https://mattermost.atlassian.net/browse/MM-16870

### DIFF
--- a/server/instance.go
+++ b/server/instance.go
@@ -35,8 +35,9 @@ type JIRAInstance struct {
 	*Plugin `json:"-"`
 	lock    *sync.RWMutex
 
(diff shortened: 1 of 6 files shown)


levb:
@crspeller 1/5 it does appear preferable to me to version the specific values rather than the entire store. Upgrading the individual keys "atomically" as needed appears to be less (testing) work than a systemic store migration with skip-versions. Not a strong conviction by any means.

crspeller:
Ya. That makes sense. It would be more difficult to do a migration given the nature of the KV store.

PR #221, opened 2019-07-15
MM-17000: redirect from /jira install, no need to display the link

levb:
Ticket: https://mattermost.atlassian.net/browse/MM-17000

cc @jasonblais

### DIFF
--- a/server/command.go
+++ b/server/command.go
@@ -96,13 +96,17 @@ func executeConnect(p *Plugin, c *plugin.Context, header *model.CommandArgs, arg
 		return p.help(header)
 	}
 
-	_, err := p.currentInstanceStore.LoadCurrentJIRAInstance()
+	instance, err := p.currentInstanceStore.LoadCurrentJIRAInstance()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


aaronrothschild:
Will the user get a confirmation message in the message window that they've failed/succeeded to authenticate with jira?

aaronrothschild:
Just so I'm clear....the URL to nstall is no longer displayed and when the user is redirected to the jira portal to install the plugin?  I'm trying to visualize the flow for the user.  

Is a message like "Successfully installed JIRA cloud Server 8.1.0 from your Mattermost 5.12 server" displayed afterwards to the user?

jasonblais:
@aaronrothschild This is for `/jira connect` --> so the user flow after these changes would be 
 - user posts `/jira connect` (or clicks "Connect to Jira" in the post menu when hovering over a message)
 - browser tab opens with Jira page, where user accepts the mapping between Jira and Mattermost users

There is no success/failure message afterward about a successful connection..there is a separate issue to add such a message: https://github.com/mattermost/mattermost-plugin-jira/issues/149

PR #223, opened 2019-07-15
Updated AccountId handling to match jira2

levb:
@cpoile made it work, but not quite the same way as it was done in 2.0.

@jasonblais this should also fix your issues with levbrouk.atlassian.net if I understand them correctly.

This time it is a little less hacky and should not require disconnecting/re-connecting as far as I can tell.

### DIFF
--- a/server/webhook_parser.go
+++ b/server/webhook_parser.go
@@ -270,7 +270,7 @@ func parseWebhookAssigned(jwh *JiraWebhook) *webhook {
 
 	// Don't send a notification to the assignee if they are the one who made the change. (They probably know already.)
 	if (jwh.User.Name != "" && jwh.User.Name == jwh.Issue.Fields.Assignee.Name) ||
-		(jwh.User.AccountID != "" && jwh.Issue.Fields.Assignee.AccountID != "" && jwh.User.AccountID == jwh.Issue.Fields.Assignee.AccountID) {
+		(jwh.User.AccountID != "" && jwh.User.AccountID == jwh.Issue.Fields.Assignee.AccountID) {
 		return wh
 	}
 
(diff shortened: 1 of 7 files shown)


server/webhook_parser.go
```diff
@@ -270,7 +270,7 @@ func parseWebhookAssigned(jwh *JiraWebhook) *webhook {
 
 	// Don't send a notification to the assignee if they are the one who made the change. (They probably know already.)
 	if (jwh.User.Name != "" && jwh.User.Name == jwh.Issue.Fields.Assignee.Name) ||
-		(jwh.User.AccountID != "" && jwh.Issue.Fields.Assignee.AccountID != "" && jwh.User.AccountID == jwh.Issue.Fields.Assignee.AccountID) {
```
levb: removed a redundant clause

levb:
@jasonblais Yes. This was my bad probably, I didn't merge it properly back before. The good news is that this fix prefers AccountId, and falls back to Name in the "original" Jira user value, so it should work with whatever user record you ended up with, see `/jira info` output.

server/webhook_parser.go
cpoile: That was to protect against matching a blank `jwh.Issue.Fields.Assignee.AccountID` with a blank `jwh.User.AccountID` -- they would both be blank if we have a jira Server instance, right?

cpoile: @levb you're right, you can remove the `jwh.Issue.Fields.Assignee.AccountID != ""` check:
```suggestion
		(jwh.User.AccountID != "" && jwh.User.AccountID == jwh.Issue.Fields.Assignee.AccountID) {
```

PR #225, opened 2019-07-15
MM-16002 Webhooks should process asynchronously

cpoile:
#### Summary
- Process webhooks using a buffered channel and a worker pool (fan-out).
- Added two settings to the `plugin.json` settings, need @jasonblais PM review on naming, text. 
- Need dev review on (along with this approach in general) whether to give the size of the buffer in terms of items (as now) or in terms of space (eg, choose size of buffer: `x MB`). Could estimate about 15k per item to do the conversion.

#### Link
[MM-16002](https://mattermost.atlassian.net/browse/MM-16002)


plugin.json
```diff
@@ -23,6 +23,20 @@
         "help_text": "When false, users cannot attach and create Jira issues in Mattermost. Does not affect Jira webhook notifications. After changing this setting to false, disable, then re-enable this plugin in **System Console > Plugins > Plugin Management** to reset the plugin state for all users. \n \n When true, install this plugin to your Jira instance with '/jira install’ to allow users to create and manage issues across Mattermost channels. See [documentation](https://about.mattermost.com/default-jira-plugin-link-application) to learn more.",
         "default": true
       },
+      {
+        "key": "WebhookMaxProcsPerServer",
+        "display_name": "Maximum # of Webhook Processors Per Server",
+        "type": "text",
+        "help_text": "The number of simultaneous go routines that will process webhooks. This is needed because to process a webhook the Jira plugin may have to request information from the Jira instance (such as permissions), and we may not want to overload Jira. \n \n Changes to this setting will take effect when the plugin is restarted.",
```
aaronrothschild:
```suggestion
        "help_text": "The number of simultaneous go routines that will process webhooks. This is needed because to process a webhook the Jira plugin may have to request information from the Jira instance (such as permissions), and we may not want to overload Jira. \n \n Changes to this setting will take effect when the plugin is restarted. Default value is 20",
```

plugin.json
```diff
@@ -23,6 +23,20 @@
...
+        "default": "20"
+      },
+      {
+        "key": "WebhookBufferSize",
+        "display_name": "Size of the Webhook Processing Buffer",
+        "type": "text",
+        "help_text": "The buffer size for webhook events to be processed by the webhook processors. If the buffer is full, the Jira plugin will ignore new webhook events and return a `503 Service Unavailable` response to the Jira instance that sent the webhook. This can be a large value, since each webhook event will take between 5k and 50k of memory (on average) while it's waiting to be processed. \n \n Changes to this setting will take effect when the plugin is restarted.",
```
aaronrothschild:
```suggestion
        "help_text": "The buffer size for webhook events to be processed by the webhook processors. If the buffer is full, the Jira plugin will ignore new webhook events and return a `503 Service Unavailable` response to the Jira instance that sent the webhook. This can be a large value, since each webhook event will take between 5k and 50k of memory (on average) while it's waiting to be processed. \n \n Changes to this setting will take effect when the plugin is restarted.  Default value is 10000",
```

aaronrothschild: Since we are adding these to help avoid troubleshooting problems - the user should be aware if the limits have been changed from their defaults easily.  (That is my reasoning behind adding the value to the description 1/5)

server/plugin.go
```diff
@@ -145,6 +155,22 @@ func (p *Plugin) OnActivate() error {
 		return errors.WithMessage(err, "OnActivate: failed to register command")
 	}
 
+	// Create our queue of webhook events waiting to be processed.
+	bufSize, err := strconv.Atoi(p.getConfig().WebhookBufferSize)
+	if err != nil {
+		return errors.WithMessage(err, "OnActivate: error parsing config setting WebhookBufferSize")
+	}
+	p.webhookQueue = make(chan []byte, bufSize)
```
crspeller: If the admin changes the config we will leak workers because we didn't close the previous queue. We should close the previous queue if it exists before creating a new one.

server/subscribe.go
```diff
@@ -320,29 +321,19 @@ func httpSubscribeWebhook(p *Plugin, w http.ResponseWriter, r *http.Request) (in
 		return http.StatusForbidden, fmt.Errorf("Request URL: secret did not match")
 	}
 
-	wh, jwh, err := ParseWebhook(r.Body)
+	bb, err := ioutil.ReadAll(r.Body)
```
crspeller: This is some copying. Could we just pass the reader?

crspeller:
@cpoile As for the approach in general. What issue are we trying to solve here? Every request is already it's own go routine and we don't really care about our latency of responding to the JIRA server. So I am unclear on what exactly this accomplishes. Maybe I am missing something?

server/subscribe.go
cpoile: Afaik, if we pass the reader, we could be keeping a pointer to a reader (and the underlying socket/whatever is backing it?) open until we get around to reading it. That might have weird unanticipated effects. I'd rather read the data and then return from the httpserve, then deal with the data sometime/somewhere else.

server/plugin.go
cpoile: I don't think that's a problem here since we're only doing the creation in OnActivate. You would have to kill the process (killing the old channel) then restart it to get a new channel. No?

cpoile:
@crspeller I think mostly the goal is to separate the httpserve from the webhook processing. Say we have a big burst of activity. Or if we have something io-heavy on our processing side (eg permissions-checking). We don't want to have to keep the http connections waiting for that -- there's nothing useful to send back to them. So, just take the webhook, read the data, and return.
Then process on our own time.

This lets us smooth out bursts, and limit the amount of io we do processing the webhooks (eg, we might want to limit the number of simultaneous permissions-checks we're asking to Jira).

We should tune those numbers (# of workers, buffer size) to solve those problems -- I'd be surprised if  the numbers I chose are the best.

server/plugin.go
crspeller: Oh sorry. GitHub made it look like this was in OnConfigurationChange.

cpoile:
@jasonblais @aaronrothschild After chatting with Lev, we'll leave it up to you two whether or not to expose these two parameters to the admin. 

The problem is, we don't know how big the admin's MM servers are. If they're small, 20 might be too many workers. If they're big, maybe 20 won't be enough. Maybe 20 is way too many for anyone. We just don't know at this point. Same thing for the buffer. 

We could scale the workers by then number of processors the machine has. But the limiting factor would probably be io, not cpu, so maybe that number we choose would be too small.

We could just set some reasonable numbers in the code, and wait to see if anyone has problems, and then patch the plugin if the do. 95% of our customers would probably never have a problem. 

Or we could put a big bold "Do not touch this setting unless you've been in contact with Mattermost and you know what you're doing." or "For experimental use only, do not modify."

crspeller:
@cpoile 
> We don't want to have to keep the http connections waiting for that
Why not? It's not a lot of overhead and like I mentioned above, we don't care about latency in responding to the JIRA server. 

I am not sure how this change limits our IO. Our only input is reading from the incoming request. 

I see the benefit of starting to drop incoming over a certain limit. I also see the benefit in responding immediately and forgetting about the request. At a certain point we would start running out of sockets and that would start breaking us in mysterious ways. At least here we know when it happens and what to expect.

Just to be clear I am not against this. Just looking to understand the justification.

server/subscribe.go
crspeller: Mmmm yes it would prevent the underlying system from reusing the connection. To get the full benefits of this I guess we should copy. Hopefully we don't end up shooting ourselves in the foot with memory pressure :P

cpoile:
@crspeller no worries at all, that's what the reviews are for. :)

> I see the benefit of starting to drop incoming over a certain limit. I also see the benefit in responding immediately and forgetting about the request. At a certain point we would start running out of sockets and that would start breaking us in mysterious ways. At least here we know when it happens and what to expect.

That's basically the justification, in my mind. I'm not sure this would speed anything up, overall (it might even add some overhead overall). I just like the idea of separating the httpserve from the processing, especially since we don't have to respond with anything meaningful. 

And I like the idea of smoothing out our io to Jira/whatever-else-we-have-to-do. Otherwise our io is bounded by nothing but the number of httpserves the user's machine can spin up, which is not something we can easily debug or limit if it gets out of hand.

cpoile:
@crspeller 
> I am not sure how this change limits our IO. Our only input is reading from the incoming request.

Our io during webhook processing would be limited by the number of workers in the pool. (Rather than by the number of webhook http requests we can process concurrently.)

jasonblais:
> We could just set some reasonable numbers in the code, and wait to see if anyone has problems, and then patch the plugin if the do. 95% of our customers would probably never have a problem.

> Or we could put a big bold "Do not touch this setting unless you've been in contact with Mattermost and you know what you're doing." or "For experimental use only, do not modify."

Do we have an idea of a customer environment where it's more likely that the customer will need to modify these values (e.g. if they process >1M webhooks, then increase # of workers)?

For comparison, many other plugins (e.g. GitHub) uses webhooks, but we don't process them asynchronously.

aaronrothschild:
> We could just set some reasonable numbers in the code, and wait to see if anyone has problems, and then patch the plugin if the do. 95% of our customers would probably never have a problem.

@cpoile Since my guess is likely not as good as yours, I think we should go with your numbers.   :)

jasonblais:
@cpoile on the question above 

> Do we have an idea of a customer environment where it's more likely that the customer will need to modify these values (e.g. if they process >1M webhooks, then increase # of workers)?

If we don't have, that's okay. In general, I'd lean towards not exposing the settings if it's not likely going to be a problem for a customer and may just confuse them if don't want the admin to modify these settings.

That said, if we know of a scenario where it may be more likely to happen, then we can re-evaluate.

cpoile:
@jasonblais @aaronrothschild 

Let's not expose them for now. This is what will happen if we guess too little: the buffer will start filling up as webhooks pile up and our worker pool chugs along processing them. The cost of storing a webhook on the buffer is so little that we can increase the buffer to 100k and it wouldn't make a dent in a server that's processing that many messages in the first place (100k webhooks = ~1 gb of memory). The server will get the webhook events, just at a smoothed out rate from the 20 webhook workers. It'll smooth out the bursts from Jira. 

These are the danger signs from our largest customers: they have a steady stream of webhooks, but it's just not fast enough -- the webhooks they see are getting older and older, and they have to wait minutes for the most recent ones to appear. Then they start missing webhooks completely. That's when we need to up the number of workers. We can expose these settings then, if we ever receive that kind of feedback. 

But, at that point, a customer that big will be running an HA system. Then they'll have multiple servers, each with its own pool of webhook workers. So I am 95% sure it will scale.

Sound good?

jasonblais:
@cpoile As discussed/confirmed in sprint update meeting, let's not expose the settings to admins

levb:
@crspeller "we don't really care about our latency of responding to the JIRA server" - I am not convinced that it's a safe assumption. Holding the response off until we are done posting seems wasteful and potentially harmful to JIRA (server) performance.

aaronrothschild:
I'm on board with not exposing the settings, but just curious - is this an example of a setting that should perhaps be located in the config.json rather than the plugin's settings?  I realize this is a plugin and config.json seems intended for the server itself, but I know that some plugins use info from config.json (like Auto-link and profile info) for their settings.  

Later, if/when we encounter issues, perhaps add a config.json setting like "JIRAMaxProcessBuffer" or something like that instead of a plugin setting.

0/5 - just thoughts for later.

cpoile:
@aaronrothschild That's definitely a possibility. I could make that change. I'd also want a sensible minimum value to protect sysadmins from putting in silly numbers that would deadlock the system. If you think that's a way to go, make a ticket and assign it to me -- should be a quick change.
Thanks Aaron!

crspeller:
@aaronrothschild @cpoile That's not really compatible with configuration in the DB. Since that's the way we are going, I would heasitate to make a `config.json` only setting.

cpoile:
@crspeller Oh, I read that as: a setting in the plugin's section of the config.json, that we don't expose to the system admin console -- like with autolink and welcomebot.

PR #236, opened 2019-07-18
MM-16933 Attach comment notifications for edited comments

cpoile:
#### Summary
- refactor comment_created notifications and reuse for edited comments
- fixed an issue where we weren't adding the commentAuthor to the notification. A notification looked like:
image
Now looks like:
image

#### Links
[MM-16933](https://mattermost.atlassian.net/browse/MM-16933)

### DIFF
--- a/server/webhook_parser.go
+++ b/server/webhook_parser.go
@@ -214,8 +214,17 @@ func parseWebhookCommentCreated(jwh *JiraWebhook) (Webhook, error) {
 		text:        truncate(jwh.Comment.Body, 3000),
 	}
 
+	attachCommentNotifications(wh)
+
+	return wh, nil
+}
+
+func attachCommentNotifications(wh *webhook) {
+	jwh := wh.JiraWebhook
+	commentAuthor := mdUser(&jwh.Comment.UpdateAuthor)
+
 	message := fmt.Sprintf("%s mentioned you on %s:\n>%s",
-		jwh.mdUser(), jwh.mdKeySummaryLink(), jwh.Comment.Body)
+		commentAuthor, jwh.mdKeySummaryLink(), jwh.Comment.Body)
 	for _, u := range parseJIRAUsernamesFromText(wh.Comment.Body) {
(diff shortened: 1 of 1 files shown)


server/webhook_parser.go
```diff
@@ -214,8 +214,17 @@ func parseWebhookCommentCreated(jwh *JiraWebhook) (Webhook, error) {
 		text:        truncate(jwh.Comment.Body, 3000),
 	}
 
+	attachCommentNotifications(wh)
+
+	return wh, nil
+}
+
+func attachCommentNotifications(wh *webhook) {
```
levb: nit/naming: `s/attach/append/`? or even `parseCommentNotifications`? I got confused by `attach` for a second, expecting that to mean "post" or "reply", but saw no `api` parameter.

server/webhook_parser.go
```diff
@@ -272,12 +279,15 @@ func parseWebhookCommentDeleted(jwh *JiraWebhook) (Webhook, error) {
 }
 
 func parseWebhookCommentUpdated(jwh *JiraWebhook) Webhook {
-	return &webhook{
+	wh := &webhook{
 		JiraWebhook: jwh,
 		eventTypes:  NewStringSet(eventUpdatedComment),
```
levb: totally OT, but shouldn't this just be, `events`? 0/5 would read better. Just asking for an opinion, not a change.

cpoile: You mean, instead of `wh`, the variable should be `events`? But there's only ever able to be one 'event'. So maybe event?

server/webhook_parser.go
cpoile: @levb good point, changed to append (also made the change for the other attachNotifications function

jasonblais:
@cpanato Ready to merge 🎉

server/webhook_parser.go
levb: No, sorry GH put it out of context, I was commenting on the "white" existing code. NM.

PR #247, opened 2019-07-25
[MM-17250] reset user.Name to nil value before passing to UpdateAssignee method.

jfrerich:
### Summary 

When assigning a user to a ticket, the API returns an error if both `user.AccountId` and `user.Name` is defined on the single user. 

To fix this, I nulled the `user.Name` field of the user struct before calling `jiraClient.Issue.UpdateAssignee()`

Note other error handling functions of the `/jira assign` slash commands worked as expected.  This is because the errors are handled inside the server code, before calling the Jira API function.

### Ticket
https://mattermost.atlassian.net/browse/MM-17250

### DIFF
--- a/server/issue.go
+++ b/server/issue.go
@@ -675,11 +675,10 @@ func (p *Plugin) assignJiraIssue(mmUserId, issueKey, assignee string) (string, e
 
 	// From Jira error: query parameters 'accountId' and 'username' are mutually exclusive.
 	// Here, we must choose one and one only and nil the other user field.
-	// Choosing user.AccountID over user.Name.
+	// Choosing user.AccountID over user.Name, but check if AccountId is empty.
+	// For server instances, AccountID is empty
 	if user.AccountID != "" {
 		user.Name = ""
-	} else {
-		user.AccountID = ""
 	}
 
 	if _, err := jiraClient.Issue.UpdateAssignee(issueKey, user); err != nil {
--- a/server/issue.go
+++ b/server/issue.go
@@ -673,6 +673,11 @@ func (p *Plugin) assignJiraIssue(mmUserId, issueKey, assignee string) (string, e
 	// user is array of one object
 	user := jiraUsers[0]
 
+	// From Jira error: query parameters 'accountId' and 'username' are mutually exclusive.
+	// Here, we must choose one and one only and nil the other user field.
+	// Choosing user.AccountID over user.Name.
+	user.Name = ""
+
 	if _, err := jiraClient.Issue.UpdateAssignee(issueKey, user); err != nil {
 		return "", err
 	}
--- a/server/issue.go
+++ b/server/issue.go
@@ -630,7 +630,7 @@ func (p *Plugin) assignJiraIssue(mmUserId, issueKey, assignee string) (string, e
 	}
 
 	// Get list of assignable assignees
-	url := fmt.Sprintf("rest/api/3/user/assignable/search?issueKey=%s&query=%s", issueKey, assignee)
+	url := fmt.Sprintf("rest/api/2/user/assignable/search?issueKey=%s&username=%s", issueKey, assignee)
 	req, err := jiraClient.NewRequest("GET", url, nil)
 	if err != nil {
 		return "", err
@@ -676,7 +676,11 @@ func (p *Plugin) assignJiraIssue(mmUserId, issueKey, assignee string) (string, e
 	// From Jira error: query parameters 'accountId' and 'username' are mutually exclusive.
 	// Here, we must choose one and one only and nil the other user field.
 	// Choosing user.AccountID over user.Name.
-	user.Name = ""
+	if user.AccountID != "" {
+		user.Name = ""
+	} else {
+		user.AccountID = ""
+	}
 
 	if _, err := jiraClient.Issue.UpdateAssignee(issueKey, user); err != nil {
 		return "", err


jfrerich:
@levb, @cpoile  I really think we need a test for something like this and was curious on your thoughts and how to achieve testing against a known jira install.  I'm wondering if we can setup a test server with data that can be used for testing Jira APIs?  What do you think?

server/issue.go
```diff
@@ -673,6 +673,11 @@ func (p *Plugin) assignJiraIssue(mmUserId, issueKey, assignee string) (string, e
 	// user is array of one object
 	user := jiraUsers[0]
 
+	// From Jira error: query parameters 'accountId' and 'username' are mutually exclusive.
```
levb: This should be preferring AccountId, but defaulting to Name to cover the Jira Server case?

cpoile: Maybe we should be testing what install we're on for this. `if ji.GetType() == JIRATypeCloud` etc.

levb: I'm down for any hack for 2.1 for this. The future really ought to be a `type User interface { ... Key() string ... }`

levb:
> if we can setup a test server with data that can be used for testing Jira APIs

levb:
(sorry hit the wrong button, no prompt)

server/issue.go
jfrerich: I went with the quick fix.

server/issue.go
```diff
@@ -630,7 +630,7 @@ func (p *Plugin) assignJiraIssue(mmUserId, issueKey, assignee string) (string, e
 	}
 
 	// Get list of assignable assignees
-	url := fmt.Sprintf("rest/api/3/user/assignable/search?issueKey=%s&query=%s", issueKey, assignee)
+	url := fmt.Sprintf("rest/api/2/user/assignable/search?issueKey=%s&username=%s", issueKey, assignee)
```
jfrerich: Note the change here.  Jira `Server` `api/3` does not exist, so I moved the api call to `api/2`.  However, this means the `query` parameter won't exist and we have to use the `username` parameter.

cpoile: Just curious about how this was able to work before?

DHaussermann:
@jfrerich  I see a change in behavior now where if you pass it a value that returns only 1 assignee, It will assign that assignee. 

But, I have a different concern. Should you be able to assign based first and last name they have in Jira? This does not seem to work because you can't pass a value with a space. So...
- Type `/jira assign EXT-248 Christopher`  
- Note the response that it matches 3 users and list their first and last names but not their usernames
- Attempt to assign the task using the info you were provided `/jira assign EXT-248 Christopher Speller` 
Observed: `Please specify both an issue key and assignee in the form /jira assign <issue-key> <assignee>.`

@jfrerich Please advise if we need a new task to address this concern. This may well be out of scope for this PR

jfrerich:
@DHaussermann, valid concern and catch.  Because jira is matching against `username` I would propose adding the username in the returned list of matches.  To make the results look better, I can throw the results into a table.  How does that sound?

DHaussermann:
Thanks @jfrerich  That was my thought as well. I clarified my comment after :) 

I will create a new Jira task to address this and we can make sure @aaronrothschild  is happy with this solution in triage.

jfrerich:
@DHaussermann Here are some notes with what I just found and will be helpful in the new traige ticket.

Looking at the results of the following:
`https://mmtest.atlassian.net/rest/api/2/user/assignable/search?issueKey=EXT-20&username=christopher`

```json

[
  {
    "self": "https://mmtest.atlassian.net/rest/api/2/user?accountId=5c9e55770cf73c3215ff95ad",
    "key": "admin",
    "accountId": "5c9e55770cf73c3215ff95ad",
    "accountType": "atlassian",
    "name": "admin",
    "emailAddress": "",
    "avatarUrls": {
      "48x48": "https://secure.gravatar.com/avatar/ba5e6a14bce80d8eb3b7caae62f429e9?d=https%3A%2F%2Favatar-management--avatars.us-west-2.prod.public.atl-paas.net%2Finitials%2FCS-3.png&size=48&s=48",
      "24x24": "https://secure.gravatar.com/avatar/ba5e6a14bce80d8eb3b7caae62f429e9?d=https%3A%2F%2Favatar-management--avatars.us-west-2.prod.public.atl-paas.net%2Finitials%2FCS-3.png&size=24&s=24",
      "16x16": "https://secure.gravatar.com/avatar/ba5e6a14bce80d8eb3b7caae62f429e9?d=https%3A%2F%2Favatar-management--avatars.us-west-2.prod.public.atl-paas.net%2Finitials%2FCS-3.png&size=16&s=16",
      "32x32": "https://secure.gravatar.com/avatar/ba5e6a14bce80d8eb3b7caae62f429e9?d=https%3A%2F%2Favatar-management--avatars.us-west-2.prod.public.atl-paas.net%2Finitials%2FCS-3.png&size=32&s=32"
    },
    "displayName": "Christopher Speller",
    "active": true,
    "timeZone": "America/Vancouver",
    "locale": "en_US"
  },
  {
    "self": "https://mmtest.atlassian.net/rest/api/2/user?accountId=5cc707053d24270e75a92316",
    "key": "chris.poile",
    "accountId": "5cc707053d24270e75a92316",
    "accountType": "atlassian",
    "name": "chris.poile",
    "emailAddress": "",
    "avatarUrls": {
      "48x48": "https://avatar-management--avatars.us-west-2.prod.public.atl-paas.net/5cc707053d24270e75a92316/922766e5-6b71-4339-8775-a553dd9178b3/128?size=48&s=48",
      "24x24": "https://avatar-management--avatars.us-west-2.prod.public.atl-paas.net/5cc707053d24270e75a92316/922766e5-6b71-4339-8775-a553dd9178b3/128?size=24&s=24",
      "16x16": "https://avatar-management--avatars.us-west-2.prod.public.atl-paas.net/5cc707053d24270e75a92316/922766e5-6b71-4339-8775-a553dd9178b3/128?size=16&s=16",
      "32x32": "https://avatar-management--avatars.us-west-2.prod.public.atl-paas.net/5cc707053d24270e75a92316/922766e5-6b71-4339-8775-a553dd9178b3/128?size=32&s=32"
    },
    "displayName": "Christopher Poile",
    "active": true,
    "timeZone": "America/Toronto",
    "locale": "en_US"
  }
]
```

server/issue.go
jfrerich: This was working because of testing done on mmtest and I can only assume server testing was skipped.

PR #248, opened 2019-07-25
[MM-17295] Attach as comment should also append file attachments

cpoile:
#### Summary
- We are now uploading file attachments when attaching a post to an issue.
- It's a little awkward to do it with Jira, you have to upload the attachment to the original issue, then add a reference to the attachment as text in the comment. 
- Looks like:
image

PM review:
- We're sending back notifications when the file attachments fail or when the comment fails to be updated. Line 534, and 580

#### Links
- Fixes: [MM-17295](https://mattermost.atlassian.net/browse/MM-17295)

### DIFF
--- a/server/issue.go
+++ b/server/issue.go
@@ -480,35 +480,6 @@ func httpAPIAttachCommentToIssue(ji Instance, w http.ResponseWriter, r *http.Req
 	jiraComment.Body = permalinkMessage
 	jiraComment.Body += post.Message
 
-	// Upload file attachments in the foreground to minimize the webhook noise produced by commenting
-	// then updating the comment for each attachment.
-	if len(post.FileIds) > 0 {
-		for _, fileId := range post.FileIds {
-			info, ae := api.GetFileInfo(fileId)
-			if ae != nil {
-				continue
-			}
-			// TODO: large file support? Ignoring errors for now is good enough...
-			byteData, ae := api.ReadFile(info.Path)
-			if ae != nil {
-				notifyOnFailedAttachment(ji, ae.Error(), mattermostUserId, info.Path, attach.IssueKey)
-				continue
-			}
-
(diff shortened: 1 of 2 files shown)


server/issue.go
```diff
@@ -495,6 +492,51 @@ func httpAPIAttachCommentToIssue(ji Instance, w http.ResponseWriter, r *http.Req
...
+				if attachments != nil {
+					// There will only ever be one attachment at a time.
+					attachment := (*attachments)[0]
+					jiraComment.Body += "\n\nAttachment: !" + attachment.Filename + "!"
+					updated = true
+				}
+			}
+
+			// Update the comment
+			if updated {
```
levb: nice touch!

server/issue.go
```diff
@@ -495,6 +492,51 @@ func httpAPIAttachCommentToIssue(ji Instance, w http.ResponseWriter, r *http.Req
...
+				jiraComment.ID = commentAdded.ID
+				_, _, err2 := jiraClient.Issue.UpdateComment(attach.IssueKey, &jiraComment)
+				if err2 != nil {
+					ji.GetPlugin().API.LogError("failed to update comment: "+err2.Error(), "issue", attach.IssueKey)
+
+					// Report the error to the user so they know they have to update the comment on their own.
+					_, _ = ji.GetPlugin().CreateBotDMtoMMUserId(mattermostUserId, "Failed to update comment to issue %s with file attachments. Please notify your system administrator.", attach.IssueKey)
```
aaronrothschild: What would their administrator need to do?  Why would this happen?

cpoile: @aaronrothschild The sysadmin would have to look at the error in the log to see why the upload didn't happen. I thought this was better than sending the error to the user in a DM.

levb:
@DHaussermann I am inclined to close this PR since all of the commits are included in https://github.com/mattermost/mattermost-plugin-jira/pull/256. Either way, you probably don't need to test this one separately.

PR #250, opened 2019-07-26
MM-15943 remove divider above plugin menu items

janvt:
#### Summary
Remove divider between normal menu items and plugin menu items on dot menu. Is now rendered in the main web app.

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/11054
https://mattermost.atlassian.net/browse/MM-15943

#### Related Pull Requests
https://github.com/mattermost/mattermost-webapp/pull/3264

### DIFF
--- a/webapp/src/components/post_menu_actions/create_issue/create_issue.jsx
+++ b/webapp/src/components/post_menu_actions/create_issue/create_issue.jsx
@@ -74,10 +74,6 @@ export default class CreateIssuePostMenuAction extends PureComponent {
         return (
             <React.Fragment>
                 <li
-                    className='MenuItem__divider'
-                    role='menuitem'
-                />
-                <li
                     className='MenuItem'
                     role='menuitem'
                 >


jasonblais:
Thanks @janvt for the PR! 

Integrations team: noting that support for this is dependant on https://github.com/mattermost/mattermost-webapp/pull/3264, which will be scheduled for 5.16 earliest. So we may need to bump up the minimum server requirement after this PR is merged, or as part of it.

lieut-data:
It's worth noting that installing this on versions prior to Mattermost 5.16 will "break" the existing divider. Should we add a hard dependency in the `plugin.json`, or just accept that as a transient issue and move on?

aaronrothschild:
> It's worth noting that installing this on versions prior to Mattermost 5.16 will "break" the existing divider. Should we add a hard dependency in the `plugin.json`, or just accept that as a transient issue and move on?

Break how?  Is there a screenshot?

lieut-data:
@aaronrothschild, since the JIRA plugin will now rely on the server to manage the divider, this will likely result in the following (faked) interface:
image

Just want to make sure we're on the same page. I don't think it's a major issue, but if we wanted to expend the extra effort, we might be able to conditionally add the divider from the JIRA plugin based on the server version.

aaronrothschild:
Thanks for the headsup, not a major issue and not worth the effort IMO.

PR #70, opened 2019-05-21
[MM-15425] Improve plugin deploy/redeploy/debug cycle

cpoile:
#### Summary
Improve the workflow for plugin development. 

make debug command:
- only recompile for the architecture the developer is actually using (speeds up the change/test cycle)
- compile with debug flags for the server plugin (allows us to attach a dlv process and debug the server component. I'll write instructions.)

make webapp-debug command:
- compile in dev mode (with source maps) for the webapp
- setup npm correctly so that webapp changes are compiled immediately, and only changes need to be compiled
- symlink the webapp directory so that webapp changes don't need to redeployed (a plugin restart will pick up the changes immediately)
  - this means that the webapp can be debugged directly from the development directory (yay)

make reset command:
- disable the plugin
- enable the plugin (to pick up new changes in the server and webapp)
  - this stops the current instance of the plugin (to prevent 20+ zombie plugin processes)

make stop command:
- stop the webapp watcher

Quick Instructions (I'll write more detailed instructions when I merge this into the demo plugin... so that everyone can use this)
1. Instead of running `make deploy`, run `make debug`. 
    - It will set up the link between the mattermost server and the plugin's directory.
1. Run `make webapp-debug` to build the webapp with source maps, and start the npm watch process.
1. Whenever a change is made to the plugin's server component, run `make debug` to recompile and reload the plugin.
1. Whenever a change is made to the webapp, run `make reset` to stop and restart the plugin process on the server. Your changes will now appear (you may need to refresh the page)
1. To debug the server part of the plugin, run the script `./debug-plugin.sh` which will attach a delve headless process to the plugin. Attach your debugger of choice to localhost:2346 
1. To debug the webapp part of the plugin, attach the debugger to localhost:8065 (as you would normally debug the mattermost webapp).

#### References
This is a rebase of https://github.com/mattermost/mattermost-plugin-jira/pull/43 onto master

#### Links
[MM-15425](https://mattermost.atlassian.net/browse/MM-15425)

PR #255, opened 2019-07-27
[MM-15425] [MM-13850] Improved plugin development cycle

cpoile:
### Summary
- I'd like to have our team use this during Jira development for a few weeks before making it 'official' in the `plugin-starter-template`, `plugin-demo`, and `plugin-sample` repos.
- Updated with the comments and suggestions from https://github.com/mattermost/mattermost-plugin-starter-template/pull/39
- Webapp watcher now resets the plugin after compiling any changes. Most of the time webapp changes don't require a reload (@asaadmahmood will like this). 
- Webapp watcher runs in the foreground (Jesse's request).
- Everything was moved into the makefile (eg, debug-plugin and kill-plugins)

### Workflow
This is how I've been using these for the last couple months. I will expand this into a tutorial for the docs.
- Export or source your MM env settings (MM_SERVICESETTINGS_SITEURL, MM_ADMIN_USERNAME, MM_ADMIN_PASSWORD)
- `make deploy` like before if you want to reset everything. I usually do this after switching branches, just in case.
- `make sd` (easy-to-type alias for `make server-debug`) to compile the go code (for your architecture only) and upload it to the local server.
- `make wd` (easy-to-type alias for `make webapp-debug`) to set up a webpack watcher. It will upload webapp changes after compiling.
- `make debug-plugin` to attach a headless delve process to the plugin. Point your debugger to localhost:2346
- `make kill-plugin` will kill any plugin processes. Do this when there more than one and the `make debug-plugin` doesn't know which one to attach to.
- To debug the webapp part of the plugin, attach the debugger to localhost:8065 (as you would normally debug the mattermost webapp).

### Links
- Old PR, closed: #70 
- Plugin sample PR with review comments (addressed): https://github.com/mattermost/mattermost-plugin-starter-template/pull/39
- [MM-15425](https://mattermost.atlassian.net/browse/MM-15425)
- server issue: https://github.com/mattermost/mattermost-server/issues/10171
- Solves improvement ticket: [MM-13850](https://mattermost.atlassian.net/browse/MM-13850)


webapp/webpack.config.js
```diff
@@ -71,4 +80,50 @@ module.exports = {
...
+                });
+            },
+        },
+    ],
 };
+
+// Development mode configuration
+if (DEV) {
+    config.mode = 'development';
+    config.devtool = 'source-map';
```
crspeller: Redundant. We always generate source-maps in the config above.

Makefile
```diff
@@ -179,3 +181,117 @@ endif
...
+
+# If we were debugging, we have to unattach the delve process or else we can't disable the plugin.
+# NOTE: we are assuming the dlv was listening on port 2346, as in the debug-plugin.sh script.
+	@DELVE_PID=$(shell ps aux | grep "dlv attach.*2346" | grep -v "grep" | awk -F " " '{print $$2}') && \
+	if [ "$$DELVE_PID" -gt 0 ] > /dev/null 2>&1 ; then \
+		echo "Located existing delve process running with PID: $$DELVE_PID. Killing." ; \
+		kill -9 $$DELVE_PID ; \
+	fi
+
+ifneq ($(CURL),)
```
crspeller: Maybe have the curl check above with the other precondition checks?

Makefile
```diff
@@ -179,3 +181,117 @@ endif
...
+ifeq ($(OS),Darwin)
+	cd server && env GOOS=darwin GOARCH=amd64 $(GO) build -gcflags "all=-N -l" -o dist/plugin-darwin-amd64;
+else ifeq ($(OS),Linux)
+	cd server && env GOOS=linux GOARCH=amd64 $(GO) build -gcflags "all=-N -l" -o dist/plugin-linux-amd64;
+else ifeq ($(OS),Windows_NT)
+	cd server && env GOOS=windows GOARCH=amd64 $(GO) build -gcflags "all=-N -l" -o dist/plugin-windows-amd64.exe;
+else
+	$(error make debug depends on uname to return your OS. If it does not return 'Darwin' (meaning OSX), 'Linux', or 'Windows_NT' (all recent versions of Windows), you will need to edit the Makefile for your own OS.)
+endif
+
```
hanzei: I'm a bit concerned that this is basically a copy of `bundle`. Can't you just call `bundle` instead?

Makefile
```diff
@@ -179,3 +181,117 @@ endif
...
+sd: server-debug
+
+# server-debug builds and deploys a debug version of the plugin for your architecture.
+# Then resets the plugin to pick up the changes.
+.PHONY: server-debug
+server-debug: server-debug-deploy reset
+
+.PHONY: server-debug-deploy
+server-debug-deploy:
+
```
hanzei: Nit: remove this line

Makefile
```diff
@@ -179,3 +181,117 @@ endif
...
+# wd is an easier-to-type alias for webapp-debug
+.PHONY: wd
+wd: webapp-debug
+
+# webapp-debug builds and deploys the plugin's webapp in watch mode with source-maps.
+# Webpack will run make-reset after detecting and compiling changes.
+.PHONY: webapp-debug
+webapp-debug:
+
```
hanzei: Nit: remove this line

Makefile
```diff
@@ -179,3 +181,117 @@ endif
...
+# NOTE: we are assuming the dlv was listening on port 2346, as in the debug-plugin.sh script.
+	@DELVE_PID=$(shell ps aux | grep "dlv attach.*2346" | grep -v "grep" | awk -F " " '{print $$2}') && \
+	if [ "$$DELVE_PID" -gt 0 ] > /dev/null 2>&1 ; then \
+		echo "Located existing delve process running with PID: $$DELVE_PID. Killing." ; \
+		kill -9 $$DELVE_PID ; \
+	fi
+
+ifneq ($(CURL),)
+	@echo "\nRestarting plugin via API"
+	$(eval TOKEN := $(shell curl -i -X POST $(MM_SERVICESETTINGS_SITEURL)/api/v4/users/login -d '{"login_id": "$(MM_ADMIN_USERNAME)", "password": "$(MM_ADMIN_PASSWORD)"}' | grep -o "Token: [0-9a-z]*" | cut -f2 -d' ' 2> /dev/null))
```
hanzei: What is the difference between this and `deploy`?

Makefile
cpoile: Sorry, I don't see what you mean -- bundle is pretty different from this. If anything, it's similar to deploy. But here we're only compiling for the user's OS (cutting 2/3 of the time off the build process).

Makefile
cpoile: Between `reset` and `deploy`? I'm not sure what you mean -- they're almost entirely different except that they both upload at the end (and deploy falls-back to copying). For instance, `deploy` calls `dist`, `reset` doesn't.

PR #258, opened 2019-07-30
Updated Webhook URL params to be a Mega Webhook

aaronrothschild:
As per MM-17252: https://mattermost.atlassian.net/browse/MM-17252

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -64,6 +64,6 @@
         "default": ""
       }
     ],
-    "footer": "Use this webhook URL to set up the Jira integration. See [documentation](https://about.mattermost.com/default-jira-plugin) to learn more.\n\n`https://SITEURL/plugins/jira/webhook?secret=WEBHOOKSECRET`"
+    "footer": "Use this webhook URL format to set up the Jira integration. See [documentation](https://about.mattermost.com/default-jira-plugin) to learn more.\n\n Jira 2.1 use: `https://SITEURL/plugins/jira/api/v2/webhook?secret=WEBHOOKSECRET` \n\n For Jira 2.0 configuration use this format: `https://SITEURL/plugins/jira/webhook?secret=WEBHOOKSECRET&team=TEAMURL&channel=CHANNELURL"
   }
 }
--- a/plugin.json
+++ b/plugin.json
@@ -64,6 +64,6 @@
         "default": ""
       }
     ],
-    "footer": "Use this webhook URL format to set up the Jira integration. See [documentation](https://about.mattermost.com/default-jira-plugin) to learn more.\n\n Jira 2.1 use: `https://SITEURL/plugins/jira/api/v2/webhook?secret=WEBHOOKSECRET` \n\n For Jira 2.0 configuration use this format: `https://SITEURL/plugins/jira/webhook?secret=WEBHOOKSECRET&team=TEAMURL&channel=CHANNELURL"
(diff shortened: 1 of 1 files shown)


plugin.json
```diff
@@ -64,6 +64,6 @@
         "default": ""
       }
     ],
-    "footer": "Use this webhook URL to set up the Jira integration. See [documentation](https://about.mattermost.com/default-jira-plugin) to learn more.\n\n`https://SITEURL/plugins/jira/webhook?secret=WEBHOOKSECRET&team=TEAMURL&channel=CHANNELURL`"
+    "footer": "Use this webhook URL to set up the Jira integration. See [documentation](https://about.mattermost.com/default-jira-plugin) to learn more.\n\n`https://SITEURL/plugins/jira/webhook?secret=WEBHOOKSECRET`"
```
jasonblais: I think it's using a different endpoint in the form

```
https://SITEURL/plugins/jira/api/v2/webhook?secret=WEBHOOKSECRET
```

Any value in having the old webhook URL format, or just keep it in the doc?

aaronrothschild: I've added the 2.0 and 2.1 webhook formats so it's obvious something has changed.  In the docs, I'm going to reference 2.0 setup (archived) and also 2.1's webhook setup.

jasonblais: Thanks! The changes here will be reflected for v2.1 only (since v2.0 already shipped), so I'm wondering if we need to call out the specific Jira version in this case

@aaronrothschild

levb:
@aaronrothschild @jasonblais is this ready to merge?

aaronrothschild:
I made the changes.  Ready to merge.

plugin.json
```diff
@@ -64,6 +64,6 @@
         "default": ""
       }
     ],
-    "footer": "Use this webhook URL to set up the Jira integration. See [documentation](https://about.mattermost.com/default-jira-plugin) to learn more.\n\n`https://SITEURL/plugins/jira/webhook?secret=WEBHOOKSECRET&team=TEAMURL&channel=CHANNELURL`"
+    "footer": "Use this webhook URL format to set up the Jira integration. See [documentation](https://about.mattermost.com/default-jira-plugin) to learn more.\n\n Jira 2.1 use: `https://SITEURL/plugins/jira/api/v2/webhook?secret=WEBHOOKSECRET` \n\n For per channel webhook configuration use this format: `https://SITEURL/plugins/jira/webhook?secret=WEBHOOKSECRET&team=TEAMURL&channel=CHANNELURL"
```
jasonblais: Sorry that I'm nitpicking .. but 

1. does `Jira 2.1 use:` add value since these changes will only appear in 2.1 and later?
2. is `per channel webhook configuration` obvious to a System Admin?

aaronrothschild: @jasonblais Not nitpicking, no prob.  I was going to refer to the "per channel webhook" in the plugin documentation in the area dedicated to webhooks (because it's no longer needed, but optional).  We need to explain the subtlety, but I'm second guessing including the "per channel webhooks" now because it muddles things up.  Perhaps I should just strip it down to the new mega webhook URL in the footer, and refer everything else to the docs?

jasonblais: Yeah, I'm 0/5. If there is no value having the old webhook URL format on the page, we can leave it out, given the link to docs for more details.

aaronrothschild:
@levb Second changes made, ready to merge

PR #266, opened 2019-08-02
MM-17389 Fixing connect on desktop app.

crspeller:
Fixes not being able to connect to JIRA using the buttons on the desktop app. 

Follow up issue fond in 2.1 where /jira connect opens in browser: https://mattermost.atlassian.net/browse/MM-17510


crspeller:
@levb As far as I can tell it never worked. It could just be a qerk of my configuration as it seems to work for others. I don't have an explanation.

PR #292, opened 2019-08-23
[MM-17840] Add strike-through for none on single change messages

cpoile:
#### Summary
- Add a strikethrough for the default "None"
- Make sure not to add it again when merging multiple fields

Looks like:
image
image

#### Links
- Fixes: [MM-17840](https://mattermost.atlassian.net/browse/MM-17840)


DHaussermann:
@cpoile I'm not sure if the expected has changed for this. For single field updates I was expecting to see   ~none~ but, the post no longer mentions that there was no previous values. Is this correct?
None

mickmister:
@DHaussermann I have a ticket in for improving the posts for multiselect fields: https://mattermost.atlassian.net/browse/MM-18168 . Things like labels and fix versions need to be handled to show the change/new state in a cleaner way. We currently just display what Jira tells us to by default.

cpoile:
@DHaussermann That is the original behavior, since labels were always handled special (I think we just wanted to say what changed, and just ignore what it changed from, since a lot could still be the same). 
@mickmister yep!

DHaussermann:
Okay Thanks Chris and Micheal for clarifying. 
I do see the ~none~on other cases with single field update such as Fix Version.

Issue #317, opened 2019-09-14
Support multiple languages (i8n)


jprusch:
Right now, the Jira plugin is English only. There should be support for multiple languages using language property files.

levb:
cc @crspeller @hanzei were you guys ever able to finalize the way forward for localization?

crspeller:
See: https://github.com/mattermost/mattermost-server/pull/10949

PR #341, opened 2019-10-08
[MM-18736] Add Confirm Modal for Subscription Delete Button Click

jfrerich:
## Summary

This PR adds a confirm modal that gets called when the user clicks the 'Delete` button on a subscription.  This is an added step to reduce the chance of an accidental subscription deletion.

There are two area where a user can delete a subscription.

1) From the Subscriptions List Modal.  This view shows saved subscriptions with an inline delete button for each subscription

image

Corresponding Confirm Modal for this case:

image

2) After a subscription is saved and a user clicks the `Edit` button on a subscription.

image

Corresponding Confirm Modal for this case: 
image

Other Notables for this ticket:
* the confirm modal was copied from the webapp repo, and converted to Typescript
* `zIndex` is set to 2001. This brings the modal in front of the full-screen modal.  https://github.com/mattermost/mattermost-plugin-jira/blob/3ea62625e2b269046b5fd3ef8e20dfed52e63dbc/webapp/src/components/confirm_modal.tsx#L172
* Tests were adjusted to understand the additional step of clicking the Confirm Button on the confirm modal to trigger a subscription delete.

## Ticket
https://mattermost.atlassian.net/browse/MM-18736

### DIFF
--- a/webapp/src/components/confirm_modal.tsx
+++ b/webapp/src/components/confirm_modal.tsx
@@ -92,8 +92,7 @@ export default class ConfirmModal extends PureComponent<Props, State> {
         return nextProps.show !== this.props.show;
     }
 
-    // UNSAFE_componentWillReceiveProps(nextProps) { // eslint-disable-line camelcase
-    componentDidUpdate(prevProps) { // eslint-disable-line camelcase
+    componentDidUpdate(prevProps) { 
         if (prevProps.show && !this.props.show) {
             document.removeEventListener('keydown', this.handleKeypress);
         } else if (!prevProps.show && this.props.show) {
--- a/webapp/src/components/confirm_modal.tsx
+++ b/webapp/src/components/confirm_modal.tsx
@@ -92,7 +92,7 @@ export default class ConfirmModal extends PureComponent<Props, State> {
         return nextProps.show !== this.props.show;
     }
 
-    componentDidUpdate(prevProps) { 
+    componentDidUpdate(prevProps) {
         if (prevProps.show && !this.props.show) {
             document.removeEventListener('keydown', this.handleKeypress);
         } else if (!prevProps.show && this.props.show) {
--- a/webapp/src/components/confirm_modal.tsx
+++ b/webapp/src/components/confirm_modal.tsx
@@ -1,0 +1,202 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React, {PureComponent} from 'react';
+import {Modal} from 'react-bootstrap';
+import {FormattedMessage} from 'react-intl';
+
+export type Props = {
+
+    /*
+     * Set to show modal
+     */
+    show: boolean;
+
+    /*
+     * Title to use for the modal
+     */
+    title: React.ReactNode;
+
+    /*
+     * Message to display in the body of the modal
+     */
+    message: React.ReactNode;
+
+    /*
+     * The CSS class to apply to the confirm button
+     */
+    confirmButtonClass: string;
+
+    /*
+     * The CSS class to apply to the modal
+     */
+    modalClass?: string;
+
+    /*
+     * Text/jsx element on the confirm button
+     */
+    confirmButtonText: React.ReactNode;
+
+    /*
+     * Text/jsx element on the cancel button
+     */
+    cancelButtonText: React.ReactNode;
+
+    /*
+     * Set to show checkbox
+     */
+    showCheckbox?: boolean;
+
+    /*
+     * Text/jsx element to display with the checkbox
+     */
+    checkboxText?: string;
+
+    /*
+     * Function called when the confirm button or ENTER is pressed. Passes `true` if the checkbox is checked
+     */
+    onConfirm?: (arg0: boolean) => void;
+
+    /*
+     * Function called when the cancel button is pressed or the modal is hidden. Passes `true` if the checkbox is checked
+     */
+    onCancel?: (arg0: boolean) => void;
+
+    /**
+     * Function called when modal is dismissed
+     */
+    onExited?: () => void;
+
+    /*
+     * Set to hide the cancel button
+     */
+    hideCancel: boolean;
+
+}
+
+export type State = {
+}
+
+export default class ConfirmModal extends PureComponent<Props, State> {
+    componentDidMount() {
+        if (this.props.show) {
+            document.addEventListener('keydown', this.handleKeypress);
+        }
+    }
+
+    componentWillUnmount() {
+        document.removeEventListener('keydown', this.handleKeypress);
+    }
+
+    shouldComponentUpdate(nextProps) {
+        return nextProps.show !== this.props.show;
+    }
+
+    UNSAFE_componentWillReceiveProps(nextProps) { // eslint-disable-line camelcase
+        if (this.props.show && !nextProps.show) {
+            document.removeEventListener('keydown', this.handleKeypress);
+        } else if (!this.props.show && nextProps.show) {
+            document.addEventListener('keydown', this.handleKeypress);
+        }
+    }
+
+    handleKeypress = (e) => {
+        if (e.key === 'Enter' && this.props.show) {
+            this.handleConfirm();
+        }
+    }
+
+    handleConfirm = () => {
+        // const checked = this.refs.checkbox ? this.refs.checkbox.checked : false;
+        const checked = false;
+        this.props.onConfirm(checked);
+    }
+
+    handleCancel = () => {
+        // const checked = this.refs.checkbox ? this.refs.checkbox.checked : false;
+        const checked = false;
+        this.props.onCancel(checked);
+    }
+
+    render() {
+        let checkbox;
+        if (this.props.showCheckbox) {
+            checkbox = (
+                <div className='checkbox text-right margin-bottom--none'>
+                    <label>
+                        <input
+                            ref='checkbox'
+                            type='checkbox'
+                        />
+                        {this.props.checkboxText}
+                    </label>
+                </div>
+            );
+        }
+
+        let cancelText;
+        if (this.props.cancelButtonText) {
+            cancelText = this.props.cancelButtonText;
+        } else {
+            cancelText = (
+                <FormattedMessage
+                    id='confirm_modal.cancel'
+                    defaultMessage='Cancel'
+                />
+            );
+        }
+
+        let cancelButton;
+        if (!this.props.hideCancel) {
+            cancelButton = (
+                <button
+                    type='button'
+                    className='btn btn-link btn-cancel'
+                    onClick={this.handleCancel}
+                    id='cancelModalButton'
+                >
+                    {cancelText}
+                </button>
+            );
+        }
+
+        return (
+            <Modal
+                className={'modal-confirm ' + this.props.modalClass}
+                dialogClassName='a11y__modal'
+                show={this.props.show}
+                onHide={this.props.onCancel}
+                onExited={this.props.onExited}
+                id='confirmModal'
+                role='dialog'
+                style={{zIndex: 2001}}
+                aria-labelledby='confirmModalLabel'
+            >
+                <Modal.Header closeButton={false}>
+                    <Modal.Title
+                        componentClass='h1'
+                        id='confirmModalLabel'
+                    >
+                        {this.props.title}
+                    </Modal.Title>
+                </Modal.Header>
+                <Modal.Body>
+                    {this.props.message}
+                    {checkbox}
+                </Modal.Body>
+                <Modal.Footer>
+                    {cancelButton}
+                    <button
+                        autoFocus={true}
+                        type='button'
+                        className={this.props.confirmButtonClass}
+                        onClick={this.handleConfirm}
+                        id='confirmModalButton'
+                    >
+                        {this.props.confirmButtonText}
+                    </button>
+                </Modal.Footer>
+            </Modal>
+        );
+    }
+}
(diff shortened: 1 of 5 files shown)


asaadmahmood:
The button should say "Delete" instead of "Confirm", that's more consistent with what we do for our post delete modal.
Screenshot 2019-10-09 at 6 02 29 AM

jfrerich:
Thanks for the feedback, @asaadmahmood.  I just submitted the code change and updated the images in the original PR text to reflect these latest changes.

webapp/src/components/modals/channel_settings/select_channel_subscription.tsx
```diff
@@ -39,6 +56,40 @@ export default class SelectChannelSubscriptionInternal extends React.PureCompone
             );
         }
 
+        const subscriptionRows = this.props.channelSubscriptions.map((sub) => (
+            <div
+                key={sub.id}
+                className='select-channel-subscriptions-row'
+            >
+                <ConfirmModal
```
mickmister: I wonder if we can have the confirm modal rendered and managed once in `channel_settings_internal.tsx`, to remove deplicated code. The function passed down to `EditChannelSettings` and `SelectChannelSubscription` would take the subscription to be deleted as a parameter, and stored in the state of the component temporarily while the modal is open. This would relieve the "ConfirmModal per subscription" loop here.

I think this would clean up the code nicely. I'd be happy to talk about this if it's not clear.

webapp/src/components/modals/channel_settings/select_channel_subscription.tsx
```diff
@@ -11,11 +13,26 @@ type Props = SharedProps & {
 
 type State = {
     error: string | null;
+    showConfirmModal: boolean | false;
```
mickmister: Same question about `| false` here

webapp/src/components/confirm_modal.tsx
```diff
@@ -0,0 +1,202 @@
...
+        return nextProps.show !== this.props.show;
+    }
+
+    UNSAFE_componentWillReceiveProps(nextProps) { // eslint-disable-line camelcase
+        if (this.props.show && !nextProps.show) {
+            document.removeEventListener('keydown', this.handleKeypress);
+        } else if (!this.props.show && nextProps.show) {
+            document.addEventListener('keydown', this.handleKeypress);
+        }
+    }
```
mickmister: We should try not to introduce these `UNSAFE_*` methods if possible. 2/5 use `componentDidUpdate` instead. It takes `prevProps`, while the current (or "next") props are in `this.props`.
```
    componentDidUpdate(prevProps) {
        if (prevProps.show && !this.props.show) {
            document.removeEventListener('keydown', this.handleKeypress);
        } else if (!prevProps.show && this.props.show) {
            document.addEventListener('keydown', this.handleKeypress);
        }
    }
```

webapp/src/components/confirm_modal.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+    handleKeypress = (e) => {
+        if (e.key === 'Enter' && this.props.show) {
+            this.handleConfirm();
+        }
+    }
+
+    handleConfirm = () => {
+        // const checked = this.refs.checkbox ? this.refs.checkbox.checked : false;
+        const checked = false;
```
mickmister: Should the commented out line be used here instead?

webapp/src/components/confirm_modal.tsx
```diff
@@ -0,0 +1,202 @@
...
+                </div>
+            );
+        }
+
+        let cancelText;
+        if (this.props.cancelButtonText) {
+            cancelText = this.props.cancelButtonText;
+        } else {
+            cancelText = (
+                <FormattedMessage
```
mickmister: Nice job taking a step to i18n the webapp. I think this could be a good help wanted campaign once we have the infrastructure set up for one/several repo(s).

I wonder if we get the webapp's `i18n` configs for free without any additions due to the way i18n is hooked into the component tree. @levb I think that is worth exploring for reuse of common translations.

webapp/src/components/modals/channel_settings/edit_channel_settings.test.tsx
```diff
@@ -367,6 +366,9 @@ describe('components/EditChannelSettings', () => {
         expect(wrapper.exists('#jira-delete-subscription')).toBe(true);
         wrapper.find('#jira-delete-subscription').simulate('click');
 
+        expect(wrapper.state().showConfirmModal).toBe(true);
+        wrapper.instance().handleConfirmDelete();
```
mickmister: This seems a little redundant since the previous test already tests for error (I introduced this though). 0/5 I suggest removing the check for failure in the previous test, or you can remove this test and keep the error check in the other.

webapp/src/components/modals/channel_settings/edit_channel_settings.tsx
```diff
@@ -49,6 +50,7 @@ export type State = {
     error: string | null;
     getMetaDataErr: string | null;
     submitting: boolean;
+    showConfirmModal: boolean | false;
```
mickmister: Why is this `| false` needed?

webapp/src/components/modals/channel_settings/edit_channel_settings.tsx
```diff
@@ -277,6 +293,21 @@ export default class EditChannelSettings extends PureComponent<Props, State> {
...
         }
 
+        const {showConfirmModal} = this.state;
+        const confirmComponent = (
+            <ConfirmModal
+                cancelButtonText={'Cancel'}
+                confirmButtonText={'Delete'}
+                confirmButtonClass={'btn btn-danger'}
+                hideCancel={false}
+                message={'Delete Subscription ?'}
```
mickmister: 1/5 change to this
```
let confirmComponent;
if (this.props.selectedSubscription) {
    confirmComponent = (
        <Confirm Modal
            ...
            message={`Delete Subscription ${this.props.selectedSubscription.id`}} // will be sub.name later
        />
    );
}

File: webapp/src/components/modals/channel_settings/edit_channel_settings.tsx [hunk 9]
```diff
@@ -277,6 +293,21 @@ export default class EditChannelSettings extends PureComponent<Props, State> {
...
+                onCancel={this.handleDeactivateCancel}
+                onConfirm={this.handleConfirmDelete}
+                show={showConfirmModal}
+                title={'Subscription'}
```
mickmister: I'm curious if "Subscription" is the best UX here. I'm 0/5 on changing it to "Delete Subscription". @asaadmahmood Is there an established convention for this?

webapp/src/components/modals/channel_settings/select_channel_subscription.tsx
mickmister: > The function passed down to EditChannelSettings and SelectChannelSubscription would take the subscription to be deleted as a parameter, and stored in the state of the component temporarily while the modal is open.

This component is already the parent that stores this in its state. So it wouldn't necessarily need to take it as a parameter. This change would involve changing tests too.

webapp/src/components/confirm_modal.tsx
levb: Let's have a separate conversation about this, we have https://github.com/mattermost/mattermost-plugin-jira/issues/317 for it, @crspeller and @hanzei need to work out their approach to the server-side localization.

webapp/src/components/modals/channel_settings/edit_channel_settings.tsx
levb: ditto strings above?

mickmister: The `string | null` makes sense, because `this.state.error` has its initial value set to `null` instead of say an empty string. Also, when an error is cleared, the state var gets assigned `null`.

This is convention in the webapp so I think it makes sense to do it the same here.

webapp/src/components/modals/channel_settings/edit_channel_settings.test.tsx
jfrerich: I'm going to remove the check for failure in the previous test. The name of this test suggests the check is in here.

webapp/src/components/modals/channel_settings/edit_channel_settings.tsx
jfrerich: I like this and wasn't aware of the `s for obtaining a variable value in a print string

jfrerich:
I just made some updates in the last commit.  

Still need to resolve:
1. @asaadmahmood feedback
2. refactoring, although I'm tempted to create another ticket for that and get this one merged soon

mickmister:
@jfrerich I can follow this up with a PR for the modal state refactor

webapp/src/components/confirm_modal.tsx
```diff
@@ -92,10 +92,11 @@ export default class ConfirmModal extends PureComponent<Props, State> {
         return nextProps.show !== this.props.show;
     }
 
-    UNSAFE_componentWillReceiveProps(nextProps) { // eslint-disable-line camelcase
-        if (this.props.show && !nextProps.show) {
+    // UNSAFE_componentWillReceiveProps(nextProps) { // eslint-disable-line camelcase
+    componentDidUpdate(prevProps) { // eslint-disable-line camelcase
```
mickmister: Please remove the commented out line here. Also we don't need the eslint disable line here anymore.

webapp/src/components/modals/channel_settings/edit_channel_settings.tsx
mickmister: @jfrerich I'm just curious, was it an editor warning that had you put the `boolean | false`?

webapp/src/components/confirm_modal.tsx
jfrerich: Done

webapp/src/components/modals/channel_settings/edit_channel_settings.tsx
```diff
@@ -294,19 +294,22 @@ export default class EditChannelSettings extends PureComponent<Props, State> {
...
-        );
+        let confirmComponent;
+        if (this.props.selectedSubscription) {
+            confirmComponent = (
+                <ConfirmModal
+                    cancelButtonText={'Cancel'}
+                    confirmButtonText={'Delete'}
+                    confirmButtonClass={'btn btn-danger'}
+                    hideCancel={false}
+                    message={`Delete Subscription ${this.props.selectedSubscription.id}}?`}
```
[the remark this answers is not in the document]
jfrerich: Done

mickmister:
Will approve after conflict is resolved

webapp/src/components/modals/channel_settings/edit_channel_settings.tsx
```diff
@@ -343,15 +343,14 @@ export default class EditChannelSettings extends PureComponent<Props, State> {
...
-                            onClick={this.handleDeleteChannelSubscription}
-                        />
-                    )}
+                    <FormButton
+                        id='jira-delete-subscription'
+                        type='button'
+                        btnClass='btn-danger pull-left'
+                        defaultMessage='Delete'
+                        disabled={!enableDeleteButton}
+                        onClick={this.deleteChannelSubscription}
```
mickmister: Looks like the `onClick` was reverted back to `deleteChannelSubscription` in the merge here

jfrerich: It was.  It was also caught by `make test` which I didn't run before pushing the commit.  Thanks buddy!

mickmister:
Thanks for following up with all the changes, LGTM!

jfrerich:
@mickmister and thanks for all the feedback.  Now awaiting @asaadmahmood's feedback here -> https://github.com/mattermost/mattermost-plugin-jira/pull/341#discussion_r332804795 and @DHaussermann's feedback and we can merge.

Thanks Everyone.

webapp/src/components/confirm_modal.tsx
```diff
@@ -0,0 +1,200 @@
...
+     * Set to hide the cancel button
+     */
+    hideCancel: boolean;
+
+}
+
+export type State = {
+}
+
+export default class ConfirmModal extends PureComponent<Props, State> {
```
mickmister: Nit: If a component has no state, then just don't pass it in here, and remove `export type State = {}` as well (Sorry for yet another requested change)
```suggestion
export default class ConfirmModal extends PureComponent<Props> {
```

webapp/src/components/modals/channel_settings/edit_channel_settings.tsx
```diff
@@ -94,18 +97,31 @@ export default class EditChannelSettings extends PureComponent<Props, State> {
...
                     this.setState({error: res.error.message});
                 } else {
-                    this.handleClose(e);
+                    this.handleClose();
                 }
             });
         }
     };
 
+    handleDeactivateCancel = () => {
```
mickmister:
```suggestion
    handleCancelDelete = () => {
```

mickmister: Not sure if `handleCancelDelete` or `handleDeleteCancel` sounds better. Just need to be consistent with the `handleConfirmDelete` or `handleDeleteConfirm` name.

webapp/src/components/confirm_modal.tsx
jfrerich: No apologies.  I really appreciate the nits too :)

webapp/src/components/modals/channel_settings/edit_channel_settings.tsx
jfrerich: I think `handleCancelDelete` is fitting.  Thanks for all the reviews this morning! I'm ready to get the feature branch merged to master soon!

PR #356, opened 2019-10-23
[MM-19639] If channel is a DM or GM, don't add the channel name to the title of the subscription modal

jfrerich:
## Summary 
When opening the subscription modal from a DM or GM Channel, show a generic Title `Jira Subscriptions`, because the channel name is not human readable.

Use `channel.display`, not `channel.name`, when displaying the channel name for non-DM and non-GM channels.

## Ticket
https://mattermost.atlassian.net/browse/MM-19639


webapp/src/components/modals/channel_settings/index.ts
```diff
@@ -24,9 +25,11 @@ import ChannelSettingsModal from './channel_settings';
 const mapStateToProps = (state) => {
     const channelId = getChannelIdWithSettingsOpen(state);
     let channel = null;
+    let isChannelDmOrGm = false;
```
levb: naming: 1/5 `s/isChannelDmOrGm/omitDisplayName/` - more intention revealing

jfrerich: That's a much better choice.  Commit the suggestion.  Thanks, @levb

DHaussermann:
@jfrerich I'm not seeing this change. I built the plugin from `MM-19639-GM-DM-shows-channel-ID` and the subscribe modal still shows the long channel name.

ChannelName

Maybe I'm misunderstanding or this has to be tested in `feature/channel-subscriptions` branch because it relies on other changes?

mickmister:
@DHaussermann I pulled down the `MM-19639-GM-DM-shows-channel-ID` branch and built the plugin. The channel name hash is hidden for me. I wonder what's different between our environments :thinking:

DHaussermann:
@mickmister that's strange. Let me try again just to be sure I deployed the right build.

DHaussermann:
Thanks @mickmister for testing also. I do see this working. I'm not sure what happened earlier. Seems it's possible  I mistakenly uploaded the worng package 😕 .

PR #364, opened 2019-10-29
[MM-18534] Resolve fields that contradict chosen issue types

jfrerich:
## Summary
This PR resolves the following issues regarding subscription filter fields

1. Only show fields that are applicable to all selected issue types
2. If the user attempts to change the issue type after the filter field has been changed, validate that all filter fields are still applicable. IF NOT:
   1. Don't add selected issue type to the select box
   1. Show warning message listing the attempted `issue type`, the conflicting filter fields, and a suggestion to fix the fields or create a separate subscription.
3. In the event, a previously saved subscription was saved in a state that has conflicting fields, 

**Example message when a user attempts to add issue type with conflicting fields**
image

**Example messages when a previously saved subscription is loaded and contained conflicting filter fields** 
BEFORE
image

AFTER:
image

## Ticket
https://mattermost.atlassian.net/browse/MM-18534


webapp/src/components/modals/channel_settings/channel_settings_filters.tsx
```diff
@@ -60,50 +62,22 @@ export default class ChannelSettingsFilters extends React.PureComponent<ChannelS
         this.props.onChange(newValues);
     };
 
-    getConflictingFields = (): {field: FilterField; issueTypes: IssueType[]}[] => {
```
jfrerich: This function was moved to the utils file because it is used in two separate component files.

levb:
@jfrerich very happy to see this submitted! thank you! Since it's all JS, @cpoile can you please review it instead of me?

A general comment about our Jira plugin complex logic. 0/5 we should start thinking about our own "fields" helper library that would deal with types, etc. In golang, on the server. And then start exposing more targeted "auto-complete"-like APIs for dynamically populating the drop-downs, and move all this complexity away from JS.

The fields logic will keep popping up all over the place, we should be thinking about DRY-ing, testing, and simplifying. Go > JavaScript.

DHaussermann:
@jfrerich  Looks like once a filter is already in invalid state it won't always tell me why.
InvalidFilterConfiguration
Is it possible in this case it's issue specific to filters?
Testing for this is ongoing.

DHaussermann:
Just a follow up on this @jfrerich - The issue seems to be specific to Jira Server. It is not related to filters. I can no longer create subscription at all on server even if they contain no filters. See below. I get this message for any issue type I attempt to select. 
NoSaveSubscription

webapp/src/components/modals/channel_settings/edit_channel_settings.tsx
```diff
@@ -141,6 +149,46 @@ export default class EditChannelSettings extends PureComponent<Props, State> {
...
+
+    handleIssueChange = (id: keyof ChannelSubscriptionFilters, value: string[]) => {
+        let finalValue = value;
+        if (!finalValue) {
+            finalValue = [];
+        } else if (!Array.isArray(finalValue)) {
+            finalValue = [finalValue];
+        }
+        const filters = {...this.state.filters};
+        filters[id] = finalValue;
```
mickmister: Since we know `issue_types` is using a select with `isMulti: true`, this first block can be shorted to:
```suggestion
		handleIssueChange = (id: keyof ChannelSubscriptionFilters, value: string[] | null) => { // react-select will pass null if no issue_types are selected
		const finalValue = value || [];
        const filters = {...this.state.filters, issue_types: finalValue};
```

webapp/src/components/modals/channel_settings/edit_channel_settings.tsx
```diff
@@ -141,6 +149,46 @@ export default class EditChannelSettings extends PureComponent<Props, State> {
...
+
+        let conflictingFields = null;
+        if (value) {
```
mickmister: 0/5 on this one
```suggestion
        if (finalValue.length) {
```

webapp/src/components/modals/channel_settings/edit_channel_settings.tsx
```diff
@@ -141,6 +149,46 @@ export default class EditChannelSettings extends PureComponent<Props, State> {
...
+            const filterFields = getCustomFieldFiltersForProjects(this.props.jiraIssueMetadata, this.state.filters.projects);
+            conflictingFields = getConflictingFields(
+                filterFields,
+                value,
```
mickmister: For type safety
```suggestion
                finalValue,
```

mickmister:
> Example messages when a previously saved subscription is loaded and contained conflicting filter fields

Text is white and message is:
`Error: Change type does not ...`

I think we should instead match the other error case:
- Text color to be red
- Remove `Error: ` prefix

webapp/src/components/modals/channel_settings/edit_channel_settings.tsx
```diff
@@ -156,27 +156,22 @@ export default class EditChannelSettings extends PureComponent<Props, State> {
...
-        if (!finalValue) {
-            finalValue = [];
-        } else if (!Array.isArray(finalValue)) {
-            finalValue = [finalValue];
-        }
-        const filters = {...this.state.filters};
+    handleIssueChange = (id: keyof ChannelSubscriptionFilters, value: string[] | null) => {
+        const finalValue = value || [];
+        const filters = {...this.state.filters, issue_types: finalValue};
         filters[id] = finalValue;
```
mickmister: We can remove this line since we are setting the `issue_types` property above.

mickmister:
@DHaussermann Can you try the testing again?

DHaussermann:
@mickmister I see an error when I try the full make that says `info lifecycle webapp@0.0.1~lint: Failed to exec lint script` among other things. 
For Now I just used `make dist`and will test this again

mickmister:
Currently up on

https://mm-18534.test.mattermost.cloud
sysadmin
Sys@dmin123

webapp/src/components/modals/channel_settings/__snapshots__/edit_channel_settings.test.tsx.snap
```diff
@@ -186,10 +186,38 @@ exports[`components/EditChannelSettings SERVER - should match snapshot after fet
...
+            Object {
+              "label": "Issue Updated: Custom - Issue Type",
+              "value": "event_updated_Issue Type",
+            },
+            Object {
+              "label": "Issue Updated: Custom - Linked Issues",
+              "value": "event_updated_Linked Issues",
+            },
+            Object {
+              "label": "Issue Updated: Custom - Project",
```
mickmister: Need to investigate why Project and other fields around this diff are showing up as available events

mickmister: This is only happening for Jira server, so creating a ticket as it is not blocking for community. https://mattermost.atlassian.net/browse/MM-19810

jfrerich:
> > Example messages when a previously saved subscription is loaded and contained conflicting filter fields
> 
> Text is white and message is:
> `Error: Change type does not ...`
> 
> I think we should instead match the other error case:
> 
> * Text color to be red
> * Remove `Error: ` prefix

Addressed these changes in the latest commit.

mickmister:
@aaronrothschild @cpoile Do you mind giving this a review?

aaronrothschild:
@jfrerich if you tab to the second field, it appears like this:
image

If you add a value twice and hit enter it clears the row for you:


Otherwise, errors seem good - didn't encounter much else.

mickmister:
> If you add a value twice and hit enter it clears the row for you:

@aaronrothschild I'm starting work for this issue: [MM-19791](https://mattermost.atlassian.net/browse/MM-19791)

asaadmahmood:
The error text needs to be aligned.
Screenshot 2019-11-01 at 7 05 16 AM

jfrerich:
@asaadmahmood, can you pull the latest commit?  I have fixes for those messages already committed.  You should see the message aligned, colored red, and the "Error:" prefix removed from the message.

jfrerich:
For reference, this is the commit.

https://github.com/mattermost/mattermost-plugin-jira/pull/364/commits/9bf2d50a359a7f39dde9563a243e21a869a882ef

mickmister:
@jfrerich I think he's referring to your screenshot

jfrerich:
@asaadmahmood, sorry, I thought what you were viewing was created from running with this PR's changes and didn't realize you were just referencing the screenshot.  I've updated the initial PR with a screenshot displaying the newest changes.

asaadmahmood:
@mickmister @jfrerich Yup, I was looking at the screenshot.
If there are any changes made, we should update the screenshot so we don't have to download and run the changes locally, and can just easily comment on the screenshot.

mickmister:
@cpoile could you give this a review when you have a chance?

cpoile:
Any width smaller than 1200px causes the "x" close button to overlap the create subscription button:
image

mickmister:
@cpoile Interesting, that should have been fixed with https://github.com/mattermost/mattermost-plugin-jira/pull/361 
cc @DHaussermann so you're aware

cpoile:
@mickmister Ah, I was using your test server above, so that wouldn't have been merged when you spun that up. :)

mickmister:
Right, I've updated that a few times but definitely not recent enough! There are tickets open for automating this process https://mattermost.atlassian.net/browse/MM-19541
https://mattermost.atlassian.net/browse/MM-19546 . Do you want me to update it again?

jfrerich:
Thanks @cpoile!

PR #381, opened 2019-11-11
[MM-19689] Improve `/jira subscribe list` UX

jfrerich:
## Summary

This PR improves the output of the `/jira subscribe list` command output by grouping channel subscriptions by team.  For DM and GM channels, post the hashes under the Generic Headline "Group and Direct Messages".

The following is directly from the commit logs, but is a comprehensive list of the changes made and testing needed for QA.

Group channels by Team Name
Include Group and Direct Message channels exposing hashes for channel names

Add tests for the following:
- team and DM/GM channels subscriptions
- team only channel subscriptions
- DM/GM only channel subscriptions
- Team Heading level and bold of channel names
- Subscription that was not previously saved with a name
- Zero subscriptions on any channels
- Team names are sorted alphabetically

## Ticket 
https://mattermost.atlassian.net/browse/MM-19689

## Screenshots
image

### DIFF
--- a/server/subscribe.go
+++ b/server/subscribe.go
@@ -53,6 +53,11 @@ type ChannelSubscriptions struct {
 	IdByEvent     map[string]StringSet           `json:"id_by_event"`
 }
 
+type TeamSubscriptions struct {
+	TeamId string
+	SubIds []string
+}
+
 func NewChannelSubscriptions() *ChannelSubscriptions {
 	return &ChannelSubscriptions{
 		ById:          map[string]ChannelSubscription{},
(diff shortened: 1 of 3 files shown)


server/subscribe.go
```diff
@@ -360,22 +365,95 @@ func (p *Plugin) listChannelSubscriptions() (string, error) {
...
+
+			channel, appErr := p.API.GetChannel(sub.ChannelId)
+			if appErr != nil {
+				return "", errors.New("Failed to get channel")
+			}
+
+			// only print channel name once for all subscriptions
+			if printChannelName {
+				rows = append(rows, fmt.Sprintf("* **~%s** (%d):", channel.Name, len(teamSubs.SubIds)))
+				printChannelName = false
```
mickmister: I don't think this loop separates the sub ids between different channels. Wouldn't this code only print out the channel for the first subscription? `printChannelName` is never set to `true` except for just before this loop begins. And this loop is for all subscription ids for a given team, not grouped by channel.

I see that you only have one channel with subscriptions per team in your screenshot. Do you mind writing a unit test with two channels in the same team?

server/subscribe.go
```diff
@@ -360,22 +365,95 @@ func (p *Plugin) listChannelSubscriptions() (string, error) {
...
+
+		// create team header for channels.  Only print header for DMs and GMs channels once.
+		if teamSubs.TeamId != "" {
+			channelTeam, appErr := p.API.GetTeam(teamSubs.TeamId)
+			if appErr != nil {
+				return "", appErr
+			}
+			rows = append(rows, fmt.Sprintf("\n### %s", channelTeam.DisplayName))
+		} else {
+			if printDMGMHeader {
```
mickmister: Why is this boolean necessary? Do we traverse into this `else` more than once throughout the loop?

server/subscribe.go
```diff
@@ -360,22 +365,95 @@ func (p *Plugin) listChannelSubscriptions() (string, error) {
...
+	}
+
 	rows := []string{}
+	printDMGMHeader := true
+
+	for _, teamSubs := range teamDMGMSubs {
+
+		// create team header for channels.  Only print header for DMs and GMs channels once.
+		if teamSubs.TeamId != "" {
+			channelTeam, appErr := p.API.GetTeam(teamSubs.TeamId)
```
mickmister:
```suggestion
			team, appErr := p.API.GetTeam(teamSubs.TeamId)
```

jfrerich: Done.

server/subscribe.go
jfrerich: Confirmed this issue occurs.  I'm reworking the code to account for multiple channels with subscriptions in the same team.

server/subscribe.go
```diff
@@ -360,22 +365,95 @@ func (p *Plugin) listChannelSubscriptions() (string, error) {
...
+				subName = sub.Name
+			}
+			rows = append(rows, fmt.Sprintf("  * %s - %s", sub.Filters.Projects.Elems()[0], subName))
+		}
+	}
+
+	return strings.Join(rows, "\n"), nil
+}
+
+func (p *Plugin) getSubscriptionsOrderedByTeamDMandGM() ([]TeamSubscriptions, error) {
```
levb: 2/5 s/getSubscriptionsOrderedByTeamDMandGM/getSubscriptionsSorted/  - the original name is difficult to understand without the context. For now, is there another use-case where we sort the list differently?

server/subscribe.go
```diff
@@ -360,22 +365,95 @@ func (p *Plugin) listChannelSubscriptions() (string, error) {
 		return "", err
 	}
 
+	teamDMGMSubs, err := p.getSubscriptionsOrderedByTeamDMandGM()
```
levb: 0/5 `s/teamDMGMSubs/subs/`

jfrerich:
@aaronrothschild, I added your text suggestion and the following for the case that an MM instance does not have any channel subscriptions.

https://github.com/mattermost/mattermost-plugin-jira/pull/381/files#diff-2e60f7eae1ef5d6971c9403310a9bdb5R378

server/subscribe.go
```diff
@@ -360,22 +367,140 @@ func (p *Plugin) listChannelSubscriptions() (string, error) {
...
+		return "", err
+	}
+
 	rows := []string{}
+
+	if len(sortedSubs) == 0 {
+		rows = append(rows, fmt.Sprintf("There are currently no channels subcriptions to Jira notifications. To add a subscription, navigate to a channel and type `/jira subscribe`\n"))
+		return strings.Join(rows, "\n"), nil
+	}
+	rows = append(rows, fmt.Sprintf("The following channels have subcribed to Jira notifications. To modify a subscription, navigate to the channel and type `/jira subscribe`"))
```
mickmister:
```suggestion
	rows = append(rows, fmt.Sprintf("The following channels have subscribed to Jira notifications. To modify a subscription, navigate to the channel and type `/jira subscribe`"))
```

server/subscribe.go
```diff
@@ -360,22 +367,140 @@ func (p *Plugin) listChannelSubscriptions() (string, error) {
...
+	}
+
+	// sort by combined team name/id values
+	sort.Strings(nameIdArray)
+
+	// save array of subs, sorted by team and grouped by channel
+	for _, teamId := range nameIdArray {
+
+		teamName := strings.Split(teamId, separator)[0]
+		teamId = strings.Split(teamId, separator)[1]
```
mickmister: Instead of using a string with a separator within it, which is sort of cryptic, can we use a struct that contains both values we want to extract?

server/subscribe.go
mickmister: Did you include the fix for this? Could you write a unit test for this case?

server/subscribe.go
jfrerich: Yep.  After looking at it for a while, instead of creating any new `type`, I just save the `model.Team` types into an array.  Also, it isn't as easy to sort by struct fields in go, but found the included methodology, which was in the go documentation.  It leverages the `sort` library, but notice it requires satisfying the sort interface which means the `Len, Swap, and Less` methods must be present. 

Overall, it is more code but doesn't have the hacky string separator from before. Let me know what you think.

server/subscribe.go
mickmister: Will re-review once this is addressed @jfrerich

server/subscribe.go
jfrerich: This has been completely removed from the function.  The new methodology has the `GetSubscriptions()` method do all of the work so that `listChannelSubscriptions` just loops through the data.

server/subscribe.go
jfrerich: @mickmister, Tests added in last two commits

jfrerich:
@mickmister, This is ready for a re-review.

server/subscribe.go
```diff
@@ -360,22 +367,167 @@ func (p *Plugin) listChannelSubscriptions() (string, error) {
 		return "", err
 	}
 
+	sortedSubs, err := p.getSubscriptionsSorted()
```
levb: sorry to be a nitpicker, but for consistency (and readability) would it be better `getSortedSubscriptions`?

server/subscribe.go
mickmister: Sounds good to me. Re-requesting @levb, to review these changes.

server/subscribe.go
jfrerich: I thought about this for a bit and I agree, this should be changed, mainly for readability.  It's clearer with <adjective><nown>

jfrerich:
@mickmister, I just checked in the commit regarding the changes we just made.  Thank you sir!

DHaussermann:
@jfrerich There is an issue with the way the linking works for subscriptions on channels that are not on the current team you're viewing.  
If the current team has channel with the same name, the ~ link will be clickable and direct the user to the channel of the same name on the current channel instead. 

This is confusing to use. Any channels not on the current team should no be listed as a link.

mickmister:
> Any channels not on the current team should no be listed as a link.

@jfrerich I'm thinking we should not use the `~` for the channels in a different team the one the user is currently viewing. Or put a space between the `~` and the channel name. Both solutions will make the channel name render as just text instead of a clickable link.

jfrerich:
image

image

jfrerich:
@DHaussermann, @mickmister, I posted the two options above and personally favor the one that doesn't include the `~` if the channel is no on the team.  My thinking is that someone will claim it as a bug, even though it is intentional.

jfrerich:
@mickmister, I re-requested your review because I made code changes.  

@DHaussermann, if you re-sync with this branch, you should see the changes noted in the images above.

PR #399, opened 2019-11-27
[MM-20744] Remove the full screen modal css from Jira and reuse the css in the webapp

cpoile:
#### Summary
- Reduce the custom fullscreen modal scss to the bare minimum we need for Jira. 
- Changing go.mod for go1.13 (it wouldn't compile for me without it)
- @jfrerich I removed the z-index change, but I'm not sure what this will break. Can you let me know what I need to test? If we need to, we can add our own class and create a special jira-plugin-fullscreen-modal, that would fix the problem that this caused for Asaad (see the Jira ticket).
- @asaadmahmood what do you think, just make our own fullscreen modal class for this css we need?

#### Links
- Fixes [MM-20744](https://mattermost.atlassian.net/browse/MM-20744)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/hashicorp/go-plugin v1.0.1 // indirect
 	github.com/lib/pq v1.1.1 // indirect
 	github.com/mattermost/go-i18n v1.11.0 // indirect
-	github.com/mattermost/mattermost-server v5.12.0+incompatible
+	github.com/mattermost/mattermost-server v1.4.1-0.20191019085930-e13dad9fd4b6
 	github.com/pelletier/go-toml v1.4.0 // indirect
 	github.com/pkg/errors v0.8.1
 	github.com/rbriski/atlassian-jwt v0.0.0-20180307182949-7bb4ae273058
--- a/go.sum
+++ b/go.sum
@@ -20,7 +20,6 @@ github.com/alecthomas/units v0.0.0-20151022065526-2efee857e7cf/go.mod h1:ybxpYRF
 github.com/andygrunwald/go-jira v1.10.0 h1:+HPPK7++6/hW8ygtr2Yc0wd+Qu139NrWiTD/r1cYxO0=
 github.com/andygrunwald/go-jira v1.10.0/go.mod h1:KEsrADP1cEXRxVWTaDtpLyyZN1LM9p6Jn8W5+sDzxhc=
 github.com/anmitsu/go-shlex v0.0.0-20161002113705-648efa622239/go.mod h1:2FmKhYUyUczH0OGQWaF5ceTx0UBShxjsH6f8oGKYe2c=
-github.com/apache/thrift v0.0.0-20180902110319-2566ecd5d999/go.mod h1:cp2SuWMxlEZw2r+iP2GNCdIi4C1qmUzdZFSVb+bacwQ=
 github.com/apache/thrift v0.12.0/go.mod h1:cp2SuWMxlEZw2r+iP2GNCdIi4C1qmUzdZFSVb+bacwQ=
 github.com/armon/consul-api v0.0.0-20180202201655-eb2c6b5be1b6/go.mod h1:grANhF5doyWs3UAsr3K4I6qtAmlQcZDesFNEHPZAzj8=
 github.com/armon/go-metrics v0.0.0-20180917152333-f0300d1749da/go.mod h1:Q73ZrmVTwzkszR9V5SSuryQ31EELlFMUz1kKyl939pY=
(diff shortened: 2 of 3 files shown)


mickmister:
@asaadmahmood Some full screen modal css is not the same as the webapp's. Specifically the positioning of the "back" and "close-x" buttons. These two buttons have been moved closer to top two corners of the modal, to take into account the placement of the form in the modal.

How do you suggest we consolidate these changes while reusing the existing css? One way would be to override only the necessary attributes for the css classes used by the buttons, or use different classes for the two buttons.

asaadmahmood:
@mickmister Another way would be to include that css in the webapp too, that can be used with another modifier class for those cases.

cpoile:
@asaadmahmood In general, we shouldn't include Jira-plugin-specific css in the webapp, right? So what do you think we could do here? Could we add a jira-modal class?

asaadmahmood:
@cpoile We shouldn't use jira specific css anywhere, not even in the plugin, if the full screen modal has a different state with the back button/close button that is a bit compact, then that should just be another variation for the modal that can be used for other plugins or other states for the full screen modal in our app.

cpoile:
@mickmister @jfrerich Since you were the ones working on this, can we remove these changes and just use the standard full screen modal css?

mickmister:
I agree with @asaadmahmood with making the css in the webapp "configurable", in that we can specify different available values based on combinations of classes. For example, we use something like `margin-bottom x3` when we want to apply a certain amount of margin.

I think we can use something like `close-x modal-corner` (not a good name example, however) to specify that we want the styles that place the buttons closer to the upper corners of the modal.

> can we remove these changes and just use the standard full screen modal css?

@cpoile Other than the changes to the "back" and "close" buttons to be closer to the upper corners of the modal, none of this css is different from what was taken from the `docup` plugin, which I believe was copied directly from the webapp. So most of the css can be used straight from the webapp, but we still need to allow the plugin to express the intent to place the buttons close to the upper corners of the modal.

cpoile:
@jfrerich Just a reminder about the z-index change -- I removed it because that's what was causing Asaad's bug. I want to make sure this doesn't break anything.

cpoile:
@asaadmahmood Looking over the changes, I'm not sure if should put this in the webapp. The change seems too few and too specific for this use case. Would you want to add in parameters to the css on the webapp like @mickmister suggested?

Otherwise, we could make a `jira-modal` class, and use that for these little changes? Your call.

jfrerich:
@cpoile, I've been looking for the code change regarding the zIndex, but so far have only found this https://github.com/mattermost/mattermost-plugin-jira/commit/d99e966f65c3b4a2c3b3e450020b55594bbf1abb#diff-9b097f02b4ddd564b16955991958f147R172, which is a change I had made in the `confirm_modal`.

jfrerich:
I would test that the `Create Jira Issue` and `Attach to Jira Issue` post commands both show up with this change just to verify nothing breaks. The other area would be the `/jira subscribe list` command.  The test for the confirm_modal would be to create a subscription and then delete it.  That will open the confirm_modal.

levb:
Should this be going into 2.3? Or is it ok to defer until later?

cpoile:
I don't know, I've been out of Jira for so long. @mickmister ?

mickmister:
If the issue found on community is fixed without this PR, then it is not necessary for 2.3 imo. I'm not sure what issues someone may run into other than the exact scenario we experienced on community.

@asaadmahmood I think we need your judgment on this

cpoile:
@mickmister I think Asaad is out for awhile, and I'm really not sure what to do with this. I don't totally agree with moving the changes into webapp, I think we should just merge this for now and then subclass or specialize the webapp css later. It's tech debt.

asaadmahmood:
@cpanato @mickmister Yes, that issue is fixed, we don't need to take this for 2.3. I think we should give this some time before pushing it.

cpanato:
@asaadmahmood why you mention me? any actions do i need to take?

cpoile:
@cpanato No, I think he meant my name. Sorry for the confusion.

asaadmahmood:
Apologies @cpanato

mickmister:
The only gotcha I can think of with merging this is the placement of the close-x and back buttons. Pulling down from master, the positioning is incorrect again. I seem to have broken it in a completely unrelated PR https://github.com/mattermost/mattermost-plugin-jira/commit/ad3cc4c2581e73c91a9fa520a4ec7a8299a4e0b3#diff-0f8116c7c9b5c1bb6abab32cd1b3cd5f . I will create a new PR for the button positioning fix.

DHaussermann:
@asaadmahmood or @mickmister  I'm a bit confused about how to validate that the issue is resolved **without this PR**. Is it simply to validate the modal is working correctly where https://github.com/mattermost/mattermost-webapp/pull/4351 has been applied?

DHaussermann:
@cpoile wonder if you might also be able to answer the question above ☝️ about validating without this PR.

I never actually saw the issue occur on community but I could try testing on a different server.

cpoile:
@DHaussermann I'm not sure what you're referring to -- the issue that Asaad was originally having was fixed by https://github.com/mattermost/mattermost-webapp/pull/4351, are you talking about verifying that /that PR/ fixes the problem?

DHaussermann:
@cpoile yes I want to make sure that  mattermost/mattermost-webapp#4351 fixes the issue without the need for this PR but, I'm not sure what this bug looks like when it's occurring.

cpoile:
@asaadmahmood Can you help @DHaussermann reproduce the original bug you were seeing? Thanks!

asaadmahmood:
@DHaussermann You can view the tooltip on the plugins marketplace modal on https://ci-extensions.azure.k8s.mattermost.com/

asaadmahmood:
Screenshot 2019-12-05 at 7 48 29 PM

levb:
@mickmister @jfrerich ping - is this waiting on you?

cpoile:
@mickmister With the changes you made to the css in the other PR, where does that leave us with this? Either we add those to the webapp css (not a big fan of that), or we subclass our own modals (jira-fullscreen-modal or something)?

jfrerich:
@cpoile, with the `z-index` removal, were you able test the `Create Jira Issue` and `Attach to Jira Issue` post commands and the confirm modal from trying to delete a subscription the `/jira subscribe list` command?

mickmister:
Per @asaadmahmood's comment above, I think we are not going to change anything more pertaining to this css for 2.3. We haven't reached an agreement on how to resolve the css duplication.

> we subclass our own modals (jira-fullscreen-modal or something)

@cpoile This sounds good, but I don't think I understand it fully. Can you explain what you mean by subclass? We cannot import any webapp components into the plugin fyi, only reuse css. I'm not sure if that's relevant.

cpoile:
@mickmister I guess I mean use our own class, instead of shadowing/redeclaring/?? `.FullScreenModal`. (got to brush up on my scss terms)

cpoile:
Tested what @jfrerich mentioned, everything works after removing the z-index.
@mickmister Your choice on what we should do, design-wise.

mickmister:
Closing in favor of https://github.com/mattermost/mattermost-plugin-jira/pull/413

PR #407, opened 2019-12-05
Updating jira icon alignment in menu

asaadmahmood:
Needs to go in for jira 2.3.
Fixes the alignment for the icons.
This class should be used now as a wrapper for all plugins that introduce icon in the menu dropdown.

Screenshot 2019-12-05 at 6 05 07 PM


levb:
Do we need this fix in the GitHub plugin too?

asaadmahmood:
@levb yup.

PR #410, opened 2019-12-10
MM-20928 Fixed hostname too long in stats key

levb:
The hostnames on community-daily and community-release proved to be too long for the key to be included directly. Hashing them now.

Fixes https://mattermost.atlassian.net/browse/MM-20928

### DIFF
--- a/server/stats.go
+++ b/server/stats.go
@@ -1,8 +1,10 @@
 package main
 
 import (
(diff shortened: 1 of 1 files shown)


cpoile:
@levb was going to ask for a test, but I don't think it would help here.

PR #437, opened 2020-01-05
MM-16948 Check if Jira server url actually is a Jira Cloud URL

popstr:
#### Summary

If Jira instance is being installed as a server instance but the URL looks like a Jira Cloud URL, return a friendly error message with a hint.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-jira/issues/425

### DIFF
--- a/server/utils/utils.go
+++ b/server/utils/utils.go
@@ -55,3 +55,11 @@ func Map(vs []string, f func(string) string) []string {
 	}
 	return vsm
 }
+
+func IsJiraCloudURL(jiraURL string) (bool, error) {
+	u, err := url.Parse(jiraURL)
+	if err != nil {
+		return false, err
+	}
+	return strings.HasSuffix(u.Hostname(), ".atlassian.net"), nil
+}
--- a/server/command.go
+++ b/server/command.go
@@ -360,6 +360,13 @@ func executeInstallServer(p *Plugin, c *plugin.Context, header *model.CommandArg
 	if err != nil {
 		return p.responsef(header, err.Error())
 	}
+	isJiraCloudURL, err := utils.IsJiraCloudURL(jiraURL)
+	if err != nil {
+		return p.responsef(header, err.Error())
+	}
+	if isJiraCloudURL {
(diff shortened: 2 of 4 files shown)


server/utils/utils.go
```diff
@@ -55,3 +55,11 @@ func Map(vs []string, f func(string) string) []string {
 	}
 	return vsm
 }
+
+func IsJiraCloudURL(jiraURL string) (bool, error) {
+	u, err := url.Parse(jiraURL)
+	if err != nil {
+		return false, err
+	}
+	return strings.Contains(u.Hostname(), ".atlassian.net"), nil
```
levb: 1/5 maybe, `...HasSuffix` is more appropriate?

levb:
Also, @popstr you should run `gofmt -w` on your changes - `lint` is failing currently. I recommend configuring your editor so that it's done automatically on save for your go files.

server/utils/utils.go
popstr: You are probably right. I used HasSuffix at first but it doesn't catch when you accidentally write something after ".net". Maybe we should sanity check the domain name, but that's a different issue.

popstr:
Added a test for jira installation since adding code in the `command.executeInstallServer` function caused the test coverage to drop.

server/utils/utils.go
```diff
@@ -55,3 +55,11 @@ func Map(vs []string, f func(string) string) []string {
 	}
 	return vsm
 }
+
+func IsJiraCloudURL(jiraURL string) (bool, error) {
+	u, err := url.Parse(jiraURL)
+	if err != nil {
+		return false, err
+	}
+	return strings.HasSuffix(u.Hostname(), ".atlassian.net"), nil
```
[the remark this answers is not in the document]
popstr: I was testing inputs like `test.atlassial.net'`, with a ' at the end. `.Hostname()` includes the ' and therefore it interprets the url as a Jira Server url (since it doesn't end with ".atlassian.net"). This is unfortunate, but acceptable I guess. We could introduce a host name syntax checker, but that would be a different task.

PR #459, opened 2020-01-29
MM-18075 Reuse output of `/jira view` in the output of `/jira create`

iamsayantan:
#### Summary
This pull request reuses code which generates view for `/jira view`  command and produces same view for `/jira create` as an ephemeral post visible only to the person who created the issue.
MM-18075
 

#### Ticket Link

Fixes [#420](https://github.com/mattermost/mattermost-plugin-jira/issues/420)
Jira: [https://mattermost.atlassian.net/browse/MM-18075](https://mattermost.atlassian.net/browse/MM-18075)


server/issue.go
```diff
@@ -168,20 +168,24 @@ func httpAPICreateIssue(ji Instance, w http.ResponseWriter, r *http.Request) (in
...
+
+	attachment, err := ji.GetPlugin().getIssueAsSlackAttachment(ji, jiraUser, created.Key)
+	if err != nil {
 		return http.StatusInternalServerError,
-			errors.WithMessage(appErr, "failed to create notification post "+create.PostId)
+			errors.WithMessage(err, "failed to create notification post "+create.PostId)
 	}
 
+	reply.AddProp("attachments", attachment)
+	_ = api.SendEphemeralPost(mattermostUserId, reply)
```
mickmister: @aaronrothschild Do we want to show the other people in the channel this formatted output? Currently, this PR's code changes the output to be an ephemeral post instead of a public post.

@iamsayantan I'm thinking we should post the expanded version to the channel. Either that, or keep the original functionality in tact, and post the expanded one with an ephemeral post. Could you post a screenshot with the ticket's description collapsed (so it says "Show More"), to show a different context?

I'll defer to @aaronrothschild on the decision.

iamsayantan: @mickmister The Jira ticket mentioned to create an ephimeral post.

> Change the `/jira create` feature to also show the “view issue” of the newly created issue as an ephemeral message visible to the person who just created the issue (and also display the URL to Jira).

mickmister: Thanks for pointing this out. What do you think about creating a public post and an ephemeral post? The public post could be:

> Created an issue (issue key) - `(summary)`

iamsayantan: Okay, I will work on it and then decide if it looks good.

iamsayantan: @mickmister Have a look and let me know your thoughts. If this looks good, I'll update the pull request.

issue_message

And this one is with colapsed issue description.
long_issue

jfrerich: I like this approach.  @iamsayantan, in the permanent post, can you link to the ticket, similar to the ephemeral post does?

iamsayantan:
I've updated the pull request. This is how it looks.

issue_post_with_link

aaronrothschild:
> I've updated the pull request. This is how it looks.
> 
> issue_post_with_link

@iamsayantan @mickmister I like how it posts both publicly and privately with information. Good stuff @iamsayantan !

@mickmister In the future, perhaps we can consider making the public post use the "Meta-Data in RHS Pane" to display extra information about the Jira ticket in the RHS - so it stays slim in the channel, but can be expanded to be viewed in the RHS optionally.  

Example:
2020-02-10_15-21-55 copy

The functionality I'm referring to is here Under props:card - https://developers.mattermost.com/integrate/incoming-webhooks/#parameters

server/issue.go
```diff
@@ -168,18 +168,44 @@ func httpAPICreateIssue(ji Instance, w http.ResponseWriter, r *http.Request) (in
...
+	publicReply := &model.Post{
+		Message:   fmt.Sprintf("Created a Jira issue: %s", mdKeySummaryLink(createdIssue)),
+		ChannelId: channelId,
+		RootId:    rootId,
+		ParentId:  rootId,
 		UserId:    mattermostUserId,
 	}
-	_, appErr = api.CreatePost(reply)
-	if appErr != nil {
+	_, postErr := api.CreatePost(publicReply)
```
jfrerich: 3/5 This seems nit-picky, but can you keep the err named `appErr`. This is the naming convention we use for errors with the `*model.AppError` type

iamsayantan: Updated.

server/issue.go
```diff
@@ -168,18 +168,44 @@ func httpAPICreateIssue(ji Instance, w http.ResponseWriter, r *http.Request) (in
...
+		RootId:    rootId,
+		ParentId:  rootId,
 		UserId:    mattermostUserId,
 	}
-	_, appErr = api.CreatePost(reply)
+	_, appErr = api.CreatePost(publicReply)
 	if appErr != nil {
 		return http.StatusInternalServerError,
-			errors.WithMessage(appErr, "failed to create notification post "+create.PostId)
+			errors.WithMessage(err, "failed to create notification post "+create.PostId)
```
jfrerich: This `err` should be `appErr`.  `err` is actually returned from `GetIssue()` in the code above.

server/issue.go
```diff
@@ -168,18 +168,44 @@ func httpAPICreateIssue(ji Instance, w http.ResponseWriter, r *http.Request) (in
...
+	if err != nil {
+		return http.StatusInternalServerError,
+			errors.WithMessage(err, "failed to create notification post "+create.PostId)
+	}
+
+	reply.AddProp("attachments", attachment)
+	_ = api.SendEphemeralPost(mattermostUserId, reply)
+
+	// Fetching issue details as Jira only returns the issue id and issue key at the time of
+	// issue creation. We will not have issue summery in the creation response.
```
mickmister: Nit for spelling:
```suggestion
	// issue creation. We will not have issue summary in the creation response.
```

server/issue.go
iamsayantan: That was a very silly overlooking on my part. Updated.

jfrerich: No problem!  Don't consider these PR reviews as being critical, but ways to catch simple items that everyone misses and ways to communicate thoughts!

server/issue.go
```diff
@@ -169,18 +169,44 @@ func httpAPICreateIssue(ji Instance, w http.ResponseWriter, r *http.Request) (in
 		return http.StatusInternalServerError, errors.Errorf("Failed to create issue. %s", err.Error())
 	}
 
+	// Reply with an ephemeral post with the Jira issue formatted as slack attachment.
 	startLink := fmt.Sprintf("/plugins/%s%s", manifest.Id, routeUserStart)
 	msg := fmt.Sprintf("Created Jira issue [%s](%s/browse/%s) by [mattermost-jira-plugin](%s)", created.Key, ji.GetURL(), created.Key, startLink)
-
-	// Reply to the post with the issue link that was created
+
```
mickmister: Looks like there is an extra space on this line that is causing CI to fail. @iamsayantan Do you mind removing the space?
```suggestion

```

iamsayantan: Done.

Issue #92, opened 2019-05-28
New command `/jira unassign <issue-key>`


jasonblais:
If you're interested please comment here and come [join our "Contributors" community channel](https://community.mattermost.com/core/channels/tickets) on our daily build server, where you can discuss questions with community members and the Mattermost core team. For technical advice or questions, please  [join our "Plugin: Jira" community channel](https://community.mattermost.com/core/channels/jira-plugin).

New contributors please see our [Developer's Guide](https://developers.mattermost.com/contribute/getting-started/) and our [Plugins Guide](https://developers.mattermost.com/extend/plugins/).

----

For a sample PR that added "/jira assign <issue-key> <assignee>" command, see https://github.com/mattermost/mattermost-plugin-jira/pull/60

A) Add "/jira unassign <issue-key>" command. <issue-key> refers to the Jira issue key such as "MM-14151"

When successful, the command does two things:
1. removes the assignee of a Jira issue.
2. sends an ephemeral message back to the channel the user posted their command with text

```
Unassigned Jira issue <issueKey>
```

B) The following is a list of error cases and the returned ephemeral message. The ephemeral message should have the same profile picture and username as other messages posted by the Jira plugin.

1. if user doesn't have permissions for the action, return ephemeral message "You do not have the appropriate permissions to perform this action. Please contact your Jira administrator.
2. if "/jira unassign" is posted, return ephemeral message "Please specify an issue key in the form `/jira unassign <issue-key>`"
3. if <issue-key> is not found, return ephemeral message "We couldn't find the issue key. Please confirm the issue key and try again."

imazik:
hi @levb i would like to take it

imazik:
@levb I am facing issue with make in my local setup. getting this error
```
go: github.com/cpoile/oauth2@v0.0.0-20190717235208-10945bd: invalid pseudo-version: revision is shorter than canonical (10945bd219c7)
rm -rf server/dist;
mkdir -p server/dist;
cd server && env GOOS=linux GOARCH=amd64 /snap/bin/go build  -ldflags '-X "main.BuildDate=Sat Oct 19 21:20:42 UTC 2019" -X "main.BuildHash=8d1a96baeb46dd18eca50511a6cc49b4892be1a8" -X "main.BuildHashShort=8d1a96b"' -o dist/plugin-linux-amd64;
[...]

PR #460, opened 2020-01-29
Add unassign slash command

jfrerich:
#### Summary
This PR adds the `/jira unassign` slash command.

#### Command testing cases
A) When successful, the command does two things:
- [x] removes the assignee of a Jira issue.
- [x] sends an ephemeral message back to the channel the user posted their command with text
		Unassigned Jira issue <issueKey>

B) The following is a list of error cases and the returned ephemeral message. The ephemeral message should have the same profile picture and username as other messages posted by the Jira plugin.

- [x] if user doesn't have permissions for the action, return ephemeral message "You do not have the appropriate permissions to perform this action. Please contact your Jira administrator.
- [x] if "/jira unassign" is posted, return ephemeral message "Please specify an issue key in the form /jira unassign <issue-key>"
- [x] if it is not found, return ephemeral message "We couldn't find the issue key. Please confirm the issue key and try again."

#### Ticket Link
https://github.com/mattermost/mattermost-plugin-jira/issues/92

### DIFF
--- a/server/issue.go
+++ b/server/issue.go
@@ -739,6 +739,9 @@ func (p *Plugin) unassignJiraIssue(mmUserId, issueKey string) (string, error) {
 	}
 
 	if err := client.UpdateAssignee(issueKey, &jira.User{}); err != nil {
+		if StatusCode(err) == http.StatusForbidden {
+			return "You do not have the appropriate permissions to perform this action. Please contact your Jira administrator.", nil
+		}
 		return "", err
 	}
 
--- a/server/issue.go
+++ b/server/issue.go
@@ -715,6 +715,39 @@ func (p *Plugin) getIssueAsSlackAttachment(ji Instance, jiraUser JIRAUser, issue
 	return parseIssue(issue), nil
 }
 
+func (p *Plugin) unassignJiraIssue(mmUserId, issueKey string) (string, error) {
+	ji, err := p.currentInstanceStore.LoadCurrentJIRAInstance()
+	if err != nil {
+		return "", err
+	}
+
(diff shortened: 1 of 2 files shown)


server/issue.go
```diff
@@ -715,6 +715,42 @@ func (p *Plugin) getIssueAsSlackAttachment(ji Instance, jiraUser JIRAUser, issue
...
+		return "", err
+	}
+
+	client, err := ji.GetClient(jiraUser)
+	if err != nil {
+		return "", err
+	}
+
+	// check for valid issue key
+	_, err = client.GetIssue(issueKey, nil)
```
levb: Why the GetIssue? The transition will fail if the issue doesn't exist or is privileged. Ditto for `assign`?

jfrerich: The spec that was asked in the ticket was to provide two different posts.  

1) "You do not have the appropriate permissions to perform this action. Please contact your Jira administrator.
2) "We couldn't find the issue key. Please confirm the issue key and try again."

The response from a failed transition does fail with "Issue does not exist or you do not have permission to see it.", but that didn't satisfy the requested ticket request.

levb: So in this case I don't think it's critical, but if we are doing it only to improve the error reporting, we should be doing it only in the case of an error (i.e. try to transition first, and if it fails - diagnose to get the exact reason for it. I'm ok with it as is since load/performance are not a huge concern for command.

jfrerich: Yes, that makes sense and is similar to a pythonic principle "ask for forgiveness, not permission".  In this case to save an additional `getIssue()` call, "transition first (without permission) and if error occurs, (ask forgiveness) and handle the error.

PR #465, opened 2020-02-06
Error installing Jira when key is too long

### LINKED ISSUE #330: Error installing Jira when key is too long
The `key` value that is generated to install Jira is based on the site URL. In some cases the key will become long and prevent you from installing Jira. 

The user will see a misleading error on cloud `An error occurred while provisioning the user for this app. Please try installing it again, and if then problem persists then contact Atlassian support.`
 
On server the Jira UI will show `error` with no details when attempting to create the application link 

We need a solution where the key would still be based on URL to keep it unique but enforce a maximum length.  

Steps (Cloud):
- Login as an admin to a server with a long URL such as https://mattermost-server-pr-12407.test.mattermost.cloud
- Enable the Jira Plugin and use the slash command to install Jira
- Copy the path to the resulting .json to the clip board (in this case the key would be `mattermost_https_mattermost_server_pr_12407_test_mattermost_cloud`)
- Login to Jira as an Admin and navigate to Manage apps
- Click Upload apps and enter the .json
**Observed:** jira UI shows the error `An error occurred while provisioning the user for this app. Please try installing it again, and if then problem persists then contact Atlassian support.`

---

AGMETEOR:
**Summary**
The key value that is generated to install Jira is based on the site URL. In some cases the key will become long and prevent you from installing Jira.

The user will see a misleading error on cloud An error occurred while provisioning the user for this app. Please try installing it again, and if then problem persists then contact Atlassian support.

On server the Jira UI will show error with no details when attempting to create the application link

We need a solution where the key would still be based on URL to keep it unique but enforce a maximum length.

Fixes #330

### DIFF
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -14,6 +14,8 @@ import (
 	"text/template"
 	"time"
 
+	"github.com/google/uuid"
+
 	"github.com/pkg/errors"
 
 	"github.com/mattermost/mattermost-server/v5/model"
@@ -289,7 +291,14 @@ func (p *Plugin) AddAutolinks(key, baseURL string) error {
 }
 
 func (p *Plugin) GetPluginKey() string {
-	return "mattermost_" + regexpNonAlnum.ReplaceAllString(p.GetSiteURL(), "_")
+	var k string
+	k = "mattermost_" + regexpNonAlnum.ReplaceAllString(p.GetSiteURL(), "_")
+	if len(k) <= 32 {
+		return k
+	}
+
+	k = uuid.New().String()
+	return k
 }
 
 func (p *Plugin) GetPluginURLPath() string {
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -290,23 +290,31 @@ func (p *Plugin) AddAutolinks(key, baseURL string) error {
 	return nil
 }
 
-func (p *Plugin) GetPluginKey() string {
+func (p *Plugin) GetPluginKey() (string, error) {
 	var k string
+
 	key := fmt.Sprintf("%s_atlassian_key", manifest.Id)
 	k = "mattermost_" + regexpNonAlnum.ReplaceAllString(p.GetSiteURL(), "_")
 	if len(k) <= 32 {
 		p.API.KVDelete(key)
-		return k
+		return k, nil
 	}
 
 	value, err := p.API.KVGet(key)
+
+	// Empty value or error, attempt to create a new key and save it
 	if len(value) == 0 || err != nil {
 		k = uuid.New().String()
-		p.API.KVSet(key, []byte(k))
-		return k
+		e := p.API.KVSet(key, []byte(k))
+
+		if e != nil {
+			return "", err
+		}
+
+		return k, nil
 	}
 
-	return string(value)
+	return string(value), nil
 }
 
 func (p *Plugin) GetPluginURLPath() string {
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -14,8 +14,6 @@ import (
 	"text/template"
 	"time"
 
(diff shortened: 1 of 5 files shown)


server/plugin.go
```diff
@@ -289,7 +291,14 @@ func (p *Plugin) AddAutolinks(key, baseURL string) error {
...
 
 func (p *Plugin) GetPluginKey() string {
-	return "mattermost_" + regexpNonAlnum.ReplaceAllString(p.GetSiteURL(), "_")
+	var k string
+	k = "mattermost_" + regexpNonAlnum.ReplaceAllString(p.GetSiteURL(), "_")
+	if len(k) <= 32 {
+		return k
+	}
+
+	k = uuid.New().String()
```
levb: I am not sure this would work as intended. As I remember, Plugin Key is supposed to be immutable, identifying the app in Atlassian. Can you please verify its uses?

AGMETEOR: `Description: A unique key to identify the app.This key must be <= 64 characters.`
That's true. It should be unique. 

My workaround this is to use `KVSet` and then retrieve that every time. Thoughts?

levb: 0/5 The intent here was to return a stable string, that would be reasonably unique and yet somewhat user-friendly if ever displayed. It may be sufficient to perhaps return `__` + the last 30 characters of the (escaped) site URL?

A KV store record seems an overkill.

AGMETEOR: Hello @levb changed the implementation. The Jira docs say the key can be up to 64 characters long. Should we consider a bigger set if characters?

jasonblais:
@DHaussermann Friendly reminder to help with PR review :)

DHaussermann:
@AGMETEOR there seems to be an additional issue with this.

The code change you have made to the key length is working. When I look in the .json I can see that it now matches the end of the URL and the key length is now well below the max of 80.

However, this still does not allow me to  install the app in Jira. We seem to be hitting a different validation in Jira where it is looking not at the key but the whole URL.
`/name: string "Mattermost Plugin (https://longurltotestjirawithfirurllength.test.mattermost.cloud)" is too long (length: 83, maximum allowed: 80)`
JiraInstall

Please let me know if you see a solution to this issue. I'm not sure why it cares what the URL is. Maybe it's looking at the address the .json is being provided from?

DHaussermann:
Just to follow-up @AGMETEOR doe this issue require a more complex solution? Maybe a new PR needs to be open.

AGMETEOR:
> @AGMETEOR there seems to be an additional issue with this.
> 
> The code change you have made to the key length is working. When I look in the .json I can see that it now matches the end of the URL and the key length is now well below the max of 80.
> 
> However, this still does not allow me to install the app in Jira. We seem to be hitting a different validation in Jira where it is looking not at the key but the whole URL.
> `/name: string "Mattermost Plugin (https://longurltotestjirawithfirurllength.test.mattermost.cloud)" is too long (length: 83, maximum allowed: 80)`
> JiraInstall
> 
> Please let me know if you see a solution to this issue. I'm not sure why it cares what the URL is. Maybe it's looking at the address the .json is being provided from?

@DHaussermann I am making some research on why this happens. Hopefully, will get back with a solution.

DHaussermann:
@jfrerich I don't see any reason we can't merge this even thought the solution is incomplete as it does not introduce any new functional problem. We could then create a separate issue to specifically address the length of `name`. Any objection?

jfrerich:
@DHaussermann.  I'm fine with the proposal.  Let's merge this ticket and I'll create a separate ticket for the length of the name and assign to @AGMETEOR :)

jfrerich:
@DHaussermann, ticket created.  I'll merge after your PR approval.

PR #503, opened 2020-03-19
MM-21696: New config setting: Hide issue descriptions and comments

levb:
#### Summary
Added a config option to drop issue description and comment bodies from webhook and subscription posts. They should continue displaying in DMs and ephemeral posts without changes.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-21696

Config: 
image

Hide ON:
image

Hide OFF (default) as before,
image

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -79,6 +79,13 @@
         "type": "text",
         "help_text": "Additional Help Text to be shown to the user along with the output of `/jira help` command",
         "default": ""
+      },
+      {
+        "key": "HideDecriptionComment",
+        "display_name": "Hide issue descriptions and comments",
+        "type": "bool",
+        "help_text": "Hide detailed issue descriptions and comments from Subscription and Webhook messages",
+        "default": false
       }
     ],
     "footer": "Use this webhook URL format to [configure the Jira integration.](https://about.mattermost.com/default-jira-plugin)  `https://SITEURL/plugins/jira/api/v2/webhook?secret=WEBHOOKSECRET`"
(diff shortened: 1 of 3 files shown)


levb:
@mickmister @aaronrothschild We should probably add a per-subscription setting, it would be fairly trivial on the backend, but we need the UX for where it'd go in the modal, and frankly I didn't feel confident to do it timely myself.

I can quickly add the backend functionality for it in this PR, and file a HW to implement the UX, maybe? @asaadmahmood: do you mind providing a mock for where it'd go in the subscription create/edit UX?

asaadmahmood:
@levb Would this setting be configurable when creating subscriptions?
So I'm assuming the system console setting would be overriden if a person has explicitly gone with one thing over the other when creating the subscription?

levb:
> @levb Would this setting be configurable when creating subscriptions?

Yes - creating and editing

> So I'm assuming the system console setting would be overriden if a person has explicitly gone with one thing over the other when creating the subscription?

I was assuming it the other way around - the plugin-scoped "Hide" (default: off) would override all of the individual subscription settings (as a policy), but I can see it either way. 1/5 we should be "opinionated" and offer simplicity here - pick what you think is most intuitive? Let's continue in https://github.com/mattermost/mattermost-plugin-jira/issues/504.

DHaussermann:
Tested and passed
- Tested applicable webhook events to ensure new flag is respected
- Regression tested some other functionality such as "/jira view" to enure the the new flag only affects webhook events
- Tested on Cloud and Server as a precaution 
- No Issues found
LGTM!

Added release testing.

PR #521, opened 2020-04-13
npm -E audit fix in prep for 2.4 release

jfrerich:
#### Summary
The following PR (https://github.com/mattermost/mattermost-plugin-jira/pull/512) was originally created to update npm packages but has also broken the build.  This is a fresh PR to update off of the latest master build.

The command that has been run is `npm -E audit fix` which updates security issues.  

`-E` is short for `--save-exact`

### DIFF
--- a/webapp/package-lock.json
+++ b/webapp/package-lock.json
@@ -128,9 +128,9 @@
           }
         },
         "minimist": {
-          "version": "1.2.0",
-          "resolved": "https://registry.npmjs.org/minimist/-/minimist-1.2.0.tgz",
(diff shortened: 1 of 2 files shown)


jfrerich:
I hated to do this, but the previously mentioned PR https://github.com/mattermost/mattermost-plugin-jira/pull/512 wasn't building.  I also had run `npm-check -E -y` which updates dependencies (not security issues).  My guess is that we should update dependencies at some point, but it will require some massaging to get the build passing.  By updating only the security issues, this build passes.

FYI, this is the last remaining item before cutting `v2.4.0.`

Issue #549, opened 2020-05-19
Transitioning an issue from view output fails


jprusch:
With rel. 2.4 an issue can be transitioned from the output of /jira view. While transitioning still works with /jira transition slash command, using the new UI results in a "Failed to transition this issue" message,
Logging shows two error messages:

```
2020-05-19T09:15:22.359+0200    error   mlog/sugar.go:23        ERROR:  {"plugin_id": "jira", "Status": "500", "Error": "\"Zurück auf offen\" is not a valid state. Please use one of: \"Offen, Laufende Arbeiten\"", "Host": "", "RequestURI": "", "Method": "POST", "query": ""}

2020-05-19T09:15:22.361+0200    error   mlog/log.go:175 Action integration error.       {"path": "/api/v4/posts/xxxxxxxxxxxxxxxxxxxxxxx/actions/f9eienree38axx6gxou9sfuyra", "request_id": "xxxxxxxxxxxxxxxxxxxxxxxxxx", "ip_addr": "xxx.xxx.xxx.xxx", "user_id": "xxxxxxxxxxxxxxxxxxxxxxxx", "method": "POST", "err_where": "DoPostAction", "http_code": 400, "err_details": "err=json: cannot unmarshal string into Go value of type model.PostActionIntegrationResponse"}
```

My guess: Mix up between the issue state & the transition name.
In this case "Zurück auf offen" is the transition name. Valid options are "Offen","Laufende Arbeiten" which are status names.

EDIT: Anonymized data,

hanzei:
Maybe this is an umlaut issue? @jprusch Could you check if a transition without an umlaut works?

jprusch:
Sorry, this is language independent. I tried in English & also with a German transition w/o Umlauts.
```
2020-05-26T13:45:03.351+0200    error   mlog/sugar.go:23        ERROR:  {"plugin_id": "jira", "Status": "500", "Error": "\"Start Progress\" is not a valid state. Please use one of: \"In Progress, Solved, Closed\"", "Host": "", "RequestURI": "", "Method": "POST", "query": ""}
[...]

PR #581, opened 2020-06-03
[GH-549] fix transition slack attachment transition

jfrerich:
#### Summary
In the case that a `transition name` does not match the `to state` in the slack attachment options picker, the user would receive an error such as below:

Note that `Start Progress` in this case is the `transition name` and the list `In Progress, Solved, Close` are the states.

`2020-05-26T13:45:03.351+0200    error   mlog/sugar.go:23        ERROR:  {"plugin_id": "jira", "Status": "500", "Error": "\"Start Progress\" is not a valid state. Please use one of: \"In Progress, Solved, Closed\"", "Host": "", "RequestURI": "", "Method": "POST", "query": ""}`

To fix this, the options picker value needs to be the state name, not the transition name.  The current modification will show the transition name in the picker, but the underlying value will be the `to state`

#### Testing
I modified my jira testing project. [Jira link](https://mmtest.atlassian.net/secure/RapidBoard.jspa?rapidView=22&projectKey=JTP)

In that instance I made the following workflow change (note the subtle spelling change)

From:
  transition name: `Backlog`
  to state name: `Backlog`

To:
  transition name: `Backlog`
  to state name: `Bcaklog`  <-- differs

### Screenshots
Note the label value in the picker still differs form the state value. Is there any reason we want to show the transition names here and not the state names?

image

image

#### Ticket Link
Fixes: #549

### DIFF
--- a/server/issue_parser.go
+++ b/server/issue_parser.go
@@ -58,7 +58,7 @@ func getTransitionActions(client Client, issue *jira.Issue) ([]*model.PostAction
 		if transition.Name != issue.Fields.Status.Name {
 			options = append(options, &model.PostActionOptions{
 				Text:  transition.Name,
-				Value: transition.Name,
+				Value: transition.To.Name,
 			})
 		}
 	}


jfrerich:
@DHaussermann ping for review

PR #623, opened 2020-08-13
[GH-612] Add ability to change telemetry key to production key on release builds

### LINKED ISSUE #612: Change telemetry key to production key on release builds
On v3.0-beta we are still using the development key for telemetry instead of the production key.

This has to change before going into production.

This ticket was originally created by @larkox and transferred from the EE repository.

jfrerich:
#### Summary
Allow setting rudderWriteKey during build time.

I used the following as references:
https://github.com/mattermost/mattermost-plugin-zoom/pull/148
https://github.com/mattermost/mattermost-plugin-todo/pull/101
https://github.com/mattermost/mattermost-plugin-mscalendar/pull/152

#### Ticket Link
fixes #612

### DIFF
--- a/build/custom.mk
+++ b/build/custom.mk
@@ -1,0 +1,5 @@
+# Include custom targets and environment variables here
+ifndef MM_RUDDER_WRITE_KEY
+	MM_RUDDER_WRITE_KEY = 1d5bMvdrfWClLxgK1FvV3s4U1tg
+endif
+GO_BUILD_FLAGS += -ldflags '-X "github.com/mattermost/mattermost-plugin-jira/server/utils/telemetry.rudderWriteKey=$(MM_RUDDER_WRITE_KEY)"'
--- a/build/custom.mk
+++ b/build/custom.mk
@@ -2,4 +2,4 @@
 ifndef MM_RUDDER_WRITE_KEY
 	MM_RUDDER_WRITE_KEY = 1d5bMvdrfWClLxgK1FvV3s4U1tg
 endif
-GO_BUILD_FLAGS += -ldflags '-X "github.com/mattermost/mattermost-plugin-jira/server/utils/telemetry.rudderWriteKey=$(MM_RUDDER_WRITE_KEY)"'
+LDFLAGS += -X "github.com/mattermost/mattermost-plugin-jira/server/utils/telemetry.rudderWriteKey=$(MM_RUDDER_WRITE_KEY)"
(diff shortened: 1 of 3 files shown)


build/custom.mk
```diff
@@ -0,0 +1,5 @@
+# Include custom targets and environment variables here
+ifndef MM_RUDDER_WRITE_KEY
+	MM_RUDDER_WRITE_KEY = 1d5bMvdrfWClLxgK1FvV3s4U1tg
+endif
+GO_BUILD_FLAGS += -ldflags '-X "github.com/mattermost/mattermost-plugin-jira/server/utils/telemetry.rudderWriteKey=$(MM_RUDDER_WRITE_KEY)"'
```
mickmister, line 5: The Makefile in this repo is not importing this `custom.mk` file. You'll need to import it like other repos are using it.

larkox: @mickmister is right. This is at the moment not working. Probably the reason he stated.
One way to check it is to see the `make build` commands used. Checking the CI, I can see the following:

`/usr/local/go/bin/go build  -ldflags '-X "main.BuildDate=Thu Aug 13 15:45:30 UTC 2020" -X "main.BuildHash=7003f3fe9d814767f77a83f58b9f36101c877dec" -X "main.BuildHashShort=7003f3f"' -o dist/plugin-darwin-amd64;`

There you should have also the other ldflag in place.

On top of that, if the environment variable is set on CircleCI, you will see the environment variable censored, instead of raw. If you see the development variable raw on CircleCI it means that you have not set the variable correctly on CircleCI.

As a side note, setting the variable in CircleCI is not by itself part of this PR (it is done directly on circleCI and it is not linked to any specific PR) but it would be great to have it done now, so we don't forget.

jfrerich: Thanks, fellas. I added the import of `custom.mk` to the makefile.  I tested locally to ensure the variable was accessible in the telemetry package and defaulted to the dev value is `MM_RUDDER_WRITE_KEY` was not set.  I also tested the value matched `MM_RUDDER_WRITE_KEY` if it was set.

jfrerich: I also added the Environment variable to CircleCI. Here is the redacted result.
image

mickmister:
LGTM :+1:

larkox:
@DHaussermann There is not much that can be tested for this change, I think, apart of what Jason already did (check that the variable is updated when the environment variable is set) and that the events are not broken (just checking that one event arrives fine is more than enough).
The only other thing would be to use the build generated by circleCI to check if the events reach the production telemetry server, but this circleCI does not create the artifacts, so nothing to do on that side.

DHaussermann:
As per @larkox it is the case that no artifacts get created from the circleCI build. Therefore I have no way to see the parameterized Rudder key is being replaced here. 

@jfrerich there is no functional testing applicable here unfortunately. 
LGTM!

Issue #663, opened 2020-09-24
Cannot view saved details of a subscription if it's project not currently set an the users current default project


DHaussermann:
After a subscription has been saved,  it will not reopen properly if it corresponds to a project that is not currently set as the default project for the user. 

There is logic that selects the default project first when a subscription is created.  This is affecting the edit functionality.  When the edit is open,  subscriptions saved for a different project will have the drop down updated to match the current project and the details cannot be loaded. 

Steps:
- Install and and connect to an instance with 2 project you have permission to
- Open the Jira create modal and not which project is selected by default as your current project ex: ProjectA
- Create a subscription that targets ProjectA
- Note the subscription can be edited without issue
- Use the the create or attach modal to add an issue or comment to ProjectB
- This updates you default project to ProjectB
- Attempt to edit the subscription created above
**Observed**: When the edit view opens ProjectB is loaded into the drop down and the subscription can not be edited  
![SubProject](https://user-images.githubusercontent.com/35588783/94148083-2af2ae00-fe44-11ea-95f6-a12aa460c4ca.gif)

PR #666, opened 2020-09-25
[GH-663] Saved subscription project value incorrectly updated to default_project_key value

jfrerich:
#### Summary
Thsi ticket resolves the case where a saved subscription is edited.  Previously, the `default_project_key` was getting set when the edit subscription modal would open. This resulted in opening a saved subscription with a different project that the saved default project would get changed to the default project key.  

The resolution is to check if the selectedProjectID exists, which is the saved subscription project ID.

#### Ticket Link
#663

### DIFF
--- a/webapp/src/components/jira_instance_and_project_selector/jira_instance_and_project_selector.tsx
+++ b/webapp/src/components/jira_instance_and_project_selector/jira_instance_and_project_selector.tsx
@@ -95,7 +95,7 @@ export default class JiraInstanceAndProjectSelector extends React.PureComponent<
             fetchingProjectMetadata: false,
         });
 
-        if (projectMetadata.default_project_key) {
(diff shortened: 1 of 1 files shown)


larkox (approved):

levb (approved):

DHaussermann (approved):
Tested and passed - This issue is resolved
- Default project is selected when adding new subscription
- Correctly saved project is always selected when editing a saved subscription 
- Regression tested Create modal will use and update default project as expected
- Regression tested that running slash commands without passing in instance will always use current default 
LGTM!

Added TM4J test coverage for this bug here https://mattermost.atlassian.net/projects/MM?selectedItem=com.atlassian.plugins.atlassian-connect-plugin%3Acom.kanoah.test-manager__main-project-page#!/testCase/MM-T3382

PR #680, opened 2020-10-01
[GH-678] when using --instance, show alias if available else show the instanceID

### LINKED ISSUE #678: /Jira issue subcommand are only showing instances with an alias set
When using `/jira issue` subcommand, the auto complete to pass in an instance at the end will only show instances that have an alias set. 
For ex: `/jira issue view dkh-123 --instance` after you select `--instance`, all connected instances should be shown. Instead it will only show instances where an alias has been saved.

- The command still work if you type out the URL
- All Issue type commands such as assign, and transition are affected
- The same thing happens if you attempt to use a Jira V2 style command ex: `/jira view ext-122 --instance` 
- The command will still use default instance as expected if none is passed

jfrerich:
#### Summary
When showing the list of available `--instance` options, show the aliases if available, else show the `instanceID`.

I tested this with 
`/jira issue view JTP-1`
`/jira issue view jstp-1 --instance server-alias`
`/jira issue view jtp-1 --instance https://mmtest.atlassian.net`

`/jira assign jstp-1 jfrerich --instance server-alias `
/jira assign jtp-1 jfrerich --instance https://mmtest.atlassian.net 

#### Ticket Link
Fixes #678

### DIFF
--- a/server/instances.go
+++ b/server/instances.go
@@ -402,16 +402,13 @@ func (p *Plugin) httpAutocompleteInstalledInstanceWithAlias(w http.ResponseWrite
 
 	instances, err := p.instanceStore.LoadInstances()
 	for _, instanceID := range info.Instances.IDs() {
-		alias := instances.getAlias(instanceID)
-		if alias != "" {
-			out = append(out, model.AutocompleteListItem{
-				Item: alias,
-			})
-		} else {
-			out = append(out, model.AutocompleteListItem{
-				Item: string(instanceID),
-			})
+		item := instances.getAlias(instanceID)
+		if item != "" {
+			item = string(instanceID)
 		}
+		out = append(out, model.AutocompleteListItem{
+			Item: item,
+		})
 	}
 	return respondJSON(w, out)
 }
--- a/server/instances.go
+++ b/server/instances.go
@@ -407,6 +407,10 @@ func (p *Plugin) httpAutocompleteInstalledInstanceWithAlias(w http.ResponseWrite
 			out = append(out, model.AutocompleteListItem{
 				Item: alias,
 			})
+		} else {
+			out = append(out, model.AutocompleteListItem{
+				Item: string(instanceID),
+			})
 		}
 	}
 	return respondJSON(w, out)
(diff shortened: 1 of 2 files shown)


jfrerich:
@DHaussermann, I did find another issue.  Could you confirm that you see this also?  

When a user installs an instance but has not opened the `create issue` modal, the default instance does not get set. 

This means that all the `/jira` commands that need the default instance do not work.

mickmister (approved):

mickmister:
> This means that all the `/jira` commands that need the default instance do not work.

@jfrerich Could we default to v2legacy or the first one we find that they are connected to?

server/instances.go
```diff
@@ -407,6 +407,10 @@ func (p *Plugin) httpAutocompleteInstalledInstanceWithAlias(w http.ResponseWrite
 			out = append(out, model.AutocompleteListItem{
 				Item: alias,
 			})
+		} else {
```
levb, line 410: suggest:
```go
item := instances.GetAlias(instanceID)
if item == "" {
  item = string(instanceID)
}
out = append(out, model.AutocompleteListItem{
				Item: item,
			})

levb (changes requested):

server/instances.go
jfrerich: yes. much clearer

DHaussermann:
@jfrerich there's a bug here that mat be a recent regression. Updating an existing alias now breaks the list of servers 
I'm running build `dd37472 ` (2nd commit)
note: Create and subscribe are not affected.

Steps:
- Set an alias
- Note that alias a non alias instances show fine after `--instance`
- Use the slash command to set a different alias on the same instance that already has one to update it
**Observed** 
A. Instance without Alias no longer shown
B. Instances with Aliases now show the URL not the saved alias 
Screen Shot 2020-10-01 at 3 30 44 PM

levb (approved):

jfrerich:
@DHaussermann looking into it now

jfrerich:
@DHaussermann please take another look.

Note that I also corrected the following commands to list an alias in the autocomplete (if available) and also accept the alias or instanceID as a value.

* /jira subscribe list --instance
* /jira instance alias
* /jira instance unalias
* /jira webhook

DHaussermann (approved):
Tested and passed
- Tested CRUD operations around alias for instances
- Tested display is correct with and without alias in slash command autocomplete with expected command 
- Tested display is correct with and without alias in create and subscribe modal
- Ensured UI updates correctly when aliases have changed
 LGTM! @jfrerich no issues found. 

Please advise if another round of dev review is needed before we merge this.

mickmister (approved):

Issue #683, opened 2020-10-02
Command responses don't get posted in the RHS


hanzei:
Same bug as https://github.com/mattermost/mattermost-plugin-github/issues/364

PR #684, opened 2020-10-02
Post command response in RHS if triggered from there

hanzei:
#### Summary
Same bug as for GH

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-jira/issues/683

### DIFF
--- a/server/command.go
+++ b/server/command.go
@@ -616,6 +616,7 @@ func executeView(p *Plugin, c *plugin.Context, header *model.CommandArgs, args .
 	post := &model.Post{
 		UserId:    p.getUserID(),
 		ChannelId: header.ChannelId,
+		RootId:    header.RootId,
 	}
 	post.AddProp("attachments", attachment)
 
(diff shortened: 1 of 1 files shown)


mickmister (approved):

jfrerich (approved):

DHaussermann:
@hanzei I believe this is working as expected. For the most part all responses are posted in the RHS when the command is sent from there. 

I notice that this is not the case for Jira `view` which creates a post with a visual representation of the issue. I'm assuming this is the expected behavior. I don't know what such a post would look like in the RHS where there is limited horizontal space for the elements. 

Can you confirm this change does not apply yo Jira `view`
Screen Shot 2020-10-09 at 9 41 07 AM

hanzei:
@DHaussermann Actually, it should post in the RHS. I've fixed that. PTAL

hanzei:
@DHaussermann I'm fairly confident that these two issues aren't related to this PR. Please open a new ticket for them.

DHaussermann (approved):
Tested and passed
- Tested all slash commands in RHS 
- Reply is also show in the RHS as expected
- Issue found with `/jira view has now been addressed`
- Created https://github.com/mattermost/mattermost-plugin-jira/issues/698 to investigate and resolve JS errors found 

LGTM!

PR #835, opened 2022-02-10
[GH - 774] Add tooltip for jira ticket links

### LINKED ISSUE #774: Add tooltip for jira ticket links
Implement something similar to https://github.com/mattermost/mattermost-plugin-github/issues/255

It should show:
- Ticket name
- Status
- Type
- First few characters from the description

sibasankarnayak:
implemented tooltip to show  Ticket name, Status, Type, First few characters from the description

ticket here
Fixes #774

### DIFF
--- a/server/http.go
+++ b/server/http.go
@@ -55,7 +55,7 @@ const (
 	routeUserConnect                            = "/user/connect"
 	routeUserDisconnect                         = "/user/disconnect"
 	routeSharePublicly                          = "/api/v2/share-issue-publicly"
-	routeGetIssueByKey                          = "/api/v2/get-isse-by-key"
+	routeGetIssueByKey                          = "/api/v2/get-issue-by-key"
 )
 
 const routePrefixInstance = "instance"
--- a/server/http.go
+++ b/server/http.go
@@ -55,6 +55,7 @@ const (
 	routeUserConnect                            = "/user/connect"
 	routeUserDisconnect                         = "/user/disconnect"
 	routeSharePublicly                          = "/api/v2/share-issue-publicly"
+	routeGetIssueByKey                          = "/api/v2/get-isse-by-key"
 )
 
 const routePrefixInstance = "instance"
@@ -126,6 +127,8 @@ func (p *Plugin) serveHTTP(c *plugin.Context, w http.ResponseWriter, r *http.Req
 		return p.httpTransitionIssuePostAction(w, r)
 	case routeSharePublicly:
 		return p.httpShareIssuePublicly(w, r)
+	case routeGetIssueByKey:
+		return p.httpGetIssueByKey(w, r)
 
 	// User APIs
 	case routeAPIUserInfo:
--- a/webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
+++ b/webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
@@ -5,23 +5,13 @@ import {Instance, GetConnectedResponse} from 'types/model';
 
 export type Props = {
     href: string;
+    show:boolean;
     connected: boolean;
     isloaded?: boolean;
-    assigneeName?: string;
-    assigneeAvatar?: string;
-    labels?: any[];
-    versions?: string;
-    description?: string;
-    summary?: string;
-    ticketId?: string;
-    jiraIcon?: string;
-    statusKey?: string;
-    issueIcon?: string;
+    ticketDetails:any;
     defaultUserInstanceID?: string;
-    installedInstances: Instance[];
     connectedInstances: Instance[];
-    getIssueByKey: (ticketId: string, instanceID: string) => void;
-    getConnected: () => Promise<GetConnectedResponse>;
+    getIssueByKey: (ticketId: string, instanceID: string) => Promise<{data: any; error?: Error}>;
     setTicket?: (ticketDetails: {}) => void;
 }
 export default class TicketPopover extends React.PureComponent<Props> {
@@ -68,43 +58,52 @@ export default class TicketPopover extends React.PureComponent<Props> {
         }
 
         let ticketID = '';
+        let ticketData = null ;
         if (this.props.href.includes('selectedIssue')) {
             ticketID = this.props.href.split('selectedIssue=')[1].split('&')[0];
-            this.props.getIssueByKey(ticketID, instanceID);
+            this.props.getIssueByKey(ticketID, instanceID).then(({data,error})=>{
+                if (error){
+                    return
+                }
+
+                this.setTicket(data)
+            });
         }
 
         if (ticketID === '' && this.props.href.includes('atlassian.net/browse')) {
-            this.props.getIssueByKey(this.props.href.split('|')[0].split('?')[0].split('/browse/')[1], instanceID);
+            this.props.getIssueByKey(this.props.href.split('|')[0].split('?')[0].split('/browse/')[1], instanceID).then(({data,error})=>{
+                if (error){
+                    return
+                }
+
+                this.setTicket(data)
+            });
         }
     }
 
-    componentDidMount() {
-        this.props.getConnected();
-        if (this.props.connected && !this.state.isloaded) {
-            this.init();
+    componentDidUpdate(prevProps:Props): void {
+        if (!this.props.connected){
+            return
         }
+
+        if (!(this.props.show && !prevProps.show)) {
+            return;
+        }
+
+        if (this.state.isLoaded) {
+            return;
+        }
+    
+        this.init();
     }
 
-    componentDidUpdate(): void {
-        if (this.props.isloaded && !this.state.isloaded && this.props.ticketId === this.state.ticketId) {
-            this.setTicket(this.props);
-        }
-    }
-
-    setTicket(ticketDetails: {}): void{
+    setTicket(ticketDetails:any): void{
         this.setState({
             isloaded: true,
-            assigneeName: this.props.assigneeName,
-            assigneeAvatar: this.props.assigneeAvatar,
-            labels: this.props.labels,
-            versions: this.props.versions,
-            description: this.props.description,
-            summary: this.props.summary,
-            jiraIcon: this.props.jiraIcon,
-            statusKey: this.props.statusKey,
-            issueIcon: this.props.issueIcon,
+            ticketDetails:ticketDetails,
         });
     }
+
     fixVersionLabel(fixVersion: string) {
         if (fixVersion) {
             const fixVersionString = 'Fix Version :';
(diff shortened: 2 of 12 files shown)


sibasankarnayak:
@hanzei @mickmister 
this is the screenshot for tooltip , please share your feedback

image

sibasankarnayak:
@mickmister can you review this PR.

hanzei:
@Rina-dsg @mickmister @levb Gentle reminder to review this PR

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+    constructor(Props) {
+        super(Props);
+        let ticketID = '';
+        if (this.props.href.includes('selectedIssue')) {
+            ticketID = this.props.href.split('selectedIssue=')[1].split('&')[0];
+        }
+
+        if (ticketID === '' && this.props.href.includes('atlassian.net/browse')) {
```
levb, line 42: will this trigger on autolink-processed links?

levb (approved):

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
sibasankarnayak: yes it does work after autolink being processed , can find a recording of same
Screen Recording 2022-03-02 at 9.39.18 PM

server/http.go
```diff
@@ -55,6 +55,7 @@ const (
 	routeUserConnect                            = "/user/connect"
 	routeUserDisconnect                         = "/user/disconnect"
 	routeSharePublicly                          = "/api/v2/share-issue-publicly"
+	routeGetIssueByKey                          = "/api/v2/get-isse-by-key"
```
catalintomai, line 58: isse->issue

webapp/src/reducers/index.js
```diff
@@ -193,6 +193,33 @@ const channelSubscriptions = (state = {}, action) => {
     }
 };
 
+const getIssueByKey = (state = {}, action) => {
+    switch (action.type) {
+    case ActionTypes.RECEIVED_JIRA_TICKETS :
+        return {
+            ...state,
+            isloaded: true,
+            assigneeName: action.data.fields.assignee && action.data.fields.assignee.displayName ? action.data.fields.assignee.displayName : '',
```
catalintomai, line 202: might make sense to save `action.data.fields.assignee` in a var and reuse it here to improve readability.

sibasankarnayak: sure

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,305 @@
...
+            ticketID = this.props.href.split('|')[0].split('?')[0].split('/browse/')[1];
+        }
+
+        this.state = {
+            href: this.props.href,
```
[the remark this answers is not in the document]
sibasankarnayak: @catalintomai the tooltip will only trigger when it found a link so href will always be present

hanzei:
@Rina-dsg Could you please review this PR?

hanzei:
@matthewbirtch Would you mind reviewing this PR while @Rina-dsg is out?

matthewbirtch:
> @matthewbirtch Would you mind reviewing this PR while @Rina-dsg is out?

@hanzei I'm happy to review. I just triggered a cloud server build, but will I need to do anything to set it up properly to review this PR?

sibasankarnayak:
> Unfortunately, the test server created via labels can't be used for plugin testing. Please use https://jira-tooltip.test.mattermost.cloud/ instead.
> 
> @sibasankarnayak is that server ready?

I wasn't able to upload the plugin  as it was throwing  error as it was created with linux  system

hanzei:
@sibasankarnayak Can you please go into more detail? Can be bundles at https://app.circleci.com/pipelines/github/mattermost/mattermost-plugin-jira/2975/workflows/3e675fbf-f7b7-4a09-b6dd-4c18731d3bd5/jobs/9400/artifacts  be used?

hanzei:
@sibasankarnayak I've updated the server. Could you please try again?

sibasankarnayak:
> @sibasankarnayak I've updated the server. Could you please try again?

@hanzei have deployed , configured the tooltip feature ,have some sample [here](https://jira-tooltip.test.mattermost.cloud/core/channels/town-square) 
please let me know the feedback

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+    tagTicketStatus(ticketStatus: string) {
+        const defaultStyle = {
+            fontFamily: 'open sans',
+            fontStyle: 'normal',
+            fontWeight: 600,
+            fontSize: '12px',
+            marginTop: '10px',
+            padding: '0px 8px',
```
matthewbirtch, line 136: change `padding: '4px 8px'`

matthewbirtch (changes requested):
I've added a number of changes in the CSS- hopefully that makes it clear.

Also, a couple other things I noticed:

1. There is no arrow pointing to the link as per the design. Is this something we can add? Here's a sample of the intended design:

image

2. The description should be truncated to 3 lines of text

sibasankarnayak:
> I've added a number of changes in the CSS- hopefully that makes it clear.
> 
> Also, a couple other things I noticed:
> 
> 1. There is no arrow pointing to the link as per the design. Is this something we can add? Here's a sample of the intended design:
> 
> image
> 
> 2. The description should be truncated to 3 lines of text

@matthewbirtch sure , Thankyou for feedback.

regarding the 1 st point showing an arrow is bit complex not sure if we can as the tooltip can open from any direction to the link , if you have any solution can share it with me will give a try. For rest of the comment i will make changes

matthewbirtch:
> regarding the 1 st point showing an arrow is bit complex not sure if we can as the tooltip can open from any direction to the link , if you have any solution can share it with me will give a try. For rest of the comment i will make changes

@sibasankarnayak we have a tour point component that utilizes an arrow like this. @AshishDhama could you help provide a possible approach for a similar component in this PR?

If it's too complicated, I think we can proceed without as this still brings value without it, but it would be nice to have it.

sibasankarnayak:
> tour point component that utilizes an arrow like this

Sure @matthewbirtch if you can share i ll sure try it up.
creating a triangle could be possible with `:before` and `:after` but not sure how to create it dynamically on position.

AshishDhama:
@sibasankarnayak @matthewbirtch, we are using an external library for that particular reason called Tippy. The tour tip component, which we have used, can be found here https://github.com/mattermost/mattermost-webapp/tree/master/components/widgets/tour_tip. @sibasankarnayak you can use the same if that can resolve the issue, but please make sure of using the same version. Please let me know if you need any help.

mickmister:
> you can use the same if that can resolve the issue, but please make sure of using the same version

@AshishDhama Just want to note that if we need to use the same version of a node module, we would want to first export this to plugins in https://github.com/mattermost/mattermost-webapp/blob/master/plugins/export.js, so we ensure the same version will _always_ be used.

@sibasankarnayak Can you try implementing this by exposing the library via the webapp's exports? We wouldn't get type completion in the plugin's source code without also installing the package in the Jira plugin's project, but the package will be available to use at runtime via the `window` object.

AshishDhama:
@mickmister Thanks will add it there.

sibasankarnayak:
adasd

sibasankarnayak:
@sibasankarnayak Siba

mickmister:
@sibasankarnayak Have you created the PR to add this to the webapp's exports? As I said, we need that to be done as soon as possible to minimize the change in `min_server_version` in the plugin

sibasankarnayak:
> @sibasankarnayak Have you created the PR to add `Tippy` to mattermost-webapp's exports? As I said, we need that to be done as soon as possible to minimize the change in `min_server_version` in the plugin

created the PR https://github.com/mattermost/mattermost-webapp/pull/10235

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,317 @@
...
+        const Tippy = window.Tippy.default;
+
+        const Tooltip = ({children, ...rest}) => <Tippy {...rest}>{children}</Tippy>;
+        Tooltip.defaultProps = {
+            animation: 'fade',
+            arrow: true,
+        };
+
+        return (
+            <Tippy arrow={true} >
```
AshishDhama, line 216: @sibasankarnayak Tippy needs a reference to a point, so create a ref of the hyperlink and provide that ref to tippy. it should work.
https://github.com/mattermost/mattermost-webapp/blob/c5b87db2305a8fa4bbf8b5095ecf44d3158a1c53/components/widgets/tour_tip/tour_tip.tsx#L188
https://github.com/mattermost/mattermost-webapp/blob/c5b87db2305a8fa4bbf8b5095ecf44d3158a1c53/components/widgets/tour_tip/tour_tip.tsx#L217

sibasankarnayak: @AshishDhama for providing reference i need to give a reference to hyperlink but as this is plugin we register this tooltip component on hover , i am not sure if we can add something like as you mentioned in this case.

let me know if we can or you can suggest the changes in PR that will help a lot.

raghavaggarwal2308:
@mickmister @AshishDhama For using Tippy or any other tooltip library we need a reference to the actual component which will act as an anchor to the tooltip and in our case the component which needs to be an anchor is present on the Mattermost webapp code so it would not be possible to refer that component directly from the plugin.
However, this can be achieved using some UI hacks but we don't think that would be feasible for rendering the arrow.
However, we have everything similar to the Figma except that arrow which is perfectly fine.
Even the tooltip present in the GitHub plugin doesn't contain the arrow as well and uses Mattermost's default tooltip component.
So, we suggest skipping the arrow and just fixing the typescript errors in this PR, please let us know your thoughts if it's different  

Screenshot from 2022-10-25 22-00-10

mickmister:
> For using Tippy or any other tooltip library we need a reference to the actual component 

@raghavaggarwal2308 There's an open PR that exposes the component to plugins to use as `window.Tippy`
https://github.com/mattermost/mattermost-webapp/pull/10235

> So, we suggest skipping the arrow and just fixing the typescript errors in this PR, please let us know your thoughts if it's different

I'm fine with skipping the arrow if necessary

raghavaggarwal2308:
@mickmister Since the PR [#10235](https://github.com/mattermost/mattermost-webapp/pull/10235) has not merged yet. We suggest merging this PR with the typescript errors fixed, without the arrow and we can revisit that change once PR [#10235](https://github.com/mattermost/mattermost-webapp/pull/10235) is merged.

Created issue #886 for the same.

PR #848, opened 2022-03-08
Create issue modal crashes when assignee is selected before switching projects

maisnamrajusingh:
#### Summary
 Fixes the issue whenever user selects a project , task and then an asignee and then switches projects -> tasks again which causes the app to crash 
#### Ticket Link
https://github.com/mattermost/mattermost-plugin-jira/issues/819

### DIFF
--- a/webapp/src/components/modals/create_issue/create_issue_form.tsx
+++ b/webapp/src/components/modals/create_issue/create_issue_form.tsx
@@ -134,9 +134,10 @@ export default class CreateIssueForm extends React.PureComponent<Props, State> {
 
             this.setState(state);
         });
-
+        
         let fields = {
-            ...this.state.fields,
+            summary: this.state.fields.summary,
+            description: this.state.fields.description,
             project: {key: projectKey},
         } as CreateIssueFields;
 
@@ -223,7 +224,7 @@ export default class CreateIssueForm extends React.PureComponent<Props, State> {
 
         const result = {} as CreateIssueFields;
         for (const key of Object.keys(fields)) {
-            if ((available[key] || key === 'project' || key === 'issuetype') && key !== 'assignee') {
+            if (available[key] || key === 'project' || key === 'issuetype') {
                 result[key] = fields[key];
             }
         }
--- a/webapp/src/components/modals/create_issue/create_issue_form.tsx
+++ b/webapp/src/components/modals/create_issue/create_issue_form.tsx
@@ -220,10 +220,10 @@ export default class CreateIssueForm extends React.PureComponent<Props, State> {
         if (Object.keys(available).length === 0) {
             return fields;
         }
-
+        
         const result = {} as CreateIssueFields;
         for (const key of Object.keys(fields)) {
-            if (available[key] || key === 'project' || key === 'issuetype') {
+            if ((available[key] || key === 'project' || key === 'issuetype') && key !== 'assignee') {
                 result[key] = fields[key];
             }
         }
--- a/webapp/src/components/modals/create_issue/create_issue_form.tsx
+++ b/webapp/src/components/modals/create_issue/create_issue_form.tsx
@@ -98,9 +98,6 @@ export default class CreateIssueForm extends React.PureComponent<Props, State> {
                 },
                 issuetype: {
                     id: '',
-                },
-                assignee: {
-                    accountId: '',
                 }
             } as CreateIssueFields,
         };
--- a/webapp/src/components/modals/create_issue/create_issue_form.tsx
+++ b/webapp/src/components/modals/create_issue/create_issue_form.tsx
@@ -98,7 +98,7 @@ export default class CreateIssueForm extends React.PureComponent<Props, State> {
                 },
                 issuetype: {
                     id: '',
-                }
+                },
             } as CreateIssueFields,
         };
     }
@@ -220,7 +220,7 @@ export default class CreateIssueForm extends React.PureComponent<Props, State> {
         if (Object.keys(available).length === 0) {
             return fields;
         }
-        
+
         const result = {} as CreateIssueFields;
         for (const key of Object.keys(fields)) {
             if ((available[key] || key === 'project' || key === 'issuetype') && key !== 'assignee') {
--- a/webapp/src/components/modals/create_issue/create_issue_form.tsx
+++ b/webapp/src/components/modals/create_issue/create_issue_form.tsx
@@ -99,6 +99,9 @@ export default class CreateIssueForm extends React.PureComponent<Props, State> {
                 issuetype: {
                     id: '',
                 },
+                assignee: {
+                    accountId: '',
+                }
             } as CreateIssueFields,
         };
     }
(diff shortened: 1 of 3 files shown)


maisnamrajusingh:
@mickmister  please ignore this until @dipak-demansol  has done the qa on this.

webapp/src/components/modals/create_issue/create_issue_form.tsx
```diff
@@ -223,7 +223,7 @@ export default class CreateIssueForm extends React.PureComponent<Props, State> {
 
         const result = {} as CreateIssueFields;
         for (const key of Object.keys(fields)) {
-            if (available[key] || key === 'project' || key === 'issuetype') {
+            if ((available[key] || key === 'project' || key === 'issuetype') && key !== 'assignee') {
```
maisnamrajusingh, line 226: @mickmister tested this a couple of times and this fixes the issue but for me this is more of a hot fix as the underlying issue is that the types don't take into account the assignee and the user selector component which seems to have conflicts with the type as well. Maybe we can create a seperate issue that fixes it.

mickmister: There are other custom user selector fields besides `assignee`, so this fix won't fix it for all cases of having a user selector.

Something to keep in mind is that different projects and different issue types can have entirely different field schemas.

There is a block in `create_issue_form.tsx` that copies all the previous fields to the new form state:

https://github.com/mattermost/mattermost-plugin-jira/blob/538f68a79144890463476474b10b33893bfa41df/webapp/src/components/modals/create_issue/create_issue_form.tsx#L138-L141

I think we should make it so the `summary` and `description` are the only fields that we keep around, in the `handleProjectChange` and `handleIssueTypeChange` functions.

```js
let fields = {
    project: {key: projectKey},
    summary: this.state.fields.summary,
    description: this.state.fields.description,
} as CreateIssueFields;
```

@maisnamrajusingh Please let me know if this makes sense and if it fixes the issue

mickmister (commented):
Please see the comment above, regarding clearing the existing fields when a project or issue type is changed

dipak-demansol:
its LGTM but i'll test it again on the browser & desktop app when the dev review is done.

webapp/src/components/modals/create_issue/create_issue_form.tsx
maisnamraju: @mickmister I have updated the logic and it's working fine. However the user selector does seem to have a problem of it's own.  I know that it's because of the types not matching in the user selector but I guess as you have mentioned, that this will have effect on other options. I'll add it as an issue that maybe we can work on in the future. 
Screenshot 2022-03-16 at 10 36 25 PM

maisnamrajusingh: @mickmister I have updated the logic and it's working fine. However the user selector does seem to have a problem of it's own.  I know that it's because of the types not matching in the user selector but I guess as you have mentioned, that this will have effect on other options. I'll add it as an issue that maybe we can work on in the future. Screenshot 2022-03-16 at 10 36 25 PM

maisnamrajusingh: @mickmister this seems to be an existing issue that already existed and is not related to the modal crash. It's related to this line here 
https://github.com/mattermost/mattermost-plugin-jira/blob/f986c159ec37a6b1cf727840d8899f84a1aff612/webapp/src/components/data_selectors/jira_user_selector/jira_user_selector.tsx#L51

the return type should be 
```
  return {
                    value: user.displayName,
                    label,
                };
                ```
since the return type of the function is 
```
ReactSelectOption
```

Comment (mickmister) [reply webapp/src/components/modals/create_issue/create_issue_form.tsx]: > however this doesn't have any effect on the way it works though.

Do you mean the feature still works correctly when we satisfy the prop type? Is it an error that `value: user` is being used, or does making it a string `user.displayName` break other logic?

It seems we need to integrate typescript into CI to catch these sort of errors.

Comment (mickmister) [reply webapp/src/components/modals/create_issue/create_issue_form.tsx]: Ticket to add typescript to CI https://github.com/mattermost/mattermost-plugin-jira/issues/852

Comment (maisnamrajusingh) [reply webapp/src/components/modals/create_issue/create_issue_form.tsx]: Yeah the feature still works correctly after we satisfy user.displayName. I am worried that the component is used elsewhere and might break other components in case it's used eslewhere

**Review by dipak-demansol — APPROVED** [review]
LGTM, tested on subpath server.

**Review by dipak-demansol — APPROVED** [review]
LGTM, 
:- The Original Crash issue is Fixed, tested with single and Multiple instance.  
:- I'm able to Create issue + Tested Without inserting data on mandatory fields and user is getting the validations.

**Comment by maisnamrajusingh** [comment]
@mickmister could you take a look ? I feel this is good to go.

File: webapp/src/components/modals/create_issue/create_issue_form.tsx [hunk 2]
```diff
@@ -136,7 +136,8 @@ export default class CreateIssueForm extends React.PureComponent<Props, State> {
         });
 
         let fields = {
-            ...this.state.fields,
+            summary: this.state.fields.summary,
+            description: this.state.fields.description,
```
mickmister, line 140: Changes LGTM, though there is one other case I want to test @dipak-demansol. We have solved this for switching projects, but not for switching issue types. A similar change could be done on line 164 to fix this.

@maisnamrajusingh I would go ahead and apply the same change there to avoid a crash in that scenario too. This would happen if a given issue type schema has a field that another issue type in the same project does not have. Even though that is probably uncommon, I'm thinking we should be ssafe there.

The drawback is that the user would lose some progress on filling th eissue out when they change issue types, but this way is much more correct and solves other potential issues as well.

mickmister (changes requested):

webapp/src/components/modals/create_issue/create_issue_form.tsx
mickmister: @dipak-demansol I don't think it's worth the trouble to set up two issue schemas in the same Jira project that have the characteristics I brought up above. No need to try to have a reproducible case for that scenario.

javaguirre: Hi! Can we solve this conversation? I think there nothing else to be fixed here right? Let me know if I can help in any way here. @mickmister

mickmister: > We have solved this for switching projects, but not for switching issue types. A similar change could be done on line 164 to fix this.

@javaguirre I think this request is still pending a change here

javaguirre: are you working on this @maisnamrajusingh ?

mickmister: @javaguirre This PR was submitted by a previous contractor, so it's sort of in a frozen state to be used as a starting point for someone else to pick up the task. I've marked the ticket https://github.com/mattermost/mattermost-plugin-jira/issues/819 as HW and linked to this PR for reference. I'll close this since the source branch won't be used. Thanks for circling back to this PR :+1:

mickmister:
Closing due to https://github.com/mattermost/mattermost-plugin-jira/pull/848#discussion_r995978868

Issue #558, opened 2020-05-28
`/jira assign` should accept `@username`


levb:
Currently, the first parameter is interpreted solely as a search string for a Jira user search. We should support accepting an `@username`. If the user is connected to the instance, we can use the AccountID; if not - can perform a Jira search on the Mattermost display name.

hanzei:
https://github.com/mattermost/mattermost-plugin-jira/pull/863 can be used as a starting point

wiersgallak:
Closing issues due to inactivity. This issue can be re-opened with more interest from our community.

wiersgallak:
I see activity on a recent PR. Reopening.

PR #863, opened 2022-07-31
Issue #558

javaguirre:
#### Summary

  Let's you use the JiraUser connected of the user assigned (via mention) instead of looking for the text itself.

#### Ticket Link

  Fixes https://github.com/mattermost/mattermost-plugin-jira/issues/558


javaguirre:
@mickmister Hi! Let me know if this is what was expected! Also if you have any opinions or see any code smells let me know, It's been around 12 years since I don't do any Go code!

Thank you! :-)

javaguirre:
I made a small video testing the functionality


https://user-images.githubusercontent.com/488556/182040226-e1008081-ec2d-42ab-9aaa-44cc1ec8e63b.mov

server/user.go
```diff
@@ -309,3 +309,18 @@ func (p *Plugin) disconnectUser(instance Instance, user *User) (*Connection, err
 
 	return conn, nil
 }
+
+func (p *Plugin) GetJiraUserFromMentions(instanceID types.ID, mentions model.UserMentionMap, userKey string) *jira.User {
+	mentionUser, found := mentions[userKey]
```
mickmister, line 314: If I'm not mistaken, won't `userKey` here contain the `@` symbol, when passed in from `executeAssign`? Maybe we should truncate it here if it exists?

server/issue.go
```diff
@@ -864,7 +864,7 @@ func (p *Plugin) UnassignIssue(instance Instance, mattermostUserID types.ID, iss
 
 const MinUserSearchQueryLength = 3
 
-func (p *Plugin) AssignIssue(instance Instance, mattermostUserID types.ID, issueKey, userSearch string) (string, error) {
+func (p *Plugin) AssignIssue(instance Instance, mattermostUserID types.ID, issueKey, userSearch string, jiraUser *jira.User) (string, error) {
```
mickmister, line 867: For readability

```suggestion
func (p *Plugin) AssignIssue(instance Instance, mattermostUserID types.ID, issueKey, userSearch string, assignee *jira.User) (string, error) {
```

server/user_test.go
```diff
@@ -61,3 +63,47 @@ func TestRouteUserStart(t *testing.T) {
...
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			user := p.GetJiraUserFromMentions(
+				testInstance1.InstanceID, *tc.mentions, tc.userSearch)
+			assert.Nil(t, user)
+		})
+	}
+}
+
+func TestGetJiraUserFromMentionsConnectReturnValidUser(t *testing.T) {
```
mickmister, line 93: The function being tested here is pretty small, and doesn't need to handle many cases. I suggest we collapse the two test functions here into one test function for readability. Instead of `assert.Nil` and `assert.NotNil`, you can assert that the value is equal to a given struct in test setup data. Then it should be easier to have all the tests in one test function.

mickmister: Do you see an opportunity to test the `executeAssign` function? This feature will potentially be used much more with this improvement, so having coverage on that function would be very helpful

mickmister (commented):
Thanks @javaguirre! I have a question about truncating the `@` symbol from the user's command submission, as well as cleaning up some of the test code.

mickmister:
I'm thinking if the user provided a valid Mattermost user mention, but the mentioned user is not connected to Jira, we should return an error to the user rather than falling back to searching in Jira for this user

levb (approved):
I ++ the (minor) change requests from @mickmister, otherwise LGTM.

javaguirre:
I'll make all the changes next week! Thank you! :-)

mickmister:
> > I'm thinking if the user provided a valid Mattermost user mention, but the mentioned user is not connected to Jira, we should return an error to the user rather than falling back to searching in Jira for this user
> 
> Wouldn't that break how it worked before? I don't mind making that change, but I didn't want to break anything or make it backward incompatible @mickmister

Technically yes, but we are changing how the feature works here. If you include at `@` sign, it should be assumed that you're wanting to assign to a MM user. At the moment, this feature really isn't used at all because of it's unpredictability. Adding a guard like this will increase the reliability of the feature.

server/user_test.go
javaguirre: I fixed the first comment, I'll write the tests for executeAssign!

javaguirre:
I've handled when the user mentioned is not found nor connected to Jira, I'll finish the `executeAssign` tests and I think It would be good to go if you're ok with it. :-)

server/user.go
```diff
@@ -311,21 +311,23 @@ func (p *Plugin) disconnectUser(instance Instance, user *User) (*Connection, err
 	return conn, nil
 }
 
-func (p *Plugin) GetJiraUserFromMentions(instanceID types.ID, mentions model.UserMentionMap, userKey string) *jira.User {
+func (p *Plugin) GetJiraUserFromMentions(
+	instanceID types.ID, mentions model.UserMentionMap, userKey string,
+) (*jira.User, error) {
```
mickmister, line 316: Putting this on multiple lines isn't necessarily common practice in our projects. It took me a second to realize what's going on here. What are your thoughts on having it formatted how it was before?

mickmister (approved):
LGTM :+1: Just one comment on formatting

server/user.go
javaguirre: If that's the standard, I'm fine with it. :-)

I try not to pass the 80 character line limit, because then when I split the screen I can't see some code, but it's ok, I'll change it!

Thank you! :-)

javaguirre:
Do I need to do something here @hanzei ? I'm not sure if `Awaiting submitter action` requires something from me. Thank you!

levb:
@javaguirre just remove the label once you finished your task

hanzei:
@DHaussermann Heads up that the PR is ready for your review

DHaussermann:
Tested this and the feature is very neat!
- It works to mention channel members for the issue assignment 
- When mentioning user who are not mapped to Jira, there is clear feedback
image


However there is another point to consider here. This PR does not seem to preserve the existing functionality where the string you pass in was used in a "begins with" type search of users in Jira. It would then assign the user if the string was specific enough to return only one match OR show a list of all potential matches if the string had multiple.
As this no longer seems to occur - I see no way of assigning to a Jira user who is not available to @ mention in Mattermost.
image


@hanzei or @m1lt0n maybe you have thoughts on next steps based on this?

hanzei:
> Perhaps when passing in a user and the value does not start with `@` it should work the existing way with Jira user assignment?

That's a good solution. Let me do that change.

hanzei:
Closing the PR and adding https://github.com/mattermost/mattermost-plugin-jira/issues/558 to the @brightscout queue.

PR #785, opened 2021-07-20
mm-768 - prepends the siteURL in the redirectConnect method

maisnamrajusingh:
#### Summary
using the `getState().entities.general.config.SiteURL` to fetch the site url and prepend it to the path as mentioned in the issue 
comments. 
Fixes #768 
#### Ticket Link
https://github.com/mattermost/mattermost-plugin-jira/issues/768

PR #870, opened 2022-09-21
[GH-768]: Fixed Issue /Jira Connect doesn't open mattermost's subpath url

Kshitij-Katiyar:
**Summary** 

- If Mattermost is running on a subpath then /jira connect command was not including the subpath for the URL. Fixed this issue by including the subpath as well for creating the Oauth url.
- This PR contains all the changes from [Pr #785](https://github.com/mattermost/mattermost-plugin-jira/pull/785). So this PR alone is enough to fix this issue so we can close [Pr #785](https://github.com/mattermost/mattermost-plugin-jira/pull/785) afterward.

**Issue** 

- [Issue #768](https://github.com/mattermost/mattermost-plugin-jira/issues/768) 
- [Pr #785](https://github.com/mattermost/mattermost-plugin-jira/pull/785)


server/plugin.go
```diff
@@ -413,6 +414,10 @@ func (p *Plugin) GetSiteURL() string {
 	return p.getConfig().mattermostSiteURL
 }
 
+func (p *Plugin) CreateFullURLPath(extensionPath string) string {
```
javaguirre, line 417: I know the code for this is trivial, but It might change in the future, so it would be nice having a test to control whether we break it in the future with new logic. :-)

Let me send you a test example for this!

webapp/src/actions/index.ts
```diff
@@ -426,8 +426,10 @@ export function handleConnectFlow(instanceID?: string) {
 
 export function redirectConnect(instanceID: string) {
     return async (dispatch, getState) => {
+        const serverUrl = getPluginServerRoute(getState());
         const instancePrefix = '/instance/' + btoa(instanceID);
-        const target = '/plugins/' + PluginId + instancePrefix + '/user/connect';
+        const target = serverUrl + instancePrefix + '/user/connect';
```
javaguirre, line 431: I think it might be simpler to maintain to use strings format here? Also to be more cohesive with the changes below, where you are using string format.

```typescript
const target = `${serverUrl}${instancePrefix}/user/connect`;
```

webapp/src/selectors/index.ts
```diff
@@ -13,17 +13,21 @@ const getPluginState = (state) => state['plugins-' + PluginId] || {};
 
 export const getPluginServerRoute = (state) => {
     const config = getConfig(state);
-
     let basePath = '';
+    let url;
```
javaguirre, line 17: If we store in a variable only the part that changes, we don't need this to be a variable nor to use twice `new URL(...)`, let me give you an example!

```typescript
const baseUrl = config?.configSiteURL ? config.SiteUrl : window.location.href;
const basepath = new URL(baseUrl).pathname;
```

javaguirre (commented):
Hi! I'm Javier Aguirre, this is my first week @ Mattermost, nice to meet you!

Just wanted to comment on a few things, let me know what you think! :-)

javaguirre (changes requested):
Let me know when you have the opportunity to change it!

javaguirre (approved):
Great thank you! 👏

javaguirre:
Could you post a video on how to use this new change? Something like this so it's easier to QA.

https://github.com/mattermost/mattermost-plugin-jira/pull/863#issuecomment-1200475963

Thank you!

javaguirre:
@Kshitij-Katiyar tests seem to be failing, can you review them so we can pass them to QA?

The video would be great too.

Thank you!

Makefile
```diff
@@ -65,7 +65,7 @@ endif
 ## Ensures NPM dependencies are installed without having to run this all the time.
 webapp/.npminstall:
 ifneq ($(HAS_WEBAPP),)
-	cd webapp && $(NPM) install
+	cd webapp && $(NPM) install --verbose
```
mickmister, line 68: We've updated CI so this is no longer necessary. Can we remove this change here?

@hanzei There may be other lingering open PRs that have this change

webapp/src/selectors/index.ts
```diff
@@ -13,17 +13,14 @@ const getPluginState = (state) => state['plugins-' + PluginId] || {};
 
 export const getPluginServerRoute = (state) => {
     const config = getConfig(state);
+    const baseUrl = new URL(config?.SiteURL ?? window.location.href);
```
mickmister, line 16: I would prefer `||` here since I think `SiteURL` may be an empty string if not provided in the config

javaguirre: Can we make this change @Kshitij-Katiyar ?

Nityanand13: Sure

javaguirre (approved):

webapp/src/selectors/index.ts
```diff
@@ -13,17 +13,14 @@ const getPluginState = (state) => state['plugins-' + PluginId] || {};
...
 
-    let basePath = '';
-    if (config && config.SiteURL) {
-        basePath = new URL(config.SiteURL).pathname;
-
-        if (basePath && basePath[basePath.length - 1] === '/') {
-            basePath = basePath.substr(0, basePath.length - 1);
-        }
+    if (basePath && basePath[basePath.length - 1] === '/') {
+        basePath = basePath.substr(0, basePath.length - 1);
```
mickmister, line 20: If we use `window.location.href`, we'll just want to take the `origin` (as we are below), instead of including the `pathname`. If we take `pathname` from `window.location.href`, that will result in a value like `/core/messages/@dylan.haussermann` in the case of `https://community.mattermost.com/core/messages/@dylan.haussermann`.

Subpaths are only supported through the `SiteURL`, so if the `SiteURL` is not set, we can expect the server to not be using a subpath. The code that was here before seems correct.

Kshitij-Katiyar: @mickmister reverted this change

mickmister (approved):

hanzei:
@DHaussermann Heads up that the PR is ready for your review

DHaussermann:
Tested and failed. 
I don't see the fix here.
1. Does not seem to solve the issue on subpath servers. I see the same behavior. 
image
2. 
image
Upgrading to his PR seems to break something in the connection process for servers not on a subpath (tested on Cloud)

@Kshitij-Katiyar can you see if you can repo this?

Nityanand13:
@DHaussermann Fixed this issue.

DHaussermann:
@Kshitij-Katiyar Sorry for the delay on this. 

I have taken another look with the latest changes and I do see this get's 1 step further now but I see a new error on the Jira side.

image

Is there any way I can help you try to repro this or provide more information?

DHaussermann:
@mickmister I also meant to cc you on the comment above ☝️ 

Is this effort salvageable if we move to oAuth2? If you have any thoughts on when\if this is moving forward - it may be a better use of our eneegy of @Kshitij-Katiyar solution to subpaths is not applicable,

DHaussermann:
> @Kshitij-Katiyar Sorry for the delay on this.
> 
> I have taken another look with the latest changes and I do see this get's 1 step further now but I see a new error on the Jira side.
> 
> image
> 
> Is there any way I can help you try to repro this or provide more information?

@Kshitij-Katiyar sorry hold off on investigating this.I've just seen it again. I think it's unrelated to this PR.

Kshitij-Katiyar:
> > @Kshitij-Katiyar Sorry for the delay on this.
> > I have taken another look with the latest changes and I do see this get's 1 step further now but I see a new error on the Jira side.
> > image
> > Is there any way I can help you try to repro this or provide more information?
> 
> @Kshitij-Katiyar sorry hold off on investigating this.I've just seen it again. I think it's unrelated to this PR.

sure @DHaussermann I was also facing the same error on the main branch as well and also created an issue 
https://github.com/mattermost/mattermost-plugin-jira/issues/945

hanzei:
@AayushChaudhary0001 Would you be open to reviewing this PR?

mickmister:
@AayushChaudhary0001 Note that a customer has asked about this today

mickmister:
Nevermind, their question was about subdomains, not subpaths. The plugin already supports hosting MM on a subdomain so nothing to address there

AayushChaudhary0001 (approved):
The PR above has been tested for the following features:-
- If MM is running on subpath, then the /jira connect command post connection includes the subpath in the URL.

The PR was working fine for the above condition, LGTM. Approved.

PR #882, opened 2022-10-26
feat : dependabot workflow automation for updating dependency

Rajpratik71:
#### Summary

Dependabot workflow automation for updating dependency

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+version: 2
+updates:
+  # Dependencies listed in go.mod
+  - package-ecosystem: "gomod"
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
+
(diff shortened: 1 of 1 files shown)


hanzei (changes requested):
I'm not sure if this small project really needs auto updates. Security updates are already provided via Dependabot.

What values do auto updates for dependencies provide?

Rajpratik71:
Manual control of dependency is fine but with a growing no. of distributed upstream dependencies, it becomes hard to manage. So, for that automation should be there to update dependencies. Further, CI Pipeline is there to test those changes.


Further, this will not update the dependencies automatically, instead, a PR will be opened with changes that can be reviewed and tested with CI.

hanzei:
> Further, CI Pipeline is there to test those changes.

That is the part at worries me. CI does test something, but manually testing is still required. And doing that on every version bump PR seems like a lot of work.

PR #887, opened 2022-11-07
[GH-774] Add tooltip to Jira ticket links

### LINKED ISSUE #774: Add tooltip for jira ticket links

raghavaggarwal2308:
### Summary

Most of the code here is from https://github.com/mattermost/mattermost-plugin-jira/pull/835. The person who created https://github.com/mattermost/mattermost-plugin-jira/pull/835 PR was not responding for a while. So, we recreated the same PR with the typescript errors and some bugs fixed.

**Changes made-**
1. Fixed typescript errors.
2. Fixed bug -> ticket id not visible on the tooltip.

### Screenshots
Screenshot from 2022-11-07 22-13-40

### Ticket
Fixes #774  (Add tooltip to Jira ticket links)


javaguirre:
Can you remove your changes in the package-lock.json please? @raghavaggarwal2308 

It seems no deps have changed but the lock changed, if that's the case We could add that in a different PR, but I don't think it's needed, so it's better be safe than sorry IMHO.

Thank you!!

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,288 @@
...
+    init() {
+        let instanceID = '';
+        if (this.props.connectedInstances.length === 1) {
+            instanceID = this.props.connectedInstances[0].instance_id;
+        } else if (this.props.defaultUserInstanceID) {
+            instanceID = this.props.defaultUserInstanceID;
+        }
+
+        let ticketID = '';
+        if (this.props.href.includes('selectedIssue')) {
```
javaguirre, line 71: instead of these, why couldn't we have a function to extract the issue Key and then use in both blocks `getIssueByKey`? I think is easier to read that way.

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,288 @@
...
+
+        if (!ticketID && this.props.href.includes('atlassian.net/browse')) {
+            ticketID = this.props.href.split('|')[0].split('?')[0].split('/browse/')[1];
+            if (ticketID) {
+                this.props.getIssueByKey(ticketID, instanceID);
+            }
+        }
+    }
+
+    componentDidMount() {
```
javaguirre, line 84: In terms of legibility:

```typescript
const {connected, ticketDetails, ...} = props;
// Use them
if (connected && whatever && ...)...
```

You can do it for state too.

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,288 @@
...
+        if (this.props.connected && !this.state.isLoaded && ((this.props.ticketDetails && this.props.ticketDetails.ticketId) !== this.state.ticketId || !this.props.isLoaded)) {
+            this.init();
+        } else if (this.props.connected && !this.state.isLoaded && this.props.ticketDetails && this.props.ticketDetails.ticketId === this.state.ticketId) {
+            this.setTicket(this.props);
+        }
+    }
+
+    componentDidUpdate() {
+        if (this.props.isLoaded && !this.state.isLoaded && this.props.ticketDetails && this.props.ticketDetails.ticketId === this.state.ticketId) {
```
javaguirre, line 93: same

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,288 @@
...
+            );
+        }
+
+        return (<span/>);
+    }
+
+    tagTicketStatus(ticketStatus: string) {
+        let ticketStatusClass = 'default-style';
+
+        switch (ticketStatus.toLowerCase()) {
```
javaguirre, line 124: I'd prefer this approach, less code, easier to maintain and update:

```typescript
const myStatusClasses = {
  statusIndeterminate: 'ticket-status--indeterminate',
  statusDone: 'ticket-status-done',
  ....

 Then you can use it like:

try {
 myStatusClasses[ticketStatus.toLowerCase]
} catch { ... // default }
```

Using switch cases is almost always a bad idea in these cases, it's much easier to maintain a data struct and self-documented, easier to read and much less code to maintain (imagine having 20 statuses, for example).

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,288 @@
...
+        return (<div className='popover-labels__label'>{labelList}</div>);
+    }
+
+    render() {
+        if (!this.state.isLoaded) {
+            return (<p/>);
+        }
+
+        const {ticketDetails, href: jiraTicketURI} = this.state;
+        const jiraAvatar = ticketDetails && ticketDetails.jiraIcon;
```
javaguirre, line 181:
```typescript
ticketDetails?.jiraIcon is equivalent
```

javaguirre: why not creating a type with all these defaulting to null and then filling it if these properties are in the ticketDetails?

javaguirre (changes requested):

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
raghavaggarwal2308: > ```ts
> ticketDetails?.jiraIcon is equivalent
> ```

Optional chaining is not set up in this plugin. Created the below issue for the same:
https://github.com/mattermost/mattermost-plugin-jira/issues/889

raghavaggarwal2308:
@javaguirre Fixed the review fixes you mentioned

javaguirre:
> @javaguirre Fixed the review fixes you mentioned

Thank you @raghavaggarwal2308 , you can also re-request the review so I get a "stronger" notification.

javaguirre:
@raghavaggarwal2308 The package-lock.json is still a change, can you remove it from the changes of the PR? Thank you! :+1:

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,312 @@
...
+}
+
+export type State = {
+    href: string;
+    isLoaded: boolean;
+    ticketId: string;
+    ticketDetails?: TicketDetails;
+};
+
+const isAssigned = ' is assigned';
```
javaguirre, line 30: `isAssignedLabel` or `isAssignedText`, It seems a boolean constant otherwise.

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,312 @@
...
+const unAssigned = 'Unassigned';
```
javaguirre, line 31: UnassignedLabel.

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,312 @@
...
+    }
+
+    constructor(Props: Props) {
+        super(Props);
+        let ticketID = '';
+        if (this.props.href.includes('selectedIssue')) {
+            ticketID = this.props.href.split('selectedIssue=')[1].split('&')[0];
+        }
+
+        if (!ticketID && this.props.href.includes('atlassian.net/browse')) {
```
javaguirre, line 59: can we do this in a separate method? I think it's better for future testing and it seems a bit flaky to me.

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,312 @@
...
+        };
+    }
+
+    getIssueKey = () => {
+        let ticketID = '';
+        if (this.props.href.includes('selectedIssue')) {
+            ticketID = this.props.href.split('selectedIssue=')[1].split('&')[0];
+        }
+
+        if (!ticketID && this.props.href.includes('atlassian.net/browse')) {
```
javaguirre, line 76: is this the same as above? so +1 to separate it in another method.

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,312 @@
...
+        const ticketID = this.getIssueKey();
+        if (ticketID) {
+            this.props.getIssueByKey(ticketID, instanceID);
+        }
+    }
+
+    componentDidMount() {
+        const {connected, ticketDetails, isLoaded: isPropsLoaded} = this.props;
+        const {ticketId, isLoaded: isStateLoaded} = this.state;
+        if (connected && !isStateLoaded && ((ticketDetails && ticketDetails.ticketId) !== ticketId || !isPropsLoaded)) {
```
javaguirre, line 99: this line and the condition in 101 is almost the same, it's better to move it to a method/function and use it twice in these lines, also easier to test.

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,312 @@
...
+        if (myStatusClass) {
+            ticketStatusClass += myStatusClass;
+        } else {
+            ticketStatusClass += ' ticket-status--default';
+        }
+
+        return <span className={ticketStatusClass}>{ticketStatus}</span>;
+    }
+
+    labelList(labels: string[]) {
```
javaguirre, line 150: if it's a render it's nice to add it to the function, like `renderLabelList`.

javaguirre (changes requested):
Thank you for the effort! We are almost there!

mickmister:
Can we add an assigning function to this? That would make this feature immensely more powerful. The ability to choose an assignee on the tooltip, just as you can in the create issue modal

raghavaggarwal2308:
> Can we add an assigning function to this? That would make this feature immensely more powerful. The ability to choose an assignee on the tooltip, just as you can in the create issue modal

@mickmister Sounds like an amazing feature but this will be a whole new feature and will take some time to implement. So, we think that instead of keeping this PR on hold we can create a separate issue for this one. Please let us know your thoughts on this.

mickmister:
@raghavaggarwal2308 Sure, makes sense. If we do something like that later, I think we should just have a `Edit Issue` button that opens a modal similar to the `Create Issue` modal, except it updates an existing issue instead of creating a new one. I think this would be :100: :100: :100:

raghavaggarwal2308:
@mickmister Create an issue on Github for the same. Feel free to make any changes if I missed something.
Issue Link:
https://github.com/mattermost/mattermost-plugin-jira/issues/899

webapp/src/reducers/index.js
```diff
@@ -193,6 +193,39 @@ const channelSubscriptions = (state = {}, action) => {
...
+            ...state,
+            isLoaded: true,
+            ticketDetails,
+        };
+    }
+    case ActionTypes.RECEIVED_JIRA_TICKETS_ERROR :
+        return {
+            ...state,
+            error: action.error,
+        };
```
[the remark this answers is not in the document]
mickmister: Nevermind I see this is in a reducer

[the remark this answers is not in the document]
mickmister: Why is this function called `getIssueByKey`? It stores at most one ticket's information at any given time

webapp/src/selectors/index.ts
```diff
@@ -38,6 +38,7 @@ export const getCurrentUserLocale = createSelector(
     }
 );
 
+export const isEnabled = (state) => getPluginState(state).enabled;
```
mickmister, line 41: Why is this necessary?

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,312 @@
...
+                                height='14'
+                                src={jiraIssueIconURI}
+                            />
+                        </a>
+                    </div>
+                </div>
+                <div className='popover-body'>
+                    <div className='popover-body__title'>
+                        <a href={jiraTicketURI}>
+                            <h5>{this.truncateString(jiraTicketTitle as string, jiraTicketTitleMaxLength)}</h5>
```
mickmister, line 249: Why do we need to cast this `as string`? I'm thinking we should let typescript take care of this by checking if `ticketDetails` is defined, and return `null` or some other short-circuiting return value so we don't need to cast things like this. All of the `as *` here are a little worrisome.

webapp/src/components/jira_ticket_tooltip/index.ts
```diff
@@ -0,0 +1,24 @@
...
+import {isUserConnected, getIssue, getUserConnectedInstances, getDefaultUserInstanceID} from 'selectors';
+import {getIssueByKey} from 'actions';
+
+import TicketPopover from './jira_ticket_tooltip';
+
+const mapStateToProps = (state: GlobalState) => {
+    return {
+        connected: isUserConnected(state),
+        ticketDetails: getIssue(state).ticketDetails,
+        isLoaded: getIssue(state).isLoaded,
```
mickmister, line 14: What happens when we hover over one Jira issue link, and then hover over a different one?

webapp/src/components/jira_ticket_tooltip/index.ts
```diff
@@ -0,0 +1,24 @@
...
+        defaultUserInstanceID: getDefaultUserInstanceID(state),
+        connectedInstances: getUserConnectedInstances(state),
+    };
+};
+
+const mapDispatchToProps = (dispatch: Dispatch) => bindActionCreators({
+    getIssueByKey,
```
mickmister, line 21: Since this is an external operation, I think this should be `fetchIssueByKey`

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,312 @@
...
+            ticketID = this.props.href.split('selectedIssue=')[1].split('&')[0];
+        }
+
+        if (!ticketID && this.props.href.includes('atlassian.net/browse')) {
+            ticketID = this.props.href.split('|')[0].split('?')[0].split('/browse/')[1];
+        }
+
+        this.state = {
+            href: this.props.href,
+            isLoaded: false,
```
mickmister, line 65: Having this `isLoaded` state here and in redux seems redundant

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
mickmister: Also we should do something like:

```js
const url = new URL(this.props.href);
if (!url.hostname.endsWith('.atlassian.net') || !url.pathname.startsWith('/browse/')) {
    return;
}
```

Also what about Jira Server instances? We should instead be looking at `connectedInstances` to figure out the available Jira domains

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,312 @@
...
+export default class TicketPopover extends React.PureComponent<Props, State> {
+    truncateString(str: string, num: number) {
+        if (num > str.length) {
+            return str;
+        }
+        return `${str.substring(0, num)}...`;
+    }
+
+    constructor(Props: Props) {
+        super(Props);
```
mickmister, line 53:
```suggestion
    constructor(props: Props) {
        super(props);
```

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,312 @@
...
+    [myStatus.DONE]: ' ticket-status--done',
+};
+
+export default class TicketPopover extends React.PureComponent<Props, State> {
+    truncateString(str: string, num: number) {
+        if (num > str.length) {
+            return str;
+        }
+        return `${str.substring(0, num)}...`;
+    }
```
mickmister, line 50: This can be defined as a flat function in another file

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,312 @@
...
+
+    constructor(Props: Props) {
+        super(Props);
+        let ticketID = '';
+        if (this.props.href.includes('selectedIssue')) {
+            ticketID = this.props.href.split('selectedIssue=')[1].split('&')[0];
+        }
+
+        if (!ticketID && this.props.href.includes('atlassian.net/browse')) {
+            ticketID = this.props.href.split('|')[0].split('?')[0].split('/browse/')[1];
```
mickmister, line 60: This looks error-prone. When is `|` present in the URL, and when is it not?

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,312 @@
...
+        }
+    }
+
+    componentDidMount() {
+        const {connected, ticketDetails, isLoaded: isPropsLoaded} = this.props;
+        const {ticketId, isLoaded: isStateLoaded} = this.state;
+        if (connected && !isStateLoaded && ((ticketDetails && ticketDetails.ticketId) !== ticketId || !isPropsLoaded)) {
+            this.init();
+        } else if (connected && !isStateLoaded && ticketDetails && ticketDetails.ticketId === ticketId) {
+            this.setTicket(this.props);
```
mickmister, line 102: When is `isStateLoaded` ever going to be `true` in this function `onComponentDidMount`?

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,312 @@
...
+    }
+
+    labelList(labels: string[]) {
+        if (!labels.length) {
+            return null;
+        }
+
+        let totalString = 0;
+        let totalHide = 0;
+        const labelList = labels.map((label: string, key: number): JSX.Element => {
```
mickmister, line 157: This way you can return `null` below instead of `<></>`

```suggestion
        const labelList = labels.map((label: string, key: number): React.ReactNode => {
```

mickmister, line 157: This is confusing to have a variable we're manipulating in a `map` function. Is there another way to implement this?

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,312 @@
...
+    init() {
+        let instanceID = this.props.defaultUserInstanceID || '';
+
+        if (this.props.connectedInstances.length === 1) {
+            instanceID = this.props.connectedInstances[0].instance_id;
+        }
+
+        const ticketID = this.getIssueKey();
+        if (ticketID) {
+            this.props.getIssueByKey(ticketID, instanceID);
```
mickmister, line 92: We should check all connected instances here to find a URL match

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
mickmister: Also what value does `truncateString` give us? Why not just `jiraTicketTitle.substring(0, jiraTicketTitleMaxLength)`

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,312 @@
...
+                        {this.fixVersionLabel(jiraTicketVersions as string)}
+                        {this.labelList(jiraTicketLabels as string[])}
+                    </div>
+                </div>
+                <div className='popover-footer'>
+                    {jiraTicketAssigneeAvatarURI ? (
+                        <img
+                            className='popover-footer__assigner-profile'
+                            src={jiraTicketAssigneeAvatarURI}
+                            alt='jira assigner profile'
```
mickmister, line 266:
```suggestion
                            className='popover-footer__assignee-profile'
                            src={jiraTicketAssigneeAvatarURI}
                            alt='jira assignee profile'
```

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,312 @@
...
+                        />
+                    ) : (
+                        <span className='default-avatar'>
```
mickmister, line 269: Can we define this component in another file? Maybe create a generic `Avatar` component, that does this calculation of default avatar vs specific avatar

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,312 @@
...
+            jiraIssueIconURI = ticketDetails.issueIcon;
+            jiraTicketKey = ticketDetails.ticketId;
+            jiraTicketTitle = ticketDetails.summary;
+            jiraTicketAssigneeAvatarURI = ticketDetails.assigneeAvatar;
+            jiraTicketAssigneeName = ticketDetails.assigneeName;
+            jiraTicketStatusName = ticketDetails.statusKey;
+            jiraTicketDescription = ticketDetails.description;
+            jiraTicketVersions = ticketDetails.versions;
+            jiraTicketLabels = ticketDetails.labels;
+        }
```
mickmister, line 214: This seems like an anti-pattern. I'm thinking accessing `ticketDetails` directly in the rendering parts may be better, unless you have reasons why this is better. What do you think @raghavaggarwal2308?

webapp/src/action_types/index.js
```diff
@@ -32,4 +32,7 @@ export default {
 
     RECEIVED_CHANNEL_SUBSCRIPTIONS: `${PluginId}_recevied_channel_subscriptions`,
     DELETED_CHANNEL_SUBSCRIPTION: `${PluginId}_deleted_channel_subscription`,
+
+    RECEIVED_JIRA_TICKETS: `${PluginId}_received_jira_tickets`,
+    RECEIVED_JIRA_TICKETS_ERROR: `${PluginId}_received_jira_tickets_error`,
```
mickmister, line 37: I think we only fetch one issue at a time right?

```suggestion
    RECEIVED_JIRA_TICKET: `${PluginId}_received_jira_ticket`,
    RECEIVED_JIRA_TICKET_ERROR: `${PluginId}_received_jira_ticket_error`,
```

mickmister (changes requested):
Great work on this @raghavaggarwal2308! I have some requests on the PR. Overall looks awesome!

mickmister:
@raghavaggarwal2308 Is this ready for re-review?

raghavaggarwal2308:
@mickmister We are working on the review comments added by you. Review comments added by @javaguirre are fixed.

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,311 @@
...
+            href: this.props.href,
+            isLoaded: false,
+            ticketId: ticketID,
+        };
+    }
+
+    getIssueKey = () => {
+        let ticketID = '';
+        if (this.props.href.includes('selectedIssue')) {
+            ticketID = this.props.href.split('selectedIssue=')[1].split('&')[0];
```
javaguirre, line 65: I think we should catch errors here, since there are many indexes that can go wrong and it will fail if one of those is not there.

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,311 @@
...
+        }
+
+        if (!ticketID && this.props.href.includes('atlassian.net/browse')) {
+            ticketID = this.props.href.split('|')[0].split('?')[0].split('/browse/')[1];
```
javaguirre, line 69: dito

javaguirre (changes requested):

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
raghavaggarwal2308: @mickmister @javaguirre We are checking the URL with all the connected instances

webapp/src/components/jira_ticket_tooltip/index.ts
raghavaggarwal2308: We get the tooltip of the second link in that case, which was already fetched when the link was posted in the channel.

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
raghavaggarwal2308: Removed this function

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
raghavaggarwal2308: I don't think it will ever be `true` here

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,312 @@
...
+import {Dispatch} from 'redux';
+
+import {Instance} from 'types/model';
+
+export type Props = {
+    href: string;
+    show: boolean;
+    connected: boolean;
+    isLoaded?: boolean;
+    ticketDetails?: TicketDetails;
```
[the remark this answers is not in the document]
raghavaggarwal2308: You can find the type [here](https://github.com/mattermost/mattermost-plugin-jira/pull/887/files#diff-a9af1196bc28f048bdbba0627546ea48757f1b6cfcf351239592a81349a65c80)

raghavaggarwal2308:
@javaguirre @mickmister Fixed the review comments you mentioned

javaguirre (changes requested):

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,256 @@
...
+        let instanceID = '';
+
+        for (const instance of this.props.connectedInstances) {
+            instanceID = instance.instance_id;
+
+            try {
+                if (this.props.href.includes(instanceID)) {
+                    if (this.props.href.includes('selectedIssue=')) {
+                        ticketID = this.props.href.split('selectedIssue=')[1].split('&')[0];
+                        break;
```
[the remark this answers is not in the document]
raghavaggarwal2308: If we get a URL match we do not need to iterate over the remaining instances

raghavaggarwal2308:
@javaguirre Fixed the review comments you mentioned

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,257 @@
...
+        let ticketID = '';
+        let instanceID = '';
+
+        for (const instance of this.props.connectedInstances) {
+            instanceID = instance.instance_id;
+
+            try {
+                if (this.props.href.includes(instanceID)) {
+                    if (this.props.href.includes('selectedIssue=')) {
+                        ticketID = this.props.href.split('selectedIssue=')[1].split('&')[0];
```
javaguirre, line 61: I think this might be much easier with a regex, and we can do both searches at the same time, but it's probably a nice to have now. It would more readable and easier to understand where it's failing.

Something like this https://regex101.com/r/OxaPmy/1


Screenshot 2023-01-12 at 09 53 28

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,257 @@
...
+    }
+
+    getIssueKey = () => {
+        let ticketID = '';
+        let instanceID = '';
+
+        for (const instance of this.props.connectedInstances) {
+            instanceID = instance.instance_id;
+
+            try {
```
javaguirre, line 58: We could save some time continuing when we know there is no instanceID.

```suggestion
            if(!this.props.href.includes(instanceID) {
              continue;
            }
            
            try { ... }
```

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,257 @@
...
+    renderLabelList(labels: string[]) {
+        if (!labels.length) {
+            return null;
+        }
+
+        return (
+            <div className='popover-labels__label'>
+                {
+                    labels.map((label: string, key: number): ReactNode => {
+                        // Return an element for the first three labels and if there are more than three labels, then return a combined label for the remaining labels
```
javaguirre, line 151: 💯

javaguirre (changes requested):
Thank you very much for your hard work! I think we are almost there :-)

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
raghavaggarwal2308: @javaguirre I think that's what is happening currently. On line 59, if the instanceID does not match we will jump to the next iteration. Let me know if I am misinterpreting it!

javaguirre: It's true, but this way you have less nested conditions and it's easier to follow, it's a nice to have though.

raghavaggarwal2308:
@javaguirre Fixed the review comments you mentioned

javaguirre (approved):
Thank you for the hard work 💯

mickmister (changes requested):
Thanks for addressing the concerns @raghavaggarwal2308 :+1:

I have a few more questions about the PR. I also want to take some time and run the code to see its behavior in different situations. I've left some feedback in the meantime

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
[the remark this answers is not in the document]
raghavaggarwal2308: @mickmister By making the type global we can use it whereever we want directly and in a code editor like vscode we can navigate to it by using `Ctrl + click`.

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,261 @@
...
+        if (ticketID) {
+            this.props.fetchIssueByKey(ticketID, instanceID);
+        }
+    }
+
+    isUserConnected() {
+        const {connected} = this.props;
+        const {isLoaded} = this.state;
+
+        return connected && !isLoaded;
```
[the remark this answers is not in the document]
raghavaggarwal2308: I think there was a confusion because of the name of the function. Updated the name now

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,261 @@
...
+
+    render() {
+        if (!this.state.isLoaded) {
+            return (<p/>);
+        }
+
+        const {ticketDetails} = this.state;
+
+        return (
+            <div className='jira-issue-tooltip'>
```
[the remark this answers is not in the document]
raghavaggarwal2308: Now we are displaying a skeleton loader till the details are loaded.

raghavaggarwal2308:
@mickmister FIxed the review fixes you mentioned

webapp/src/types/tooltip.d.ts
```diff
@@ -0,0 +1,42 @@
...
+    project: {avatarUrls: AvatarUrls};
+    versions: string[];
+    status: {name: string};
+    issuetype: {iconUrl: string};
+}
+
+type Action = {
+    type: string;
+    data: TicketData;
+}
```
mickmister, line 42: There are other actions in this plugin, so it seems wrong to call this specific action `Action`

webapp/src/utils/jira_issue_metadata.test.tsx
```diff
@@ -590,4 +590,99 @@ describe('utils/jira_issue_metadata', () => {
             expect(actual).toEqual('Project = KT AND IssueType IN (Bug) AND Priority IS EMPTY');
         });
     });
+
+    describe('jiraIssueToReducer', () => {
+        it('should return the ticket details with all fields', () => {
+            const action: Action = {
+                data: {
```
mickmister, line 597: It's unfortunate we need this unit test, when we could just store the jira issue itself in redux instead. It's more future-proof that way, as we only need to edit where it's being used, and not edit the reducer.

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+            const wrapper = shallow(
+                <TicketPopover
+                    {...props}
+                    href='https://something.atlassian.net/browse/TICKET-1234'
+                />
+            );
+            const instance = wrapper.instance() as TicketPopover;
+            const expectedOutput = {ticketID: 'TICKET-1234', instanceID: '1'};
+            expect(instance.getIssueKey()).toEqual(expectedOutput);
+        });
```
mickmister, line 39: Isn't `instance_id` supported to be a URL? Why did `1` pass for the instance `https://something.atlassian.net` here?

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+        test('should return the expected output when URL matches the second regex pattern', () => {
+            const wrapper = shallow(
+                <TicketPopover
+                    {...props}
+                    href='https://something.atlassian.net/jira/issues/?selectedIssue=TICKET-1234'
+                />
+            );
+            const instance = wrapper.instance() as TicketPopover;
+            const expectedOutput = {ticketID: 'TICKET-1234', instanceID: '1'};
+            expect(instance.getIssueKey()).toEqual(expectedOutput);
```
mickmister, line 50: Same question here

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+        test('should return null when URL does not match any pattern', () => {
+            const wrapper = shallow(
+                <TicketPopover
+                    {...props}
+                    href='https://something.atlassian.net/not-a-ticket'
+                />
+            );
+            const instance = wrapper.instance() as TicketPopover;
+            expect(instance.getIssueKey()).toEqual(null);
```
mickmister, line 61: Can we add some more cases to avoid false positives/negatives and manage corner cases? @DHaussermann We'll want to think of every way a user may be copying a Jira issue URL with the intention to paste in Mattermost.

webapp/src/utils/jira_issue_metadata.tsx
```diff
@@ -356,3 +356,21 @@ export function generateJQLStringFromSubscriptionFilters(issueMetadata: IssueMet
 
     return [projectJQL, issueTypesJQL, filterFieldsJQL].filter(Boolean).join(' AND ');
 }
+
+export function jiraIssueToReducer(action: Action): TicketDetails {
```
mickmister, line 360: We should use existing types for these from `types/model`

mickmister (changes requested):

webapp/src/utils/jira_issue_metadata.test.tsx
mickmister: Also, we can use more realistic test data from existing test data files https://github.com/mattermost/mattermost-plugin-jira/tree/master/webapp/src/testdata

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,277 @@
...
+                )
+                }
+                <div className='popover-footer'>
+                    {!ticketDetails || ticketDetails.assigneeAvatar ? (
+                        <img
+                            className={`popover-footer__assignee-profile ${!ticketDetails && 'skeleton-loader'}`}
+                            src={ticketDetails && ticketDetails.assigneeAvatar}
+                            alt={ticketDetails ? 'jira assignee profile' : ''}
+                        />
+                    ) : <DefaultAvatar/>
```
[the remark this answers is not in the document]
mickmister: Generally I think we should have a separate block or method like `renderTicketIsLoading()` so we can have completely separated logic for the two situations.

webapp/src/utils/jira_issue_metadata.tsx
```diff
@@ -356,3 +356,21 @@ export function generateJQLStringFromSubscriptionFilters(issueMetadata: IssueMet
...
+    const assignee = action.data && action.data.fields && action.data.fields.assignee ? action.data.fields.assignee : null;
+    const ticketData = action.data;
+    const ticketDetails: TicketDetails = {
+        assigneeName: assignee && assignee.displayName ? assignee.displayName : '',
+        assigneeAvatar: assignee && assignee.avatarUrls && assignee.avatarUrls['48x48'] ? assignee.avatarUrls['48x48'] : '',
+        labels: ticketData.fields.labels,
+        description: ticketData.fields.description,
+        summary: ticketData.fields.summary,
+        ticketId: ticketData.key,
+        jiraIcon: ticketData.fields.project.avatarUrls['48x48'],
```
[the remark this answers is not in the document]
raghavaggarwal2308: I think so. But, if in any case this is not defined then we will display the default avatar.

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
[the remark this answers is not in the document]
raghavaggarwal2308: We are displaying a skeleton loader until the details are loaded. 

For the avatar: we are displaying a skeleton loader until the ticket details are not loaded and a default avatar if it is not defined.

If we create a separate function for the loading state, we will repeat a lot of code as we are just using a class-name on the actual components to display the loader.

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,277 @@
...
+                            target='_blank'
+                            rel='noopener noreferrer'
+                        >
+                            <img
+                                src={ticketDetails && ticketDetails.jiraIcon}
+                                width={14}
+                                height={14}
+                                alt={ticketDetails ? 'jira-avatar' : ''}
+                                className={`popover-header__avatar ${!ticketDetails && 'skeleton-loader'}`}
+                            />
```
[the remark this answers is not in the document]
raghavaggarwal2308: We are adding a classname `skeleton-loader` until the details are not loaded

webapp/src/utils/jira_issue_metadata.test.tsx
raghavaggarwal2308: the data we require is not present in the testData directory. Do you want us to create a new file?

raghavaggarwal2308:
@mickmister Fixed the review comments you mentioned

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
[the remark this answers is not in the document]
mickmister: > If we create a separate function for the loading state, we will repeat a lot of code as we are just using a class-name on the actual components to display the loader.

I don't think this is the case.The majority of the render function consists of ternaries and conditionals specific to when we have ticket details available. I would prefer to separate a block/function for this that says "this is what is rendered when we are loading"

Can you a gif of what this looks like so I can better understand the current structure?

webapp/src/utils/jira_issue_metadata.test.tsx
mickmister: Yes that sounds good

webapp/src/utils/jira_issue_metadata.tsx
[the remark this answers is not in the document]
mickmister: Note that in the case of `avatarUrls` not being defined, this will throw an exception.

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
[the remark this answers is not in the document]
mickmister: @raghavaggarwal2308 Why use an `img` tag with an invalid `src`, instead of a `div` in replacement of the `img` tag?

Also when `ticketDetails` is defined, we are adding `false` to the class name. We should avoid doing this.

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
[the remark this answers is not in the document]
raghavaggarwal2308: @mickmister 
[screen-capture.webm](https://user-images.githubusercontent.com/55234496/231369600-cfe888e3-0111-4895-a01c-ce4175f5a307.webm)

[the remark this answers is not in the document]
raghavaggarwal2308: @mickmister 
[Skeleton-loader.webm](https://user-images.githubusercontent.com/55234496/231370609-ced61a79-9b85-49a2-9c3d-10bda84262a0.webm)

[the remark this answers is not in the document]
mickmister: > @mickmister [Skeleton-loader.webm](https://user-images.githubusercontent.com/55234496/231370609-ced61a79-9b85-49a2-9c3d-10bda84262a0.webm)

This is a slick look, nice work on this :+1:

@matthewbirtch do you have any thoughts on this skeleton loader approach?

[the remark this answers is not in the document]
matthewbirtch: Nice! Can we just make a small adjustment to make these circles and remove the borders on them?

Screenshot 2023-04-18 at 9 51 24 AM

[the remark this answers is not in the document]
raghavaggarwal2308: Sure, @matthewbirtch

raghavaggarwal2308:
@mickmister @matthewbirtch FIxed the review comments you mentioned
[screen-capture (3).webm](https://user-images.githubusercontent.com/55234496/233415186-4f8fe6c0-2f3b-43cf-8986-a8ca37dff72a.webm)

mickmister:
The skeleton loader is a nice touch, but it seems a bit off since the rest of the application doesn't have this. Just my opinion, but I think we shouldn't use this here. Though I'll defer to @matthewbirtch of course.

matthewbirtch:
> The skeleton loader is a nice touch, but it seems a bit off since the rest of the application doesn't have this. Just my opinion, but I think we shouldn't use this here. Though I'll defer to @matthewbirtch of course.

That's a fair point @mickmister. You're right, the skeleton loader pattern is not used much in the webapp at all. The only other place we currently have a skeleton loader is on the Insights page. To be honest, I'd be fine if this was just a spinner loader instead. this might be a bit more visual noise than is needed for this case.

raghavaggarwal2308:
@mickmister @matthewbirtch Does it look good now?
[Demo video for spinner loader](https://user-images.githubusercontent.com/55234496/235860313-26b3e98d-e016-49ec-b134-35cfbda5e87e.webm)

matthewbirtch:
> @mickmister @matthewbirtch Does it look good now? [Demo video for spinner loader](https://user-images.githubusercontent.com/55234496/235860313-26b3e98d-e016-49ec-b134-35cfbda5e87e.webm)

Thanks @raghavaggarwal2308. Can we reduce the size of the spinner to 28x28? Also, maybe we can adjust the color of the spinner to 72% opacity to tone it down a little?

raghavaggarwal2308:
@matthewbirtch 
[Demo video for spinner with reduces size](https://user-images.githubusercontent.com/55234496/235932054-6f622531-18d7-495c-936c-2ee6d291a6b5.webm)

matthewbirtch:
> @matthewbirtch How about now? [Demo video for spinner with reduces size](https://user-images.githubusercontent.com/55234496/235932054-6f622531-18d7-495c-936c-2ee6d291a6b5.webm)

Looks good to me @raghavaggarwal2308. Eventually we will probably update the style of the spinner, but since this is used elsewhere in the product right now, we'll leave it as you have it.

Is there a test server where I can actually test this out and do a more detailed review of the fonts/spacing/etc?

mickmister:
@DHaussermann Can I use your MM server dedicated to Jira QA to deploy this PR, so @matthewbirtch can take a look at the PR?

DHaussermann:
Hey @matthewbirtch I have setup a server that has this running here https://dkh-jira-tool-tip.test.mattermost.cloud and our test cloud instance installed. 
image

DM me if you need credentials to the server. Also if you need me to make you an account on the test Jira instance let me know.

matthewbirtch:
Thanks @DHaussermann. This is perfect.

@raghavaggarwal2308, only one minor thing I noticed: the status tag is set to `sans-serif` which will just use the system default sans-serif font. Can you update it to use `font-family: 'Open Sans', sans-serif;`?

Other than that, I think we're good!

raghavaggarwal2308:
@matthewbirtch Fixed the font-family

matthewbirtch (approved):

mickmister (changes requested):

raghavaggarwal2308:
@mickmister Fixed the review comments you mentioned

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -0,0 +1,297 @@
...
+        }
+
+        return null;
+    }
+
+    isUserConnectedAndStateNotLoaded() {
+        const {connected} = this.props;
+        const {ticketDetails} = this.state;
+
+        return connected && ticketDetails;
```
mickmister, line 87: Since this method name implies that we are returning a boolean

```suggestion
        return Boolean(connected && ticketDetails);
```

raghavaggarwal2308:

raghavaggarwal2308:
@DHaussermann Gentle reminder for review

raghavaggarwal2308:
@DHaussermann Gentle reminder for review

raghavaggarwal2308:
@DHaussermann Gentle reminder for review

hanzei:
@AayushChaudhary0001 Would you be open on reviewing this PR?

mickmister (approved):

AayushChaudhary0001 (commented):
Tested this PR.
@raghavaggarwal2308 I am facing an issue that the tooltip is not getting rendered and undergoing loading state for a very long period.

Screenshot from 2023-12-01 14-23-54

AayushChaudhary0001:
Tested and approved, LGTM.

AayushChaudhary0001 (approved):

PR #894, opened 2022-11-28
[GH-626]: Supported filtering on comment visibility for subscriptions

Nityanand13:
#### Summary
Added "Comment Visibility" field to subscription modal to filter notifications in the channel. 
1. When a user creates a comment in Jira then the plugin is getting notified through a webhook. Now, with this [API](https://brightscout-dev.atlassian.net/rest/api/2/issue/10001/comment/11063) we are getting all the information of a comment. If that particular comment has a visibility attribute that it should be visible to the users who belong to a certain group. Then notification is sent to the subscribed channel in mattermost only when that particular group is included in the subscriptions and vice versa.
2. Apart from adding groups to which a user belongs, he can also add a group ‘visible-to-all-users’. If a particular comment has no visibility attribute, then notification is sent to the subscribed channel in mattermost only when ‘visible-to-all-users’ is present or no filter is present on the mattermost side.

#### Ticket Link
Fixes [#626](https://github.com/mattermost/mattermost-plugin-jira/issues/626)


server/client.go
```diff
@@ -252,6 +255,18 @@ type AutoCompleteResult struct {
 	Results []Result `json:"results"`
 }
 
+type Item struct {
```
javaguirre, line 258: What's Item? We can find a more relevant name I think

server/client.go
```diff
@@ -252,6 +255,18 @@ type AutoCompleteResult struct {
...
+	Name string `json:"name"`
+}
+
+type Group struct {
```
javaguirre, line 262: Is this a UserGroup?

javaguirre: or VisibilityUserGroup? Is this group being used for anything else?

server/webhook_worker.go
```diff
@@ -49,7 +51,30 @@ func (ww webhookWorker) process(msg *webhookMessage) (err error) {
 		return err
 	}
 
-	channelsSubscribed, err := ww.p.getChannelsSubscribed(v, msg.InstanceID)
+	// To check if this is a comment-related webhook payload
+	isCommentEvent := wh.Events().Intersection(commentEvents).Len() > 0
```
javaguirre, line 55: I think a big part of this code can be done as a function/method some place else and call it here, you could then remove the comment if you put a relevant name to the function and it's easier to maintain/test in the future since the function has only one responsibility (Check if this is a comment-related webhook payload).

webapp/src/utils/jira_issue_metadata.tsx
```diff
@@ -264,6 +276,10 @@ export function isLabelField(field: JiraField | FilterField): boolean {
     return field.schema.system === 'labels' || field.schema.custom === 'com.atlassian.jira.plugin.system.customfieldtypes:labels';
 }
 
+export function isCommentVisibilityField(field: JiraField | FilterField): boolean {
+    return field.key === 'commentVisibility';
```
javaguirre, line 280: You are using the same string key in 280 and 215, I think you can use a constant for that key and use it in both places to void future problems.

javaguirre (changes requested):

server/client.go
Nityanand13: Now I have changed this to `GroupItem`. It indicates the individual group.

server/client.go
Nityanand13: Now I have changed this to `GroupsInfo`.It indicate all the groups to which a commentator belongs to.

mickmister: How about `JiraUserGroupCollection` for the collection and `JiraUserGroup` for the singular group?

hanzei:
@mickmister Head up that the PR is ready for your  review

Kshitij-Katiyar:
@mickmister Fixed the merge conflicts and this PR is ready for review.

server/webhook_worker.go
```diff
@@ -55,7 +80,14 @@ func (ww webhookWorker) process(msg *webhookMessage) (err error) {
 		return err
 	}
 
-	channelsSubscribed, err := ww.p.getChannelsSubscribed(v, msg.InstanceID)
+	visibilityAttribute := ""
+	if isCommentRelatedWebhook(wh) {
+		if visibilityAttribute, err = ww.getVisibilityAttribute(msg, v); err != nil {
+			return err
+		}
+	}
```
mickmister, line 88: If the comment author is not connected to Mattermost, we end up not processing the webhook event no matter what. I think there's a design gap in the security model of this feature. This is a difficult problem to solve, the case of a comment author not connected to Mattermost.

---

Let's say a subscription is set to "comment visibility must be empty". If we can't fetch the comment because the author is not connected to MM, then we can't know the visibility of the comment. And so we would then always be avoiding the webhook events with every comment made by a user that is not connected. That's not really behavior that we want.

Though we can only assume that it is a sensitive comment if we can't fetch its visibility. Otherwise the feature would create posts for sensitive comments.

@esarafianou Do you have any thoughts on this? Essentially, this feature allows subscriptions to filter out sensitive comments, but if the comment author is not connected to MM, then we have no way of knowing whether the comment is sensitive or not. Defaulting to secure avoids unwanted data leakage, but it then requires all comment authors to be connected to MM for any comments to be posted ever.

server/client.go
```diff
@@ -26,8 +26,10 @@ import (
 )
 
 const autocompleteSearchRoute = "2/jql/autocompletedata/suggestions"
+const commentVisibilityRoute = "2/user"
```
mickmister, line 29: I'm thinking we want to use [project roles](https://developer.atlassian.com/cloud/jira/platform/rest/v2/api-group-project-roles/#api-rest-api-2-project-projectidorkey-role-get) for this

I have a WIP here that implements the data fetching piece, but not the comment filtering piece https://github.com/mattermost/mattermost-plugin-jira/compare/master...comment-security

https://github.com/mattermost/mattermost-plugin-jira/blob/a86928d1d0c0c2493df5a2690ea8f54f515fd9a5/server/autocomplete_search.go#L108-L121

server/issue.go
```diff
@@ -384,6 +388,47 @@ func (p *Plugin) GetCreateIssueMetadataForProjects(instanceID, mattermostUserID
...
+	client, _, connection, err := p.getClient(types.ID(instanceID), types.ID(mattermostUserID))
+	if err != nil {
+		return http.StatusInternalServerError, err
+	}
+
+	params := map[string]string{
+		"fieldValue": r.FormValue(fieldValueQueryParam),
+		"expand":     r.FormValue(expandQueryParam),
+		"accountId":  connection.AccountID,
+	}
```
mickmister, line 411: Seems weird to allow the frontend to specify the `expand` value

server/subscribe.go
```diff
@@ -124,7 +125,7 @@ func (p *Plugin) getUserID() string {
 	return p.getConfig().botUserID
 }
 
-func (p *Plugin) matchesSubsciptionFilters(wh *webhook, filters SubscriptionFilters) bool {
+func (p *Plugin) matchesSubsciptionFilters(wh *webhook, filters SubscriptionFilters, visibilityAttribute string) bool {
```
mickmister, line 128: I'm thinking we can add a `commentVisibility` field onto the `webhook` struct, and avoid adding this `visibilityAttribute` param to multiple functions

server/subscribe.go
```diff
@@ -170,6 +171,12 @@ func (p *Plugin) matchesSubsciptionFilters(wh *webhook, filters SubscriptionFilt
 		}
 
 		value := getIssueFieldValue(issue, field.Key)
+		if visibilityAttribute != "" {
+			value[visibilityAttribute] = true
+		} else if field.Key == CommentVisibility {
+			value[visibleToAllUsers] = true
+		}
+
```
mickmister, line 179: I think this may read a little better, as the implementation of `StringSet` as a map should be hidden away:

```suggestion
		value := getIssueFieldValue(issue, field.Key)
		if visibilityAttribute != "" {
			value.Add(visibilityAttribute)
		} else if field.Key == CommentVisibility {
			value.Add(visibleToAllUsers)
		}
```

Also I think the first block be inside a `field.Key == CommentVisibility` condition like so:

```go
		value := getIssueFieldValue(issue, field.Key)
		if field.Key == CommentVisibility {
			if visibilityAttribute != "" {
				value.Add(visibilityAttribute)
			} else {
				value.Add(visibleToAllUsers)
			}
		}
```

webapp/src/components/modals/channel_subscriptions/__snapshots__/channel_subscription_filter.test.tsx.snap
```diff
@@ -50,6 +50,34 @@ exports[`components/ChannelSubscriptionFilter should match snapshot 1`] = `
...
                 "value": "versions",
               },
+              Object {
+                "label": "Comment Visibility",
+                "value": "commentVisibility",
+              },
+              Object {
+                "label": "Comment Visibility",
+                "value": "commentVisibility",
+              },
```
mickmister, line 60: Why does this show up so many times?

webapp/src/components/modals/channel_subscriptions/channel_subscription_filter.tsx
```diff
@@ -269,7 +270,16 @@ export default class ChannelSubscriptionFilter extends React.PureComponent<Props
...
 
         let valueSelector;
-        if (isEpicLinkField(this.props.field)) {
+        if (isCommentVisibilityField(field)) {
+            valueSelector = (
+                <JiraCommentVisibilitySelector
+                    {...selectProps}
+                    fieldName={field.name}
+                    value={value.values}
+                    onChange={this.handleEpicLinkChange}
```
mickmister, line 279: This should be a different function than `handleEpicLinkChange` right? Looks like that function isn't actually specific to epic links though, and does what we want here too, and it's being used in other places. Maybe we can change it to something like `handleValueChangeWithoutName`

webapp/src/utils/jira_issue_metadata.tsx
```diff
@@ -212,7 +214,19 @@ export function getCustomFieldFiltersForProjects(metadata: IssueMetadata | null,
...
+            name: 'Comment Visibility',
+            schema: {
+                type: 'commentVisibility',
+            },
+            values: [],
+            issueTypes: field.validIssueTypes,
+        } as FilterField;
+    });
+
+    const result = populatedFields.concat(userDefinedFields, commentVisibility);
```
mickmister, line 229: I think this `selectFields.map` is why we see several "Comment Security" objects in the jest snapshot

Will this work:

```ts
const issueTypes: IssueTypeIdentifier[] = metadata.projects[0].issuetypes.map(({id, name}) => ({id, name}));

const commentVisibility = {
    key: commentVisibilityFieldKey,
    name: 'Comment Visibility',
    schema: {
        type: 'commentVisibility',
    },
    values: [],
    issueTypes,
} as FilterField;
```

webapp/src/components/data_selectors/jira_commentvisibility_selector/jira_commentvisibility_selector.tsx
```diff
@@ -0,0 +1,54 @@
...
+const JiraCommentVisibilitySelector = (props: Props) => {
+    const {value, isMulti, instanceID, searchCommentVisibilityFields} = props;
+    const fetchInitialSelectedValues = async (): Promise<ReactSelectOption[]> =>
+        ((!value || (isMulti && !value.length)) ? [] : commentVisibilityFields(''));
+
+    const commentVisibilityFields = async (inputValue: string): Promise<ReactSelectOption[]> => {
+        const params = {
+            fieldValue: inputValue,
+            instance_id: instanceID,
+            expand: 'groups',
```
mickmister, line 35: If this is always set to `groups` for this endpoint, then it seems like we can omit `expand` from the API request, and have the server just always use that value when fetching the visibilities from Jira. Otherwise we unnecessarily allow arbitrary expands for the endpoint.

webapp/src/components/data_selectors/jira_commentvisibility_selector/jira_commentvisibility_selector.tsx
```diff
@@ -0,0 +1,54 @@
...
+    searchCommentVisibilityFields: (params: {fieldValue: string}) => (
+        Promise<{data: {groups: {items: {name: string}[]}}; error?: Error}>
+    );
+    fieldName: string;
+};
+
+const JiraCommentVisibilitySelector = (props: Props) => {
+    const {value, isMulti, instanceID, searchCommentVisibilityFields} = props;
+    const fetchInitialSelectedValues = async (): Promise<ReactSelectOption[]> =>
+        ((!value || (isMulti && !value.length)) ? [] : commentVisibilityFields(''));
```
mickmister, line 29: This is a little hard to read, partially because it's fairly terse one-liner that spans to two lines, and it also references a variable defined below, which is a bit confusing

mickmister (commented):
Great work on this improvement @Nityanand13 :+1:

I have a few topics for discussion, one in particular is how to properly handle the case when the Jira comment author is not connected to Mattermost. This poses a challenge with the tradeoffs of defaulting to secure or not. There are some other suggestions as well, but mostly LGTM

mickmister:
Also this will need to be tested on Jira Cloud and Jira Server

server/webhook_worker.go
esarafianou: @mickmister Seems I had missed this. Is this still relevant?

server/client.go
raghavaggarwal2308: @mickmister We are getting the user `groups` here. I don't think the project roles API returns that response

server/webhook_worker.go
raghavaggarwal2308: @mickmister I have removed the extra API call for getting the comment visibility data as in the current implementation of Jira plugin we are always making an API call to get the comment event data. So, the user have to be connected in every case of comment events.

This will be fixed with the API token approach we discussed earlier. So, maybe we can proeed with this for now and it will be fixed later.

raghavaggarwal2308:
@mickmister Fixed the review comments added by you.

wiggin77:
@raghavaggarwal2308 can you shepherd this through the outstanding comments?

server/webhook_worker.go
raghavaggarwal2308: API token approch implemented via https://github.com/mattermost/mattermost-plugin-jira/pull/1102

raghavaggarwal2308:
@wiggin77 There are two unresolved comments above:
1. https://github.com/mattermost/mattermost-plugin-jira/pull/894#discussion_r1357668970: This has been resolved now as the [API token approach](https://github.com/mattermost/mattermost-plugin-jira/pull/1102) is implemented now. Basically the issue was that when we receive a comment event we needed to fetch some more comment details using the author ID of the comment but if the author is not connected to Jira from MM then we will not have the auth token and wont be able to fetch the details and end us not posting the event in the channel at all.

2. https://github.com/mattermost/mattermost-plugin-jira/pull/894#discussion_r1357670721: Michael was suggesting to use some alternate API, which did not seem to be valid for our use case.

raghavaggarwal2308:
@esarafianou Gentle reminder for review

esarafianou:
@raghavaggarwal2308 I hope I can get to this PR this week. Apologies for the delay.

wiggin77 (approved):
LGTM :+1:

wiggin77:
@raghavaggarwal2308 @AayushChaudhary0001 what is the QA status on this?  QA should be completed before security review.

raghavaggarwal2308:
@wiggin77 We will get it done as soon as the PR is approved from @esarafianou . So that if there are an review fixes, they are covered in the QA testing

arush-vashishtha:
Hey @Kshitij-Katiyar, While testing this PR, I found an issue related to the "issue events" in the Jira subscription. If a user creates a subscription for issues and applies the filter on "comment visibility," no events related to issues are received in the subscribed channel.

AayushChaudhary0001 (approved):
The above issue has been resolved and the code changes are working fine, LGTM. Approved.

server/client.go
```diff
@@ -266,6 +281,17 @@ func (client JiraClient) SearchAutoCompleteFields(params map[string]string) (*Au
...
+// SearchCommentVisibilityFields searches fieldValue specified in the params and returns the comment visibility suggestions
+// for that fieldValue
+func (client JiraClient) SearchCommentVisibilityFields(params map[string]string) (*CommentVisibilityResult, error) {
+	result := &CommentVisibilityResult{}
+	if err := client.RESTGet(commentVisibilityRoute, params, result); err != nil {
+		return nil, err
+	}
+	result.Groups.JiraUserGroups = append(result.Groups.JiraUserGroups, &JiraUserGroup{visibleToAllUsers})
+	return result, nil
+}
```
esarafianou, line 293: Correct me if i'm wrong but this doesn't look complete. 

The /2/user endpoint returns both groups and applicationRoles, but we're only using the groups information.  Comment Visibility has both a `Value` and a `Type`. The type can be groups or roles. Seems like we don't do anything for the roles, neither utilize the `Type` of CommentVisibility: https://github.com/andygrunwald/go-jira/blob/7257f0349598537d526d26f937bf8847c50258c9/cloud/issue.go#L506C4-L511

server/issue.go
```diff
@@ -407,6 +412,47 @@ func (p *Plugin) GetCreateIssueMetadataForProjects(instanceID, mattermostUserID
...
+
+	instanceID := r.FormValue(instanceIDQueryParam)
+	client, _, connection, err := p.getClient(types.ID(instanceID), types.ID(mattermostUserID))
+	if err != nil {
+		return http.StatusInternalServerError, err
+	}
+
+	params := map[string]string{
+		"fieldValue": r.FormValue(fieldValueQueryParam),
+		"expand":     expandValueGroups,
```
esarafianou, line 433: Here's where we should also expand `applicationRoles`

esarafianou (changes requested):
First round of review

server/client.go
Kshitij-Katiyar: This PR aims to provide filter support upon `Groups` for comment visibility. 

We can use the `Application Role` in the future to provide filter support on that too, but we think it should be done in a new PR because it would be quite a task. I will let @wiggin77 decide on this.

I am not sure what code be done with the `type` of group in this PR. Can you please elaborate? @esarafianou

server/subscribe.go
```diff
@@ -170,6 +171,18 @@ func (p *Plugin) matchesSubsciptionFilters(wh *webhook, filters SubscriptionFilt
 		}
 
 		value := getIssueFieldValue(issue, field.Key)
+		if field.Key == CommentVisibility {
+			if wh.Comment.Visibility.Value != "" {
```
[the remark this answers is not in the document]
Kshitij-Katiyar: @esarafianou Can you please explain why you think we should use `FilterIncludeAll` for comment visibility? And where are you suggesting to use it? 

Are you suggesting adding the `FilterIncludeAll` check under the condition below?
```
if field.Key == CommentVisibility {
			if wh.Comment.Visibility.Value != "" {
```

server/issue.go
Kshitij-Katiyar: https://github.com/mattermost/mattermost-plugin-jira/pull/894#discussion_r1967568194

server/subscribe.go
```diff
@@ -170,6 +171,18 @@ func (p *Plugin) matchesSubsciptionFilters(wh *webhook, filters SubscriptionFilt
...
+				value = value.Add(wh.Comment.Visibility.Value)
+			} else {
+				value = value.Add(visibleToAllUsers)
+			}
+		}
+
+		if !wh.eventTypes[commentCreated] {
```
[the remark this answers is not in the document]
Kshitij-Katiyar: Great catch, Added to the code

esarafianou:
@Kshitij-Katiyar please resolve conflicts before I review. Seems that something odd happened when merging the master branch and now the PR has 5K LoC additions.

server/client.go
esarafianou: This means that with this PR, users with an Application role that can see a comment in Jira won't receive a subscription notification. This is an unexpected functionality limitation. 
Tagging @wiggin77 to chime in here. My comment is not security related, but for usability and customer expectations.

esarafianou: Regarding the `type`: 

From my understanding of the Jira API, the `type` field represents `role` for application roles and `groups` for groups. 

Currently, since we only fetch` groups`, the `type` field may seem unnecessary. However, 1) I believe we should include application roles as well (see my earlier comment), and 2) I strongly recommend adding a `type` check now, even if we don’t immediately incorporate `applicationRoles`. This would make the implementation more robust and future-proof. 

The scenario I’m considering involves a potential overlap between a group and an application role having the same name. Without verifying that the `type` matches `groups` during the comparison of user groups to visibility settings, we could mistakenly provide access to an application role named `nameA` instead of the intended group named `nameA`.

esarafianou (changes requested):
It's just the `Type` that I'd like checked and equal to `groups`. Otherwise, it looks good!

arush-vashishtha:
This PR has been tested for the following scenarios considering the include/exclude functionality of the filter;
1. The events are received in the Mattermost subscribed channel by creating a comment in Jira for the included group in the filter. This scenario is working fine and events are received in the channel.
2. The events are received in the Mattermost subscribed channel upon creating a comment in Jira for one of the included groups in the filter. This scenario is working fine and events are received in the channel.
3. The events are received in the Mattermost subscribed channel upon creating a comment in Jira for `visible-to-all`. This scenario is working fine and events are received in the channel.
4. The events are received in the Mattermost subscribed channel for other actions than comments in Jira when the subscription has a filter on comment visibility to include specific groups. This scenario is working fine and events are received in the channel.
5. No events are received in the Mattermost subscribed channel upon creating a comment in Jira for the excluded group in the filter. This scenario is working fine and no events are received in the channel.
6. No events are received in the Mattermost subscribed channel upon creating a comment in Jira for any one of the excluded groups in the filter. This scenario is working fine and no events are received in the channel.
7. No events are received in the Mattermost subscribed channel upon creating a comment in Jira for `visible-to-all`. This scenario is working fine and no events are received in the channel.
8. The events are received in the Mattermost subscribed channel for other actions than comments in Jira when the subscription has a filter on comment visibility to exclude specific groups. This scenario is not working fine and the events are not received for other actions such as `Creating issue`, `Updating issue`, etc;

However, when using the filter with the inclusions set to "Empty," no event is received because general comments by default contain the "visible-to-all" field in the comment visibility. Additionally, the "Include or Empty" inclusions for the filter function are similar to the "Include".
cc @Kshitij-Katiyar

server/client.go
Kshitij-Katiyar: @esarafianou Logged an issue to provide support for role-based filtering 
https://github.com/mattermost/mattermost-plugin-jira/issues/1178

Kshitij-Katiyar: @esarafianou I got your point, I have added the check, you can see that in the attached screenshot
Screenshot from 2025-02-27 15-01-05

server/subscribe.go
```diff
@@ -171,6 +172,18 @@ func (p *Plugin) matchesSubsciptionFilters(wh *webhook, filters SubscriptionFilt
...
 		value := getIssueFieldValue(issue, field.Key)
+		if field.Key == CommentVisibility {
+			if wh.Comment.Visibility.Value != "" {
+				value = value.Add(wh.Comment.Visibility.Value)
+			} else {
+				value = value.Add(visibleToAllUsers)
+			}
+		}
+
+		if !(wh.eventTypes[commentCreated] || wh.eventTypes[commentUpdated]) && field.Inclusion != FilterIncludeAll {
```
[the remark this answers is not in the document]
Kshitij-Katiyar: No, it is to make sure that function `isValidFieldInclusion` works fine for `FilterIncludeAll`.
It is done so because the condition for `FilterIncludeAll` is:
```
containsAll := value.ContainsAll(field.Values.Elems()...)
if inclusion == FilterIncludeAll && !containsAll {
   return false
}
```
As we are adding `visibleToAllUsers` into `field.Values` but `value` variable remains unchanged so it will fail the check for `FilterIncudeAll`

server/client.go
esarafianou: Looks good for now. When we add role-based filtering, we'll need to make some modifications but we can explore it then.

esarafianou (approved):

PR #897, opened 2022-12-13
[GH-604]: Feature to create a subscription template for Jira subscriptions.

Kshitij-Katiyar:
**Summary** 

- This PR added the feature to create subscription templates and use them to create Jira subscriptions. So if a Mattermost user needs to create each channel subscription from scratch. This can be cumbersome, especially when there is plenty of duplicated settings across subscriptions. Now, we can instead have a subscription template feature that allows us to create a shell for a subscription, which can then be selected when creating a new subscription.
- Each template is associated with a given Jira project. This is because each project's issue schemas can be a bit different, which is why we have associated subscriptions to a single project.
- This PR contains all the changes from [Pr #783](https://github.com/mattermost/mattermost-plugin-jira/pull/783) So this PR alone is enough to fix this issue so we can close [Pr #783](https://github.com/mattermost/mattermost-plugin-jira/pull/783) afterward.

**Issue** 

- Fixes https://github.com/mattermost/mattermost-plugin-jira/issues/604
- [Pr #783](https://github.com/mattermost/mattermost-plugin-jira/pull/783)


hanzei:
Let's put this in the icebox as it will conflict with other work

mickmister:
@Kshitij-Katiyar Can you please provide some screenshots of this PR in action?

Kshitij-Katiyar:
@mickmister Some screenshots of the feature 
Screenshot from 2023-09-13 17-46-57
Screenshot from 2023-09-13 17-48-43
Screenshot from 2023-09-13 17-49-01
Screenshot from 2023-09-13 17-49-15
Screenshot from 2023-09-13 17-49-29
Screenshot from 2023-09-13 17-49-35

server/subscribe.go
```diff
@@ -92,13 +107,39 @@ type Subscriptions struct {
 	Channel       *ChannelSubscriptions
 }
 
+type ListSubscriptionTemplate map[string]*SubscriptionTemplate
```
mickmister, line 110: This is a map and not a list. Maybe name this `SubscriptionTemplateCollection`, or `SubscriptionTemplatesByID` since I think ids are what the keys are

server/subscribe.go
```diff
@@ -55,6 +62,14 @@ type ChannelSubscription struct {
 	InstanceID types.ID            `json:"instance_id"`
 }
 
+type SubscriptionTemplate struct {
+	ID         string               `json:"id"`
+	ChannelID  string               `json:"channel_id"`
```
mickmister, line 67: I'm thinking subscription templates should be just per-project, and not per-channel. We want users to be able to use the templates across different channels. We'll also only want to expose templates for projects that the user has access to in Jira. Maybe we can lump in the templates in one of the existing API endpoints related to fetching subscriptions, to avoid duplicate project checking

mickmister (commented):
Hi @Kshitij-Katiyar, apologies for the late comment here. I did a quick review over the PR and left a few comments. I can give a more in-depth review once they can be addressed

mickmister:
Thanks for the work on this btw. This will have a big impact on the workflow of using subscriptions across a team

server/subscribe.go
Kshitij-Katiyar: @mickmister Sorry I didn't get you, can you explain it a bit. 
Currently we are storing the subscription template with the project Key only. Here is the code related to the [addSubscriptionTemplate](https://github.com/Brightscout/mattermost-plugin-jira/blob/eec7f78830a9313d5f06191fc1b2e477375977fe/server/subscribe.go#L403)

mickmister: @Kshitij-Katiyar What I'm not understanding is why does the template have a `ChannelID` property? The template should be agnostic to which channel it will be used in

raghavaggarwal2308: @mickmister @wiggin77 The channel ID basically used to post a ephemeral in the channel when a template is added/updated/deleted. 
The template is created per-project only and the channel ID has not effect in it.

fmartingr: Should that message be posted in all channels where the template is being used?

raghavaggarwal2308: @fmartingr It's an ephemeral message so it is visible only to the user who is updating the template. So, I don't think it should be posted in all the channels.

Kshitij-Katiyar:
@wiggin77 I have synced the branch with the master

webapp/src/components/modals/channel_subscriptions/select_channel_subscription.tsx
```diff
@@ -176,6 +252,23 @@ export default class SelectChannelSubscriptionInternal extends React.PureCompone
...
+                    <tbody>
+                        {subscriptionTemplates.map((element) => (
+                            this.renderRow(element, true)
+                        ))}
+                    </tbody>
+                </table>
+            );
+        } else {
+            subscriptionTemplateRows = (
+                <p>{'Click "Create Template" to create subscription templates.'}</p>
```
wiggin77, line 268:
```suggestion
                <p>{'Click "Create Template" to create a subscription template.'}</p>
```

wiggin77 (approved):

avas27JTG (approved):

abbas-dependable-naqvi:
@AayushChaudhary0001 Please re test this PR

PR #921, opened 2023-03-07
MM-50981 Remove usage of model.AppError 

srkgupta:
#### Summary
Remove usage of model.AppError in JIRA plugin

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-50981


m1lt0n (approved):
Great changes and cleanup of the deprecated built-in go APIs too!

hanzei:
@srkgupta Heads up that there is a merge conflict.

server/issue.go
```diff
@@ -338,16 +337,16 @@ func (p *Plugin) CreateIssue(in *InCreateIssue) (*jira.Issue, error) {
...
+	if err != nil {
+		return nil, errors.WithMessage(err, "failed to create notification post "+in.PostID)
 	}
 
 	if post != nil && len(post.FileIds) > 0 {
 		go func() {
 			conf := instance.Common().getConfig()
 			for _, fileID := range post.FileIds {
-				mattermostName, _, _, e := client.AddAttachment(p.API, created.ID, fileID, conf.maxAttachmentSize)
+				mattermostName, _, _, e := client.AddAttachment(*p.client, created.ID, fileID, conf.maxAttachmentSize)
```
hanzei, line 349: While we are here: Can we rename the `error` to the standard `err`?

server/kv.go
```diff
@@ -104,19 +106,10 @@ func (store store) get(key string, v interface{}) (returnErr error) {
...
-	data, appErr := store.plugin.API.KVGet(key)
-	if appErr != nil {
-		return appErr
-	}
-	if data == nil {
-		return kvstore.ErrNotFound
-	}
-
-	err := json.Unmarshal(data, v)
+	err := store.plugin.client.KV.Get(key, &v)
```
hanzei, line 109: Previously, `v` was passed to `json.Unmarshal` meaning it had to be an pointer. Why does a pointer to `v` get passed now?

server/kv.go
```diff
@@ -393,25 +381,26 @@ func (store store) EnsureRSAKey() (rsaKey *rsa.PrivateKey, returnErr error) {
...
 	}
 	return nil
 }
 
 func (store store) LoadOneTimeSecret(key string) (string, error) {
-	b, appErr := store.plugin.API.KVGet(hashkey(prefixOneTimeSecret, key))
-	if appErr != nil {
-		return "", errors.WithMessage(appErr, "failed to load one-time secret "+key)
+	var secret string
+	err := store.get(hashkey(prefixOneTimeSecret, key), secret)
```
hanzei, line 394: That looks suspicious. The stored data is not in JSON format, but `store.get` will try to unmarshal it. Are you sure this line works?

server/kv.go
```diff
@@ -483,25 +463,18 @@ func (store *store) LoadInstance(instanceID types.ID) (Instance, error) {
...
 	si := serverInstance{}
-	err := json.Unmarshal(data, &si)
+	err := store.plugin.client.KV.Get(fullkey, &si)
 	if err != nil {
 		return nil, err
 	}
-
+	if si.InstanceCommon == nil {
+		return nil, nil
+	}
```
hanzei, line 473: Previously, the code returned an error in this case. Why was it removed?

server/plugin.go
```diff
@@ -261,18 +261,18 @@ func (p *Plugin) OnActivate() error {
...
+	profileImage, err := os.Open(filepath.Join(bundlePath, "assets", "profile.png"))
 	if err != nil {
 		return errors.Wrap(err, "couldn't read profile image")
 	}
 
-	if appErr := p.API.SetProfileImage(botUserID, profileImage); appErr != nil {
-		return errors.Wrap(appErr, "couldn't set profile image")
+	if err = p.client.User.SetProfileImage(botUserID, profileImage); err != nil {
+		return errors.Wrap(err, "couldn't set profile image")
 	}
```
hanzei, line 277: While we are here, let's make use of https://pkg.go.dev/github.com/mattermost/mattermost-plugin-api@v0.1.2#ProfileImagePath.

go.mod
```diff
@@ -9,13 +9,89 @@ require (
 	github.com/dgrijalva/jwt-go v3.2.0+incompatible
 	github.com/gorilla/mux v1.8.0
 	github.com/jarcoal/httpmock v1.0.8
-	github.com/mattermost/mattermost-plugin-api v0.0.27
+	github.com/mattermost/mattermost-plugin-api v0.1.2-0.20230302185923-1714af2b0110
```
hanzei, line 12: Why not `github.com/mattermost/mattermost-plugin-api v0.1.2`?

server/plugin_test.go
```diff
@@ -91,9 +91,9 @@ func TestPlugin(t *testing.T) {
 		},
 		"CreatePostError": {
 			Configuration:      validConfiguration,
-			CreatePostError:    model.NewAppError("foo", "bar", nil, "", http.StatusInternalServerError),
+			CreatePostError:    model.NewAppError("foo", "bar", nil, "", http.StatusBadRequest),
 			Request:            httptest.NewRequest("POST", "/webhook?team=theteam&channel=thechannel&secret=thesecret", validRequestBody()),
-			ExpectedStatusCode: http.StatusInternalServerError,
+			ExpectedStatusCode: http.StatusBadRequest,
```
hanzei, line 96: Why was there a behaviour change?

server/subscribe.go
```diff
@@ -647,50 +646,10 @@ func (p *Plugin) hasPermissionToManageSubscription(instanceID types.ID, userID,
...
-		}
-
-		currentAttempt++
-		if currentAttempt >= retryLimit {
-			return errors.New("reached write attempt limit")
-		}
-
-		time.Sleep(retryWait)
+func (p *Plugin) atomicModify(key string, modify func(initialValue []byte) (interface{}, error)) error {
+	err := p.client.KV.SetAtomicWithRetries(key, modify)
```
hanzei, line 650: Is `p.atomicModify` need any longer? Can we use `p.client.KV.SetAtomicWithRetries`?

hanzei (changes requested):
Thanks for taking over this huge effort :+1: 

1. `plugin_store.go` was left untouched. Why is that the case? Can the code in there also get migrated?
2. Let's double-check that the code migration around the KV store is correct. There are some edge cases we need to be aware of. More details in the comments below :arrow_down:

srkgupta:
Thanks for the detailed review @hanzei. Will be making the suggested changes shortly and request your re-review once its committed.

server/kv.go
srkgupta: Nice catch. Yes, this looks like the code was accidentally over-written. It should read:

```suggestions
err := store.plugin.client.KV.Get(hashkey(prefixOneTimeSecret, key), secret)
```

I will push a commit to fix this.

server/plugin_test.go
srkgupta: Earlier, we were returning the status Code from the appError. However, since the error no longer has access to the returned Status Code, I had hardcoded it as `http.StatusBadRequest` which had caused this change. I will revert and hardcode the return status as `http.StatusInternalServerError` instead.

Screenshot 2023-03-24 at 10 45 21 AM

server/issue.go
```diff
@@ -672,7 +671,7 @@ func (p *Plugin) AttachCommentToIssue(in *InAttachCommentToIssue) (*jira.Comment
 		conf := instance.Common().getConfig()
 		extraText := ""
 		for _, fileID := range post.FileIds {
-			mattermostName, jiraName, mime, e := client.AddAttachment(p.API, in.IssueKey, fileID, conf.maxAttachmentSize)
+			mattermostName, jiraName, mime, e := client.AddAttachment(*p.client, in.IssueKey, fileID, conf.maxAttachmentSize)
```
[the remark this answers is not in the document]
srkgupta: Go lint error `server/issue.go:674:36: shadow: declaration of "err" shadows declaration at line 633 (govet)` is thrown in this case, since `err` is already declared at line 633. I will have to let it be as is.

mickmister (approved):
No issues found other than Ben's comments above :+1:

server/command.go
```diff
@@ -370,7 +370,7 @@ func (p *Plugin) help(args *model.CommandArgs) *model.CommandResponse {
 	return &model.CommandResponse{}
 }
 
-func (p *Plugin) ExecuteCommand(c *plugin.Context, commandArgs *model.CommandArgs) (*model.CommandResponse, *model.AppError) {
+func (p *Plugin) ExecuteCommand(c *plugin.Context, commandArgs *model.CommandArgs) (*model.CommandResponse, error) {
```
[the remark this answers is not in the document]
srkgupta: Thanks for catching this @hanzei. I thought we supported the return type as `error` based on the server changes here:
https://github.com/mattermost/mattermost-server/blob/master/model/client4.go#L6463

But the slash commands were not working. So I have reverted this change and pushed a new commit.

server/kv.go
```diff
@@ -420,31 +409,22 @@ func (store store) StoreOauth1aTemporaryCredentials(mmUserID string, credentials
...
 func (store store) OneTimeLoadOauth1aTemporaryCredentials(mmUserID string) (*OAuth1aTemporaryCredentials, error) {
-	b, appErr := store.plugin.API.KVGet(hashkey(prefixOneTimeSecret, mmUserID))
-	if appErr != nil {
-		return nil, errors.WithMessage(appErr, "failed to load temporary credentials for "+mmUserID)
-	}
-	// If the key expired, appErr is nil, but the data is also nil
-	if len(b) == 0 {
-		return nil, errors.Wrapf(kvstore.ErrNotFound, "temporary credentials for %s not found or expired, try to connect again"+mmUserID)
-	}
-
```
hanzei, line 439: Why was this case removed?

server/kv.go
```diff
@@ -483,25 +463,18 @@ func (store *store) LoadInstance(instanceID types.ID) (Instance, error) {
...
 	}
-
+	if si.InstanceCommon == nil {
+		return nil, errors.Wrap(kvstore.ErrNotFound, fullkey)
+	}
 	switch si.Type {
 	case CloudInstanceType:
 		ci := cloudInstance{}
-		err = json.Unmarshal(data, &ci)
+		err := store.plugin.client.KV.Get(fullkey, &ci)
```
hanzei, line 477: Given that this is a kind of custom JSON unmarshal case, I would also be fine with simply getting a byte slice using `store.plugin.client.KV.Get` and keeping the rest of the code as it is.

hanzei (approved):
Two nits, but the rest LGTM :+1:

server/kv.go
srkgupta: Thanks @hanzei. Reverted to use existing custom JSON unmarshal logic while using `store.plugin.client.KV.Get`

hanzei (approved):
:rocket:

hanzei:
@srkgupta Heads up that there are merge conflicts to resolve

DHaussermann (commented):
Smoke tested Jira

For the most part I saw no issue. 

One problem I see on this branch is that authenticating to Jira server seems to no longer work.  (no issues with Cloud authentication)
Steps:
- Install a Jira server instance (mine is 9.5.0)
- Run the `/jira connect` command
- In the new tab click allow
image

- On the oAuth complete page I get a 405 response
image
-The ngrok tunnel to Jira shows nothing after the 302 (the other 2 calls in the screenshot happened later and are unrelated)
image


@srkgupta unfortunately, because of a licensing issue on Jira server I was still working to resolve Friday, I did not confirm this was working on https://github.com/mattermost/mattermost-plugin-jira/pull/923 and deffered testing until now so, I am unsure of which PR introduces the behavior.

srkgupta:
Thanks @DHaussermann. I have fixed the issue now. This was again one of those cases, where the HTTP method wasn't clearly defined previously in the OAuth1 Complete endpoint. Please use the latest code and try again.

DHaussermann (approved):
Tested and passes
I have spent some time smoke testing and see no more regressions
Tested against both Cloud and Serever
- Setup 
- Authentication flow (browser and desktop)
- Issue creation and attachment
- Subscription and delivery
- Multiple instances connected
- Tested various commands for assign, view and transition etc..

LGTM!

Thanks @srkgupta!

PR #928, opened 2023-03-30
MM-51310 Implement OAuth2 Authentication

srkgupta:
#### Summary
This PR implements the OAuth2 authentication in the Jira Plugin. 

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-51310

#### Screenshots

**Install OAuth2 using setup flow**
Whenever the system admin runs the command `/jira setup`, the setup flow to configure OAuth2 client would look something like this:

Screenshot 2023-04-25 at 12 23 20 PM

Screenshot 2023-04-25 at 12 23 39 PM

Screenshot 2023-04-17 at 4 33 06 PM

Screenshot 2023-04-17 at 4 34 06 PM


m1lt0n (changes requested):
Nice one @srkgupta ! One comment only: I suggest having telemetry for the cloud-oauth flow in the same manner we have it for the rest of the setup steps in `setup_flow`

srkgupta:
Thanks @m1lt0n. Added the telemetry events now in setup_flow for OAuth2 events.

server/setup_flow.go
```diff
@@ -311,6 +325,67 @@ func (p *Plugin) stepCloudUploadApp() flow.Step {
...
+	return flow.NewStep(stepCloudOAuthConfigure).
+		WithPretext("##### :white_check_mark: Step 2(a): Register an OAuth 2.0 Application in Jira").
+		WithText(fmt.Sprintf("Complete the following steps, then come back here to select **Configure**.\n\n"+
+			"1. Follow [these instructions](https://developer.atlassian.com/cloud/confluence/oauth-2-3lo-apps/#enabling-oauth-2-0--3lo-) to register an OAuth 2.0 application in Jira.\n"+
+			"2. Set the following values:\n"+
+			"	- Name: `Mattermost Jira Plugin - <your company name>`\n"+
+			"3. Select **Permissions** in the left menu. Next to the JIRA API, select **Add**\n"+
+			"4. Then select **Configure** and ensure following scopes are selected:\n"+
+			"   - Scopes: `%s`\n"+
+			"3. Copy the **Client ID** and **Secret** from the registered 0Auth Application's **Settings** page and keep it handy.\n", JiraScopes)).
```
mickmister, line 338:
```suggestion
			"5. Copy the **Client ID** and **Secret** from the registered 0Auth Application's **Settings** page and keep it handy.\n", JiraScopes)).
```

You can also put `1.` on each of them, and the markdown renderer will automatically number them:

```
1. Item
1. Item
1. Item
```

1. Item
1. Item
1. Item

server/instance_cloud_oauth.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+type JiraAccessibleResources []struct {
+	ID string
+}
+
+var _ Instance = (*cloudOAuthInstance)(nil)
+
+const (
+	JiraScopes       = "read:jira-user,read:jira-work,write:jira-work"
```
mickmister, line 44: What exactly is `jira-work`, and does this allow the user to create issues and comments? We'll also want to be able to assert that the user has "browse project" permission for a given project, to assert the user has permissions to read issues from a given project.

server/instance_cloud_oauth.go
```diff
@@ -0,0 +1,186 @@
...
+	contents, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		return "", fmt.Errorf("failed read accesible resources response: %s", err.Error())
+	}
+
+	var resources JiraAccessibleResources
+	err = json.Unmarshal(contents, &resources)
+
+	if err != nil {
+		return "", fmt.Errorf("failed marshall: %s", err.Error())
```
mickmister, line 177:
```suggestion
		return "", errors.Wrap(err, "failed to unmarshal JiraAccessibleResources")
```

server/instance_cloud_oauth.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	// We return the first resource ID only
+	if len(resources) < 1 {
+		return "", errors.New("No resources available for this Jira Cloud Account")
```
mickmister, line 182:
```suggestion
		return "", errors.New("No resources are available for this Jira Cloud Account.")
```

mickmister (changes requested):
Awesome work on this @srkgupta :+1: I have just a few requests and comments on this

server/setup_flow.go
```diff
@@ -197,6 +205,12 @@ func (p *Plugin) stepChooseEdition() flow.Step {
 			},
 			OnDialogSubmit: p.submitCreateCloudInstance,
 		}).
+		WithButton(
+			flow.Button{
+				Name:    "Jira Cloud (OAuth 2.0)",
+				Color:   flow.ColorPrimary,
+				OnClick: flow.Goto(stepCloudOAuthConfigure),
+			}).
```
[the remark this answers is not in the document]
srkgupta: When a user clicked on the button to configure `Cloud OAuth 2.0`, the steps were slightly different, i.e. we needed to display the instructions on how to register an OAuth 2.0 app in JIRA and then provide the `Dialog` with fields like `Jira cloud organization`,  `Client ID` and `Client Secret`. Hence this flow was slightly different from rest of the other implementations. Anyways, I have now changed the flow and kept it in consistency with the rest of the flows.

server/setup_flow.go
```diff
@@ -311,6 +325,67 @@ func (p *Plugin) stepCloudUploadApp() flow.Step {
 		})
 }
 
+func (p *Plugin) stepCloudOAuthConfigure() flow.Step {
+	return flow.NewStep(stepCloudOAuthConfigure).
+		WithPretext("##### :white_check_mark: Step 2(a): Register an OAuth 2.0 Application in Jira").
+		WithText(fmt.Sprintf("Complete the following steps, then come back here to select **Configure**.\n\n"+
+			"1. Follow [these instructions](https://developer.atlassian.com/cloud/confluence/oauth-2-3lo-apps/#enabling-oauth-2-0--3lo-) to register an OAuth 2.0 application in Jira.\n"+
```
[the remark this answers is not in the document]
srkgupta: I have added this as a separate task, so we dont forget this in future and can handle the required changes:
https://mattermost.atlassian.net/browse/MM-51913

server/instance_cloud_oauth.go
srkgupta: Yes `jira-work` permission allows users to view/create issues and comments. 

Screenshot 2023-04-04 at 3 36 19 PM

There's no specific permission for `browse project`. The closest would be `manage project`, but that does not seems appropriate as per what we want to achieve.

mickmister: Yes it seems `read:jira-work` satisfies the issue based on your screenshot :+1:

server/setup_flow.go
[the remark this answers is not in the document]
mickmister: Can we reuse this dialog in replacement of the webapp component added for this? Then we can have one source of truth implementation, and avoid adding code to the webapp. We generally favor server-driven functionality when possible.

srkgupta:
> Great work on this @srkgupta 👍
> 
> I have one remaining topic of potentially removing the webapp component in favor of the interactive dialog that is implemented here

Thanks @mickmister. I have removed the webapp components and this is now ready for your review 🎉

DHaussermann:
Hi @srkgupta I have done a round of testing on this. For now I did a pass of "happy path" testing to ensure the solution works and I can connect using oAuth. 
It does work without turning on developer mode AND I works in a browser with 3rd party cookies disabled 🎉 

At this point I only have some non technical feedback that would make the setup process a bit smoother. 
1. Why is oAuth 2.0 a separate option? Do we still anticipate some people will use the older authentication method?
image

2. Step 1. was a bit confusing. Can you save me the effort of figuring out their UI to get to the console and just please include this link `https://developer.atlassian.com/console/myapps`  
ex: 
Follow [these instructions](https://developer.atlassian.com/cloud/confluence/oauth-2-3lo-apps/#enabling-oauth-2-0--3lo-) to register an OAuth 2.0 application in Jira from the [Developer console](https://developer.atlassian.com/console/myapps)

3. Maybe there should be a step "2.5" telling me I'm creating an OAuth 2.0 with this image 
image

4. For the call back URL There is a leading space in front  of the `h` which causes user error when they paste.
image

5. You'll need to click **Add** here.   Not **Configure**

image

6. Maybe there should be a step 10. explicitly telling me to click configure and save the values?

7. 0/5 - Can we change the filed type here so that cashed values don't show? (this may not be acceptable from a security perspective but I don't think we eveer had a better solution for other integrations with user side input) 
image

mickmister:
> Why is oAuth 2.0 a separate option? Do we still anticipate some people will use the older authentication method?

@DHaussermann I agree we should probably remove the old option from the setup flow. Maybe keep it in the `instance install` command and label it there as deprecated.

> 0/5 - Can we change the filed type here so that cashed values don't show? 

Dylan is recommending we change the `client_secret` field to `text`, so the browser doesn't think we're interacting with a login form with username/password. We've tried to fix this in the past in the webapp but the fix doesn't seem to work properly.

---

Some other issues I noticed while writing tests:

1. When executing `/jira instance install cloud-oauth https://mmtest.atlassian.net`, the modal didn't use the entered url when it opened

2. `/jira setup` says this message even if I'm already in the bot DM channel:

> continue in the direct conversation with @jira bot.

3. OAuth complete page does not automatically close. I fixed this in the e2e PR https://github.com/mattermost/mattermost-plugin-jira/pull/934

4. Not sure how this happened, but at some point the setup flow assumed a Jira URL that I used for a previous usage of the setup flow.

5. Since plugin doesn't allow localhost as siteurl, we will need to support a special case for CI to use that.

6. The webhook secret is not set in the webhook URL shown in the modal. Maybe this doesn't happen every time. Maybe the secret was not set somehow?

7. Sometimes the plugin thinks I'm trying to install 2 Jira instances when entering the client id & secret, even though it's the same Jira URL I entered before. Maybe it's `/jira instance install` that messes this up by queueing the instance in the kv store.

srkgupta:
Thanks for your feedback @DHaussermann and @mickmister. I will fix these issues shortly. 

As of now, I have fixed one of the major issues highlighted by @mickmister, i.e. the logic to refresh the OAuth token automatically. FYI @DHaussermann, currently Jira expires the connection token after 1 hour. Hence if a user connects to Jira using OAuth2, their connection would not work beyond 1 hour interval. However this issue is fixed now and you should no longer see any errors related to this. You can confirm this by connecting your Jira account once and randomly issuing any Jira slash commands in the next few hours. Please let me know if you see any issues around this new logic.

srkgupta:
@DHaussermann I have updated the code with your comments. Here's how the new setup flow instruction looks like:

Screenshot 2023-04-25 at 12 23 20 PM
Screenshot 2023-04-25 at 12 23 39 PM

srkgupta:
@mickmister I have fixed the following issues and have updated the code. 

```
2. /jira setup says this message even if I'm already in the bot DM channel:
3. OAuth complete page does not automatically close. I fixed this in the e2e PR https://github.com/mattermost/mattermost-plugin-jira/pull/934
6. The webhook secret is not set in the webhook URL shown in the modal. Maybe this doesn't happen every time. Maybe the secret was not set somehow?
```

However, I am having troubles reproducing the following issues. Can you please re-check and see if you can reproduce this with the updated code. I am suspecting that there might have been some confusion while setting up with cloud & cloud-oauth flow. With the new code, the cloud setup flow is removed (as highlighted by @DHaussermann)

``` 
1. When executing /jira instance install cloud-oauth https://mmtest.atlassian.net, the modal didn't use the entered url when it opened
4. Not sure how this happened, but at some point the setup flow assumed a Jira URL that I used for a previous usage of the setup flow.
7. Sometimes the plugin thinks I'm trying to install 2 Jira instances when entering the client id & secret, even though it's the same Jira URL I entered before. Maybe it's /jira instance install that messes this up by queueing the instance in the kv store.
```

Also, regarding your following comment, can we handle this as a separate task/PR as this requirement is generic and applicable for cloud/cloud-oauth/server type Jira instances. 

```
5. Since plugin doesn't allow localhost as siteurl, we will need to support a special case for CI to use that.
```

mickmister:
> However, I am having troubles reproducing the following issues. Can you please re-check and see if you can reproduce this with the updated code. I am suspecting that there might have been some confusion while setting up with cloud & cloud-oauth flow. With the new code, the cloud setup flow is removed

Sure I'll take another look with using the code. These items could possibly be resolved at this point.

> Also, regarding your following comment, can we handle this as a separate task/PR as this requirement is generic and applicable for cloud/cloud-oauth/server type Jira instances.

Yes I should handle this in the e2e PR https://github.com/mattermost/mattermost-plugin-jira/pull/934

webapp/src/types/model.ts
```diff
@@ -157,7 +157,6 @@ export type ChannelSubscription = {
 export enum InstanceType {
     CLOUD = 'cloud',
     SERVER = 'server',
-    CLOUD_OAUTH = 'cloud_oauth'
```
mickmister: This seems wrong to get rid of this. Where are the other values used on the frontend?

server/user.go
```diff
@@ -33,7 +33,8 @@ type Connection struct {
 	Oauth1AccessSecret string        `json:",omitempty"`
 	OAuth2Token        *oauth2.Token `json:",omitempty"`
 	Settings           *ConnectionSettings
-	DefaultProjectKey  string `json:"default_project_key,omitempty"`
+	DefaultProjectKey  string   `json:"default_project_key,omitempty"`
+	MattermostUserID   types.ID `json:"mattermost_user_id"`
```
mickmister, line 37: We will want to make sure this change is backwards compatible. This value not be present for existing users

server/command.go
```diff
@@ -76,7 +79,7 @@ const commonHelpText = "\n" +
 
 const sysAdminHelpText = "\n###### For System Administrators:\n" +
 	"Install Jira instances:\n" +
-	"* `/jira instance install cloud [jiraURL]` - Connect Mattermost to a Jira Cloud instance located at <jiraURL>\n" +
+	"* `/jira instance install cloud [jiraURL]` - Deprecated: Connect Mattermost to a Jira Cloud instance located at <jiraURL>\n" +
```
mickmister, line 82: Maybe show this one last in the list, and say it this way:

```suggestion
	"* `/jira instance install cloud [jiraURL]` - Connect Mattermost to a Jira Cloud instance located at <jiraURL>. (Deprecated. Please use `cloud-oauth` instead.)\n" +
```

server/command.go
```diff
@@ -168,7 +171,7 @@ func createInstanceCommand(optInstance bool) *model.AutocompleteData {
 
 	jiraTypes := []model.AutocompleteListItem{
 		{HelpText: "Jira Server or Datacenter", Item: "server"},
-		{HelpText: "Jira Cloud (atlassian.net)", Item: "cloud"},
+		{HelpText: "(Deprecated) Jira Cloud (atlassian.net)", Item: "cloud"},
```
mickmister, line 174:
```suggestion
		{HelpText: "Jira Cloud (atlassian.net) (Deprecated. Please use `cloud-oauth` instead.)", Item: "cloud"},
```

server/command.go
```diff
@@ -799,6 +802,44 @@ func executeInstanceInstallCloud(p *Plugin, c *plugin.Context, header *model.Com
 	})
 }
 
+func executeInstanceInstallCloudOAuth(p *Plugin, c *plugin.Context, header *model.CommandArgs, args ...string) *model.CommandResponse {
+	authorized, err := authorizedSysAdmin(p, header.UserId)
+	if err != nil {
+		return p.responsef(header, "%v", err)
+	}
+	if !authorized {
+		return p.responsef(header, "`/jira install` can only be run by a system administrator.")
```
mickmister, line 811:
```suggestion
		return p.responsef(header, "`/jira install` can only be run by a Mattermost system administrator.")
```

server/setup_flow.go
```diff
@@ -93,6 +85,26 @@ func (p *Plugin) NewSetupFlow() *flow.Flow {
...
+			p.stepWebhook(),
+			p.stepWebhookDone(),
+			p.stepConnect(),
+			p.stepConnected(),
+			p.stepAnnouncementQuestion(),
+			p.stepAnnouncementConfirmation(),
+			p.stepCancel(),
+			p.stepDone(),
+		).
+		// WithDebugLog().
```
mickmister, line 104:
```suggestion
```

server/setup_flow.go
```diff
@@ -93,6 +85,26 @@ func (p *Plugin) NewSetupFlow() *flow.Flow {
...
+			p.stepInstalledJiraApp(),
+			p.stepWebhook(),
+			p.stepWebhookDone(),
+			p.stepConnect(),
+			p.stepConnected(),
+			p.stepAnnouncementQuestion(),
+			p.stepAnnouncementConfirmation(),
+			p.stepCancel(),
+			p.stepDone(),
+		).
```
mickmister, line 103: Interesting approach. So this is sort of a copy of the setup flow, but assumes the admin is using Cloud OAuth?

server/issue_parser.go
```diff
@@ -21,13 +20,8 @@ func parseJiraLinksToMarkdown(text string) string {
...
 
-func mdKeySummaryLink(issue *jira.Issue) string {
-	// Use Self URL only to extract the full hostname from it
-	pos := strings.LastIndex(issue.Self, "/rest/api")
-	if pos < 0 {
-		return ""
-	}
-	return fmt.Sprintf("[%s](%s%s)", issue.Key+": "+issue.Fields.Summary+" ("+issue.Fields.Status.Name+")", issue.Self[:pos], "/browse/"+issue.Key)
+func mdKeySummaryLink(issue *jira.Issue, instance Instance) string {
+	return fmt.Sprintf("[%s](%s%s)", issue.Key+": "+issue.Fields.Summary+" ("+issue.Fields.Status.Name+")", instance.GetJiraBaseURL(), "/browse/"+issue.Key)
```
mickmister, line 24: I know this was already here but I think we should make this change to make it a little easier to read:

```suggestion
	return fmt.Sprintf("[%s: %s (%s)](%s%s)", issue.Key, issue.Fields.Summary, issue.Fields.Status.Name, instance.GetJiraBaseURL(), "/browse/"+issue.Key)
```

server/instance_cloud_oauth.go
```diff
@@ -80,11 +83,37 @@ func (ci *cloudOAuthInstance) GetClient(connection *Connection) (Client, error)
...
 func (ci *cloudOAuthInstance) getClientForConnection(connection *Connection) (*jira.Client, *http.Client, error) {
 	oauth2Conf := ci.GetOAuthConfig()
-
 	ctx := context.Background()
 	tokenSource := oauth2Conf.TokenSource(ctx, connection.OAuth2Token)
 	client := oauth2.NewClient(ctx, tokenSource)
 
+	// Get a new token if Access Token has expired
+	if ci.Plugin.IsAccessTokenExpired(connection.OAuth2Token) {
+		ci.client.Log.Debug("AccessToken is about to expire (or) has expired. Refreshing the accessToken.")
```
mickmister, line 92:
```suggestion
		ci.client.Log.Debug("AccessToken is about to expire (or has expired). Refreshing the accessToken.")
```

server/instance_cloud_oauth.go
```diff
@@ -80,11 +83,37 @@ func (ci *cloudOAuthInstance) GetClient(connection *Connection) (Client, error)
 
 func (ci *cloudOAuthInstance) getClientForConnection(connection *Connection) (*jira.Client, *http.Client, error) {
 	oauth2Conf := ci.GetOAuthConfig()
-
 	ctx := context.Background()
 	tokenSource := oauth2Conf.TokenSource(ctx, connection.OAuth2Token)
 	client := oauth2.NewClient(ctx, tokenSource)
 
+	// Get a new token if Access Token has expired
+	if ci.Plugin.IsAccessTokenExpired(connection.OAuth2Token) {
```
mickmister, line 91: The Golang OAuth library will automatically refresh the token based on the expiry, so this is an unnecessary check. Though we are now using this as a signal to store a new refresh token as well, which may work fine. To solve this in other plugins, we usually just call `TokenSource`, and see if the OAuth library performed a refresh of the _refresh_ token, and store the new refresh token if it changed from beforehand.

server/instance_cloud_oauth.go
```diff
@@ -80,11 +83,37 @@ func (ci *cloudOAuthInstance) GetClient(connection *Connection) (Client, error)
...
+
+		connection.OAuth2Token = newToken
+		tokenSource = oauth2Conf.TokenSource(ctx, newToken)
+		client = oauth2.NewClient(ctx, tokenSource)
+
+		// Store this new access token & refresh token to get a new access token in future when it has expired
+		err = ci.Plugin.userStore.StoreConnection(ci.Common().InstanceID, connection.MattermostUserID, connection)
+		if err != nil {
+			return nil, nil, err
+		}
```
mickmister, line 114: Seems odd that we are calling `TokenSource` twice, but I don't think this will cause issues. The access token will not need refresh the second time, so we won't be receiving a new refresh token, so this may work well.

Here's a snippet of what I was referring to in regards to explicitly checking if the refresh token changed. Also, note that in the Zoom plugin, the OAuth client is being instantiated by the `token` instead of the `tokenSource`. https://github.com/mattermost/mattermost-plugin-zoom/blob/fcdab70c5da49c704ac701abb55ac5ee3c2489d7/server/zoom/oauth.go#L158-L183

server/instance_cloud_oauth.go
```diff
@@ -41,9 +42,10 @@ type JiraAccessibleResources []struct {
 var _ Instance = (*cloudOAuthInstance)(nil)
 
 const (
-	JiraScopes       = "read:jira-user,read:jira-work,write:jira-work"
-	JiraResponseType = "code"
-	JiraConsent      = "consent"
+	JiraScopes        = "read:jira-user,read:jira-work,write:jira-work"
+	JiraScopesOffline = JiraScopes + ",offline_access"
```
mickmister, line 46: Unless we require them to be separate (referencing `JiraScopes` const other than these two lines), I think we should just put `offline_access` in the `JiraScopes` constant and remove the `JiraScopesOffline` constant

server/user_cloud_oauth.go
```diff
@@ -141,3 +89,37 @@ func (p *Plugin) httpOAuth2Complete(w http.ResponseWriter, r *http.Request, inst
 		RevokeURL:             path.Join(p.GetPluginURLPath(), instancePath(routeUserDisconnect, instance.GetID())),
 	})
 }
+
+func (p *Plugin) GenerateOAuthToken(mattermostUserID string, instanceID types.ID, code string) (*Connection, error) {
```
mickmister, line 93:
```suggestion
func (p *Plugin) GenerateInitialOAuthToken(mattermostUserID string, instanceID types.ID, code string) (*Connection, error) {
```

mickmister (changes requested):
Thanks for addresing the feedback @srkgupta :+1: I have a few more requests and comments for discussion

server/user.go
srkgupta: Yes, for existing Cloud and Server Instance, this value will be nil. Currently we are only using this when a user creates a new Cloud Oauth instance. It is not used for any other instance type and can remain to be nil.

webapp/src/types/model.ts
srkgupta: Looking at the webapp code, I could only find the usage of `InstanceType` in the Connect Modal flow, i.e. `/jira connect`. However, even this works fine without the requirement of having the InstanceType `CLOUD_OAUTH = 'cloud_oauth'`. This was initially added to handle the slash command `/jira instance install cloud-oauth` via the webapp flow. However since this behavior was changed as per your review comment [here](https://github.com/mattermost/mattermost-plugin-jira/pull/928#discussion_r1156449925). Hence we no longer have any dependence or usage of having `CLOUD_OAUTH` in the InstanceType. Please let me know if I have overlooked something here?

server/setup_flow.go
srkgupta: The main purpose of using this approach is because the existing setup flow might or might not have triggered. Since we are reusing the code from the setup_flow when a user directly runs the slash command `/jira instance install cloud-oauth <url>`, we are starting the oauth2Flow for the user instead of initiating the setup flow from an intermediate step from the `NewSetupFlow`. I have tried with the later case and it does not seems to be working at all cases, i.e user might/might not have triggered the setup flow before.

server/instance_cloud_oauth.go
srkgupta: Yes, the purpose of having these separate is because, we reference `JiraScopes` in the setup instructions for users to add the specified scopes. We do not want users to get confused trying to find `offline_access` in step 5 here:

setup

server/instance_cloud_oauth.go
srkgupta: This is resolved with the latest changes.

mickmister:
> 6. The webhook secret is not set in the webhook URL shown in the modal. Maybe this doesn't happen every time. Maybe the secret was not set somehow?

I'm still experiencing this issue:



---

As well as the one below. Maybe you're using an MM Enterprise license so it's not reproducing in your environment?

> 7. Sometimes the plugin thinks I'm trying to install 2 Jira instances when entering the client id & secret, even though it's the same Jira URL I entered before. Maybe it's /jira instance install that messes this up by queueing the instance in the kv store.

srkgupta:
> Sometimes the plugin thinks I'm trying to install 2 Jira instances when entering the client id & secret, even though it's the same Jira URL I entered before. Maybe it's /jira instance install that messes this up by queueing the instance in the kv store.

@mickmister I have fixed the above issue. Thanks for adding the details allowing me to replicate and fix the issue. 

Regarding the webhook secret issue, this is something I haven't touched in this PR and should not be affected. The webhook secret is picked up from the plugin config. Hence in the E2E test, there might be a high chance that this config is empty/not set and hence you might be seeing that issue. Can you please recheck this?  I am unable to replicate this as I get an error `Using http://localhost:8065/ as your Mattermost SiteURL is not permitted, as the URL is not reachable from Jira. If you are using Jira Cloud, please make sure your URL is reachable from the public internet.` if I try to use the jira slash commands on localhost instance. 

> The webhook secret is not set in the webhook URL shown in the modal. Maybe this doesn't happen every time. Maybe the secret was not set somehow?

webhook

server/command.go
```diff
@@ -171,7 +171,7 @@ func createInstanceCommand(optInstance bool) *model.AutocompleteData {
 
 	jiraTypes := []model.AutocompleteListItem{
 		{HelpText: "Jira Server or Datacenter", Item: "server"},
-		{HelpText: "(Deprecated) Jira Cloud (atlassian.net)", Item: "cloud"},
+		{HelpText: "Jira Cloud (atlassian.net) (Deprecated. Please use cloud-oauth instead.)", Item: "cloud"},
 		{HelpText: "Jira Cloud OAuth 2.0 (atlassian.net)", Item: "cloud-oauth"},
```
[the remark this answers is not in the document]
srkgupta: This is updated in the latest code.

server/command.go
```diff
@@ -79,7 +79,7 @@ const commonHelpText = "\n" +
 
 const sysAdminHelpText = "\n###### For System Administrators:\n" +
 	"Install Jira instances:\n" +
-	"* `/jira instance install cloud [jiraURL]` - Deprecated: Connect Mattermost to a Jira Cloud instance located at <jiraURL>\n" +
+	"* `/jira instance install cloud [jiraURL]` - Connect Mattermost to a Jira Cloud instance located at <jiraURL>. (Deprecated. Please use `cloud-oauth` instead.)\n" +
 	"* `/jira instance install cloud-oauth [jiraURL]` - Connect Mattermost to a Jira Cloud instance using OAuth 2.0 located at <jiraURL>\n" +
 	"* `/jira instance install server [jiraURL]` - Connect Mattermost to a Jira Server or Data Center instance located at <jiraURL>\n" +
```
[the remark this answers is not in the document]
srkgupta: This is updated in the latest code.

mickmister:
> I am unable to replicate this as I get an error `Using http://localhost:8065/ as your Mattermost SiteURL is not permitted, as the URL is not reachable from Jira. If you are using Jira Cloud, please make sure your URL is reachable from the public internet.` if I try to use the jira slash commands on localhost instance.

The plugin will need to be built with `E2E_TESTING=true make deploy` in order to avoid this error.

> Hence in the E2E test, there might be a high chance that this config is empty/not set and hence you might be seeing that issue. Can you please recheck this?

There are defaults set for the config (including the webhook secret), during the plugin's start up time:

https://github.com/mattermost/mattermost-plugin-jira/blob/a6839f1dd85f5cb1e6b147850fc4bed02b7e45bc/server/plugin.go#L245

The e2e tests clear the plugin's config before starting the test suite, which clears the secret. So maybe the e2e tests need to restart the plugin after clearing the config, in order to let the plugin do its normal work of generating the default config before the tests run.

mickmister:
> So maybe the e2e tests need to restart the plugin after clearing the config, in order to let the plugin do its normal work of generating the default config before the tests run.

This indeed fixed the issue :+1:

mickmister (approved):
LGTM :tada:

mickmister:
@Dhaussermann This is ready for testing

DHaussermann:
I have tested this again with latest changes
- Setup looks good (I did not see the issue where the webhook secret is missing)
- Connect and disconnect is working
- ensured user remains connected over several days with no issue. 

@srkgupta I do sometimes see one issue but I'm not sure if anything can be done to improve it from the mattermost side. 
In one browser I now see the following message when I try to click the connection link 
image

- No issue occurs when a try a new session incognito with fresh cache. 
- When I open a tab to myatlassian - The correct user who made the oAuth app is singed in
image

Any thoughts on what's happening here? Perhaps there are multiple G-Suite amounts logged in and it is no letting me select my `@mattermost` user? My other thought is that multiple rounds testing with multiple oAuth apps has somehow contaminated my browser cache.

srkgupta:
@DHaussermann did you happen to copy the URL when you see this issue? I am mainly looking for the jwt token present in the context parameter of the URL. You can decode the jwt tokens using the website jwt.io and please share here and I can check if there's anything wrong?

DHaussermann:
@srkgupta to make the test a bit cleaner, I have spun up a new server and can still reproduce the issue with a new oauth app. 
The token seems to contain the correct details.
image
However, I see "Invalid Signature" at the bottom of the page.

Here is the token so you can examine it. Please ping me if I can provide any more details.
`eyJraWQiOiJiOWY2OTVlZmM3MzgwN2QxNmNhNmNkMzQiLCJhbGciOiJSUzI1NiIsInR5cCI6Imp3dCJ9.eyJuYmYiOjE2ODM1NTU5MjcsImlzcyI6ImF1dGguYXRsYXNzaWFuLmNvbSIsImlhdCI6MTY4MzU1NTkyNywiZXhwIjoxNjgzNTU5NTI3LCJhdWQiOiJhcGkuYXRsYXNzaWFuLmNvbSIsInVzZXJfaWQiOiI1YTYwMGZlM2ZmODExNzY5ODcxZjIyNDUiLCJyZWRpcmVjdF91cmkiOiIvYXV0aG9yaXplP2F1ZGllbmNlPWFwaS5hdGxhc3NpYW4uY29tJmNsaWVudF9pZD1nb3lZUlJyQWJDUzdQbnZrZ242T0VaTE9lSXhFSE5nWiZwcm9tcHQ9Y29uc2VudCZyZWRpcmVjdF91cmk9aHR0cHM6Ly9ka2gtamlyYS1mcmVzaDUudGVzdC5tYXR0ZXJtb3N0LmNsb3VkL3BsdWdpbnMvamlyYS9pbnN0YW5jZS9hSFIwY0hNNkx5OXRiWFJsYzNRdVlYUnNZWE56YVdGdUxtNWxkQSUzRCUzRC9vYXV0aDIvY29tcGxldGUuaHRtbCZyZXNwb25zZV90eXBlPWNvZGUmc3RhdGU9eGt5cmdpazl6dHI2aXFtX3Jvb3E4a3lkcTc4cDVla3llcmFjb3R6OGllJnNjb3BlPW9mZmxpbmVfYWNjZXNzK3JlYWQ6amlyYS11c2VyK3JlYWQ6amlyYS13b3JrK3dyaXRlOmppcmEtd29yayIsImNsaWVudCI6eyJjbGllbnRfaWQiOiJnb3lZUlJyQWJDUzdQbnZrZ242T0VaTE9lSXhFSE5nWiIsIm5hbWUiOiJNYXR0ZXJtb3N0IEppcmEgUGx1Z2luIC0gZGtoLW1tLWppcmEtZnJlc2g1IiwibG9nb191cmkiOiJodHRwczovL2F2YXRhci1tYW5hZ2VtZW50LS1hdmF0YXJzLnVzLXdlc3QtMi5wcm9kLnB1YmxpYy5hdGwtcGFhcy5uZXQvZGVmYXVsdC1hdmF0YXIucG5nIn19.ac4oQD2A5jqX8pRbGVNTOqDTkbRLVXA6QADsk-vsYd5dza757ctc-Qrjw3-MWKOmefxZlNzTInVQpq3m0ZRPNYZ-lZ2x0QmYNhERh1mcdA8PfY9ozFHc2q1D5JBolxBYbEdtdGFF5DTXCquYNBpHduwpsZOnVNNVyAkuA2i_o3zsx1AaMQGVZ7Cu7ABkx6Dw8cnDTki0DZg3OLm8CZs9a8cPooT3sDJ3Vh48BJ9bl6iou9rmi2fgR_DwXHwJJWhXrZNzXEctmDt_pbQSI6g2oQENcJDyJtPvWKc8oN-YYNyrGKIp8prroCDL4kfs9LBZnNrUtvjFhqsbXCbtTBopMg`

hanzei:
Closing as not a priority

mickmister:
Reopening to resolve https://github.com/mattermost/mattermost-plugin-jira/issues/945#issuecomment-1568816602

mickmister:
I updated e2e tests https://github.com/mattermost/mattermost-plugin-jira/pull/934 with this PR and they are passing for me in my test environment

mickmister:
@jupenur @esarafianou I'm wondering your thoughts on the workload of reviewing this PR

There is an issue https://github.com/mattermost/mattermost-plugin-jira/issues/945 that has started occurring with Jira's authentication API, that is making it so no MM servers can create new user connections to Jira Cloud instances.

Replacing our existing authentication strategy with OAuth2 resolves the issue, which has made this PR more relevant. We're looking into potentially solving the issue more directly with less change, but there are no findings there thus far, so this fix may be necessary to have Jira Cloud user authentication working again.

mickmister:
@DHaussermann There is a Jira Cloud-specific piece of the webhook logic that will need to be tested here. When comment events come through, we have to fetch extra data when working with a Jira Cloud instance. We'll want to make sure this still works correctly with the OAuth2 setup.

https://github.com/mattermost/mattermost-plugin-jira/blob/7c1b921343f583c9d16fbe06463d5c6f58ad80fe/server/webhook_jira.go#LL87C1-L87C1-L94

mickmister:
@Kshitij-Katiyar Note that the above comment will need to be addressed in this PR. The below if statement may need to be modified like so:

```diff
- if isCommentEvent && instance.Common().Type == "cloud" {
+ if isCommentEvent && instance.Common().Type == "cloud" || instance.Common().Type == "cloud-oauth" {
```

or create new method for this:
```go
if isCommentEvent && instance.Common().IsJiraCloud() {
```

and also ensure that the comment received event feature works correctly with the new setup

mickmister:
Closing and reopening with another PR

PR #983, opened 2023-10-10
[GH-939]: Updated the message for webhookURL in step 2 of plugin setup

Kshitij-Katiyar:
### Summary
- Updated the message in webhookURL modal in the second step of the plugin setup
#### Old
Screenshot from 2023-10-10 13-45-05

#### New
Screenshot from 2023-10-10 13-43-30

### Issue link
#939 

### What to test
- Content of the modal in step 2 of plugin setup(webhook URL step)
### How to test
- Install the plugin
- run `jira setup`
- open the JIRA bot dm
- Complete Step 1
- Modal to be tested will be visible at the end of step 2
Screenshot from 2023-10-10 13-48-04

### DIFF
--- a/server/setup_flow.go
+++ b/server/setup_flow.go
@@ -58,7 +58,7 @@ const (
 )
 
 const (
-	newline    = "\n"
+	lineBreak    = "\n"
 	webhookURL = "\n [{{.WebhookURL}}]({{.WebhookURL}})"
 )
 
@@ -405,7 +405,7 @@ func (p *Plugin) stepWebhook() flow.Step {
 			Color: flow.ColorPrimary,
 			Dialog: &model.Dialog{
 				Title:            "Jira Webhook URL",
-				IntroductionText: "Please copy and use the link below as webhook URL. Once you have entered all options and the webhook URL, select **Create**. \n" + fmt.Sprintf("%s %s", newline, webhookURL),
+				IntroductionText: fmt.Sprintf("Please copy and use the link below as webhook URL. Once you have entered all options and the webhook URL, select **Create**. %s", lineBreak) + fmt.Sprintf("%s %s", lineBreak, webhookURL),
 				SubmitLabel:      "Continue",
 			},
 			OnDialogSubmit: flow.DialogGoto(stepWebhookDone),
--- a/server/setup_flow.go
+++ b/server/setup_flow.go
@@ -58,8 +58,8 @@ const (
 )
 
 const (
-	lineBreak    = "\n"
-	webhookURL = "\n [{{.WebhookURL}}]({{.WebhookURL}})"
+	lineBreak  = "\n"
+	webhookURL = "[{{.WebhookURL}}]({{.WebhookURL}})"
 )
 
 func (p *Plugin) NewSetupFlow() *flow.Flow {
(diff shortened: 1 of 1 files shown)


server/setup_flow.go
```diff
@@ -400,7 +405,7 @@ func (p *Plugin) stepWebhook() flow.Step {
 			Color: flow.ColorPrimary,
 			Dialog: &model.Dialog{
 				Title:            "Jira Webhook URL",
-				IntroductionText: "Please scroll to select the entire URL if necessary.\n\n```{{.WebhookURL}}```\n\nOnce you have entered all options and the webhook URL, select **Create**",
+				IntroductionText: fmt.Sprintf("Please copy and use the link below as webhook URL. Once you have entered all options and the webhook URL, select **Create**. %s", lineBreak) + fmt.Sprintf("%s %s", lineBreak, webhookURL),
```
mickmister, line 408: Putting a line break into a variable seems kind of overkill to me. Is this a functional requirement or for readability? The developer has to trace positional `%s`'s to determine where those are being used. This is a small piece of code so it doesn't matter that much but just bringing it up for discussion

mickmister (approved):
LGTM

server/setup_flow.go
cwarnermm, line 408:
```suggestion
				IntroductionText: fmt.Sprintf("Please copy the link below as the webhook URL. Once you have entered all options and the webhook URL, select **Create**. %s", lineBreak) + fmt.Sprintf("%s %s", lineBreak, webhookURL),
```

cwarnermm: @Kshitij-Katiyar - The updated text doesn't match the the popup dialog itself. The user only has 2 options: Cancel or Continue. The user isn't able to see any fields to populate.

I recognize that your PR hasn't made changes to that existing text. Are you open to updating your proposed text to provide more guidance to the user who encounters this dialog? If so, what would you recommend? What instructions on this dialog would help guide the user interested in setting up their Jira Webhook?

mickmister: @cwarnermm This dialog is actually instructing the admin to interact with a form in Jira's UI, for installing the integration on that end. There is indeed no indication on the dialog itself, but I believe there is an instruction in a post that opens the dialog.

https://github.com/mattermost/mattermost-plugin-jira/blob/7534a9747b3e53a5f69a74660aaa7bc9f3c00869/server/setup_flow.go#L385-L400

The last line "Select **View Webhook URL**" is referring to the button below the post that opens the dialog

cwarnermm: Thanks, @mickmister, for this context.
@Kshitij-Katiyar - I approve this PR assuming the suggested change is applied to the dialog text. Thanks!

cwarnermm (approved):

server/setup_flow.go
mickmister: I'm thinking we should specify that the create button is in Jira though. For example:

> Once you have entered all options and the webhook URL, select **Create** on the Jira create webhook page.

cwarnermm: I agree with your proposal, @mickmister! Thank you! @Kshitij-Katiyar - Please note and update the text as proposed.

mickmister: FYI there's an unresolved request in this thread, to make sure we mention the `Create` button is in Jira's UI

cwarnermm: @Kshitij-Katiyar - Is this update something you're open to creating a PR for?

PR #1089, opened 2024-06-26
[GH-1078]: fixed the alignment issue of loading ellipses

Kshitij-Katiyar:
#### Summary
Fix the alignment issue of the loading ellipses

##### Existing
Screenshot from 2024-06-27 16-11-04

##### New
Screenshot from 2024-06-28 13-26-00

#### Issue
Fixes https://github.com/mattermost/mattermost-plugin-jira/issues/1078


webapp/src/components/loading.scss
```diff
@@ -0,0 +1,3 @@
+.loading-panel {
+    align-items: baseline !important;
```
raghavaggarwal2308, line 2: do we need !important here?

webapp/src/components/loading.tsx
raghavaggarwal2308, line 1: Add both of these files to a new folder `loading`

webapp/src/components/loading.scss
raghavaggarwal2308: Also, this should be centre aligned

raghavaggarwal2308 (changes requested):

webapp/src/components/loading.scss
Kshitij-Katiyar: @raghavaggarwal2308 Mattermost's styling is overriding the styling provided by us hence using `!important` is necessary.
the existing styling with which the ellipses are not aligned with the text is `center-aligned`, using `baseline` fixes the issue.

webapp/src/components/modals/channel_subscriptions/edit_channel_subscription.tsx
```diff
@@ -8,7 +8,7 @@ import ReactSelectSetting from 'components/react_select_setting';
 import ConfirmModal from 'components/confirm_modal';
 import FormButton from 'components/form_button';
 import Input from 'components/input';
-import Loading from 'components/loading';
+import Loading from 'components/loading/loading';
```
raghavaggarwal2308, line 11: rename loading.tsx to index.tsx. Then you will not have to make this change

raghavaggarwal2308 (changes requested):

Kshitij-Katiyar:
@raghavaggarwal2308 Updated the file name as per the suggestion, please have a look.

raghavaggarwal2308 (approved):

webapp/src/components/loading/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+import React, {PureComponent} from 'react';
+import './loading.scss';
+
+type Props = {
+    position?: 'absolute' | 'fixed' | 'relative' | 'static' | 'inherit';
+    style?: object;
+};
+
+export default class Loading extends PureComponent<Props> {
```
mickmister, line 12: There are two component files with the same contents

webapp/src/components/loading/loading.scss
```diff
@@ -0,0 +1,3 @@
+.loading-panel {
+    align-items: baseline !important;
```
mickmister, line 2: Is this approach taken from some resource like a stack overflow post?

matthewbirtch:
@Kshitij-Katiyar Based on the screenshot it looks like the loading ellipses is still not vertically centered with the text. It's sitting too low. Can we tweak that?

webapp/src/components/loading/loading.scss
Kshitij-Katiyar: @mickmister  Using `!important` here because mattermost also applies some styling to the `loading__content` class which overrides any styling we provide.
This approach was just hit & trial, not exactly taken from any resource.

Kshitij-Katiyar:
> @Kshitij-Katiyar Based on the screenshot it looks like the loading ellipses is still not vertically centered with the text. It's sitting too low. Can we tweak that?

@matthewbirtch @mickmister Sure, here is the new alignment for the ellipses, please have a look.
Screenshot from 2024-06-28 13-26-00

matthewbirtch (approved):
Thanks @Kshitij-Katiyar. Looks better

webapp/src/components/loading/loading.scss
mickmister: > Using `!important` here because mattermost also applies some styling to the `loading__content` class which overrides any styling we provide.

If it's practical to just use a new css class, we may want to do that to avoid css conflicts

webapp/src/components/loading/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+// See LICENSE.txt for license information.
+
+import React from 'react';
+import './loading.scss';
+
+const Loading = () => {
+    return (
+        <div
+            className='loading-screen'
+            style={{position: 'relative'}}
```
mickmister, line 11: This PR addas a scss file and also does some style here. I'm thinking we should use just use `style` and avoid adding the small scss file

webapp/src/components/loading/loading.scss
```diff
@@ -0,0 +1,3 @@
+.loading-text {
+    margin: 20px 0;
+}
```
mickmister, line 3: This could have an impact on the core web application or any other plugin if they happen to use this css class. We should either have class name include `jira` or just put stuff in `style`

webapp/src/components/loading/index.tsx
Kshitij-Katiyar: @mickmister Updated the file with inline styling and deleted the scss file. Please have a look.

mickmister (approved):

wiggin77 (approved):

PR #1103, opened 2024-07-16
[GH-1011] Added conditional to avoid posting multiple notifications in case of duplicate subscriptions

### LINKED ISSUE #1011: The user is getting multiple notifications in the channel or DM/GM on MM for a event on Jira when duplicate subscriptions are created.
### Summary
The user can create duplicate subscription for an existing subscription and it results in generating same notification the number of times the subscription is duplicated.
### Steps to reproduce
1. Create a subscription using the `/jira subscribe edit` slash command in any desired channel or DM/GM on MM.
2. Create a duplicate subscription for the above added subscription using the `/jira subscribe edit` slash command in the same desired channel or DM/GM on MM.
3. Create the event on Jira for which the above subscriptions were created in the desired channel or DM/GM on MM.
### Expected behavior
The user should get only one notification for the event in the desired channel or DM/GM on MM.
### Actual behavior
The user is getting N number of notifications for the event in the desired channel or DM/GM where N is the number of subscriptions created for that event on MM.
### Screenshots
image

raghavaggarwal2308:
#### Summary
- Added conditional to avoid posting multiple notifications in case of duplicate subscriptions

#### What to test:
- Not getting multiple notifications in a channel in case of duplicate subscriptions.

#### How to test:
1. Connect you Jira account.
2. Create multiple subscription fora same event.
3. Trigger the desired event from Jira.

#### Ticket Link
Fixes  #1011

### DIFF
--- a/server/subscribe.go
+++ b/server/subscribe.go
@@ -211,6 +211,7 @@ func (p *Plugin) getChannelsSubscribed(wh *webhook, instanceID types.ID) ([]Chan
 	for _, sub := range subIds {
 		if p.matchesSubsciptionFilters(wh, sub.Filters) {
 			if !subscriptionMap[sub.ChannelID] {
+				subscriptionMap[sub.ChannelID] = true
 				channelSubscriptions = append(channelSubscriptions, sub)
 			}
 		}
--- a/server/subscribe.go
+++ b/server/subscribe.go
@@ -206,10 +206,11 @@ func (p *Plugin) getChannelsSubscribed(wh *webhook, instanceID types.ID) ([]Chan
 	}
 
 	var channelSubscriptions []ChannelSubscription
+	subscriptionMap := make(map[string]bool)
 	subIds := subs.Channel.ByID
 	for _, sub := range subIds {
 		if p.matchesSubsciptionFilters(wh, sub.Filters) {
-			if !HasSubscription(sub, channelSubscriptions) {
+			if !subscriptionMap[sub.ChannelID] {
 				channelSubscriptions = append(channelSubscriptions, sub)
 			}
 		}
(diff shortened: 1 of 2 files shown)


server/subscribe.go
```diff
@@ -209,7 +209,9 @@ func (p *Plugin) getChannelsSubscribed(wh *webhook, instanceID types.ID) ([]Chan
 	subIds := subs.Channel.ByID
 	for _, sub := range subIds {
 		if p.matchesSubsciptionFilters(wh, sub.Filters) {
-			channelSubscriptions = append(channelSubscriptions, sub)
+			if !HasSubscription(sub, channelSubscriptions) {
```
wiggin77, line 212: Why not use a map for avoid duplicates instead of walking the list for every append?

Kshitij-Katiyar: Hey @wiggin77, I have implemented the suggestion. Please have a look.

server/subscribe.go
```diff
@@ -206,10 +206,13 @@ func (p *Plugin) getChannelsSubscribed(wh *webhook, instanceID types.ID) ([]Chan
 	}
 
 	var channelSubscriptions []ChannelSubscription
+	subscriptionMap := make(map[string]bool)
 	subIds := subs.Channel.ByID
 	for _, sub := range subIds {
 		if p.matchesSubsciptionFilters(wh, sub.Filters) {
-			channelSubscriptions = append(channelSubscriptions, sub)
+			if !subscriptionMap[sub.ChannelID] {
+				channelSubscriptions = append(channelSubscriptions, sub)
```
wiggin77, line 214: You need to add it to the map for it to avoid collisions.

wiggin77 (changes requested):

server/subscribe.go
Kshitij-Katiyar: @wiggin77 updated the map for every iteration. Please have a look

wiggin77 (approved):

AayushChaudhary0001 (approved):
This PR has been tested for the following scenarios:-
1. Multiple subscriptions for the same event
2. Duplicate subscription for the same event

The PR is working fine for the above conditions, LGTM. Approved

Issue #1106, opened 2024-07-25
Update the default max attachment size


raghavaggarwal2308:
#### Summary
Update the value of default maximun attachment size to the max attachment size configured for Mattermost

For more details about `MaxAttchmentSize` refer: https://github.com/mattermost/mattermost-plugin-jira/issues/977

PR #1110, opened 2024-07-30
[MM-1106]: Updated the default value for maximum attachment size for Jira plugin

Kshitij-Katiyar:
#### Summary
Updated the default value for maximum attachment size for Jira plugin

#### Ticket Link
  Fixes https://github.com/mattermost/mattermost-plugin-jira/issues/1106

### DIFF
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -83,7 +83,7 @@ type externalConfig struct {
 	DisplaySubscriptionNameInNotifications bool
 }
 
-const defaultMaxAttachmentSize = utils.ByteSize(10 * 1024 * 1024) // 10Mb
+const defaultMaxAttachmentSize = utils.ByteSize(100 * 1024 * 1024) // 10Mb
 
 type config struct {
 	// externalConfig caches values from the plugin's settings in the server's config.json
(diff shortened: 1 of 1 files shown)


server/plugin.go
```diff
@@ -83,7 +83,7 @@ type externalConfig struct {
 	DisplaySubscriptionNameInNotifications bool
 }
 
-const defaultMaxAttachmentSize = utils.ByteSize(10 * 1024 * 1024) // 10Mb
+const defaultMaxAttachmentSize = utils.ByteSize(100 * 1024 * 1024) // 100Mb
```
wiggin77, line 86: My understanding from our last conversation was that the max file attachment size for Jira would match the max upload setting configured for MM server, unless the user explicitly overrides the setting in Jira.

Kshitij-Katiyar: @wiggin77 I have made the changes as suggested by you, please have a look. Also, increased the `defaultMaxAttachmentSize` from 10 to 100, which is the same as the default value indicated by Mattermost for `maxFileSize`.

wiggin77, line 86: I'm not understanding why we would need a `defaultMaxAttachmentSize` in the Jira plugin.  The value is either the MM server `MaxFileSize` or the value specified in the plugin config.  It would never be this default.

raghavaggarwal2308 (approved):

AayushChaudhary0001 (approved):
The PR is tested and code changes are working fine for updating default maximum attachment size.
LGTM, Approved

Issue #794, opened 2021-08-16
Data missing from Jira issue when unsupported custom fields are mandatory


DHaussermann:
In a scenario where a user is trying to save an issue in Jira and there are required custom fields required to save the issue an alternate workflow occurs. An ephemeral post is shown with details about which fields are not supported. The user is given a link to follow to finish saving the issue. 

In this scenario, the links should have all the populated fields included in the URL and should then populated on the jira form when you navigate to Jira. Now, some of the fields I passed in via the modal are not shown in the Jira UI when I go to finish saving the issue but the rest are there.

This behavior only occurs when one of the following fields is mandatory...
`Date Picker`
`Date Time Picker`
`Number Field`

The values for the fields are supplied in the modal but do not make it to JIra. The following fields are affected:

- Labels
- Select List (multiple choices)
- User Picker (single user)
- Checkboxes
- Components
- Fix version

Steps:
1. On a Jira instance create the following custom fields `Date Picker`, `Date Time Picker`, `Number Field`  
2. Configure the fields to be mandatory for the Screen / Project / Issue type you are using
3. Attempt to create an issue through Mattermost 
4. Provide values for all fields 
5. Create the issue
6. Follow the link

**Observed**" Some values provided in step 4. are not shown in Jira

PR #1127, opened 2024-09-26
[MM-794]: Fixed the issue of fields not prefilled when creating issue with custom fields

Kshitij-Katiyar:
### Summary
Fixed the issue of fields not being prefilled when creating issues with custom fields. Unable to support the fields related to the users because the Username is not available in the create issue data and all the user picker fields on the Jira create issue form work on the username.
##### Added Support for: 
- Custom Labels
- Select list(multiple choices)
- Checkboxes
#####  Fields still not supported
- User picker
- Assignee
- Reporter
- Labels

### Issue link
 Fixes https://github.com/mattermost/mattermost-plugin-jira/issues/794

### Video Link
https://flonnect.com/video/6d26327abd0d-4a7d-855a-3f4c909e8648

### How to test
- On a Jira instance create the following custom fields Date Picker, Date Time Picker, Number Field
- Configure the fields to be mandatory for the Screen / Project / Issue type you are using
- Attempt to create an issue through Mattermost
- Provide values for all fields
- Create the issue
- Follow the link


server/client.go
```diff
@@ -333,14 +333,30 @@ func MakeCreateIssueURL(instance Instance, project *jira.Project, issue *jira.Is
...
+	for key, field := range issue.Fields.Unknowns {
+		switch fieldTyped := field.(type) {
+		case string:
+			q.Add(key, fieldTyped)
+		case map[string]interface{}:
+			if id, ok := fieldTyped["id"].(string); ok {
+				q.Add(key, id)
+			}
+		case []interface{}:
+			for _, element := range fieldTyped {
```
wiggin77, line 345: @Kshitij-Katiyar can you add some comments indicating what field type(s) each case represents.

wiggin77 (changes requested):
Looks good.  Needs some comments.

server/client.go
Kshitij-Katiyar: @wiggin77 Added the comment for what fields are handled in each case.
please have a look

wiggin77 (approved):

raghavaggarwal2308 (approved):

AayushChaudhary0001:
@Kshitij-Katiyar While testing this PR, it was found that the Labels were not working on contrary to what it is written in the decsription. On further testing, it was found that the labels already created from Jira are not supported, whereas the labels field which is self created(custom) is supported.

Kshitij-Katiyar:
> @Kshitij-Katiyar While testing this PR, it was found that the Labels were not working on contrary to what it is written in the decsription. On further testing, it was found that the labels already created from Jira are not supported, whereas the labels field which is self created(custom) is supported.

@AayushChaudhary0001 Yes, due to technical limitations, supporting Jira labels for this use case is not possible as of now, I have updated the description.

AayushChaudhary0001:
Apart from this, rest everything is working fine, LGTM. Approved.

AayushChaudhary0001 (approved):

PR #1145, opened 2025-01-09
[MM-62462] Plugin link tooltip for Jira plugin opens takes a very long time to load

M-ZubairAhmed:
#### Summary
Changes to adopt with the new plugin link tooltip component for this plugin

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-62462

#### Screenshot

CleanShot 2025-01-09 at 15 12 23

#### Notes
This is a breaking change and should be included in Mattermost v10.5 thats where https://github.com/mattermost/mattermost/pull/29727 is scheduled to be added


webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
```diff
@@ -77,18 +77,17 @@ export default class TicketPopover extends React.PureComponent<Props, State> {
         return null;
     };
 
-    componentDidUpdate() {
+    componentDidMount() {
```
hanzei, line 80: Does that mean we query the API on every mouseover?

hanzei (changes requested):
A couple of questions from my side:

1. Do we also need to update https://github.com/mattermost/mattermost-plugin-github?
2. Given this is a breaking change, do we need to mention it in the changelog?
3. Do we need to cut a new release of the JIRA plugin and repackage it with v10.5?

M-ZubairAhmed:
> 1. Do we also need to update https://github.com/mattermost/mattermost-plugin-github?



> A couple of questions from my side:
> 
>     1. Do we also need to update https://github.com/mattermost/mattermost-plugin-github?
> 
>     2. Given this is a breaking change, do we need to mention it in the changelog?
> 
>     3. Do we need to cut a new release of the JIRA plugin and repackage it with v10.5?

1. That plugin is not effected because it uses the useEffect with correct dependency
2. Yes we need to mention, i mentioned in the webapp changelogs
3. Yes added you to the thread in Staff team

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
M-ZubairAhmed: Yes, plugins can implement caching on their side to avoid repeated API calls. However, from the webapp performance perspective, this component will get unmounted and remounted.

hanzei: Does the JIRA plugin implement such caching? In other words, does the tooltip load as quickly as before the webapp before on the 2nd mouseover?

M-ZubairAhmed: It loads depending on the network request resolution.

hanzei (approved):
I'm still not sure if the number of requests to the JIRA plugin changes with this PR or not. Per https://github.com/mattermost/mattermost-plugin-jira/pull/1145/files#r1910298062, I suppose it doesn't.

QA should check that during testing.

webapp/src/components/jira_ticket_tooltip/jira_ticket_tooltip.tsx
wiggin77: Ticket created:  https://mattermost.atlassian.net/browse/MM-62571

wiggin77 (approved):

wiggin77:
@M-ZubairAhmed let me know when this is merged so I can create a release and update the prepackaged version in MM.

wiggin77:
@M-ZubairAhmed the minimum server version will need to be updated in plugin.json

M-ZubairAhmed:
> @M-ZubairAhmed let me know when this is merged so I can create a release and update the prepackaged version in MM.

Does this need additional dev review? It has 2 approved.

wiggin77:
> > @M-ZubairAhmed let me know when this is merged so I can create a release and update the prepackaged version in MM.
> 
> Does this need additional dev review? It has 2 approved.

It looks like @hmhealey 's main concern was a ticket to handle the continuous mouse-over case.  That ticket is created.  @M-ZubairAhmed this is good to merge once the merge conflict is resolved.

PR #1171, opened 2025-02-14
fixed issue of plugin config not prefilling from previous build

Kshitij-Katiyar:
#### Summary
Fixed the issue of config not prefilling when uploading a new build of release 4.2.0

#### Testing step
- Upload release 4.1.1 build
- Generate and set all plugin config
- Create & test webhook
- Upload new build
- Bool and string configs should be prefilled to the previous configs and webhooks should remain working

### DIFF
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -193,7 +193,7 @@ func (p *Plugin) OnConfigurationChange() error {
 	}
 
 	encryptionKey := ec.EncryptionKey
-	if encryptionKey == "" {
+	if encryptionKey == "" && ec.AdminAPIToken != "" {
 		p.client.Log.Warn("Encryption key required to encrypt admin API token")
 		return errors.New("failed to encrypt admin token. Encryption key not generated")
 	}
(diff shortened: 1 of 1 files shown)


server/plugin.go
```diff
@@ -193,7 +193,7 @@ func (p *Plugin) OnConfigurationChange() error {
 	}
 
 	encryptionKey := ec.EncryptionKey
-	if encryptionKey == "" {
+	if encryptionKey == "" && ec.AdminAPIToken != "" {
```
raghavaggarwal2308, line 196: Should have separate errors for these

raghavaggarwal2308 (changes requested):

server/plugin.go
Kshitij-Katiyar: encryption key is only needed when admin API token is present

raghavaggarwal2308, line 196:
```suggestion
	if ec.AdminAPIToken != "" && encryptionKey == "" {
```

raghavaggarwal2308 (approved):

PR #1176, opened 2025-02-24
Add aider files to gitignore

wiggin77:
#### Summary
Add Aider files to `.gitignore`

#### Ticket Link
NONE

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,5 +26,4 @@ server/manifest.go
 
 # Aider
 .aider.*
-!.aider.conf.yml
 
--- a/.gitignore
+++ b/.gitignore
@@ -25,7 +25,7 @@ server/manifest.go
 .notice-work
 
 # Aider
-.aider*
+.aider.*
 !.aider.conf.yml
 !.aiderignore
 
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -23,3 +23,9 @@ server/manifest.go
 
 # notice
 .notice-work
+
+# Aider
+.aider*
+!.aider.conf.yml
```
BenCookie95, line 29: I don't think this should be excluded from the ignore, I have my anthropic api key in the conf file so I think it makes sense to ignore it. Also, aider configs are likely different for every user

BenCookie95 (changes requested):

.gitignore
```diff
@@ -23,3 +23,9 @@ server/manifest.go
 
 # notice
 .notice-work
+
+# Aider
+.aider.*
+!.aider.conf.yml
```
wiggin77, line 29:
```suggestion

```

.gitignore
```diff
@@ -23,3 +23,9 @@ server/manifest.go
...
+!.aiderignore
```
wiggin77, line 30:
```suggestion

```

BenCookie95 (approved):

Issue #1142, opened 2024-12-03
Convert reducer files from JS to TS


Kshitij-Katiyar:
Reference PR
https://github.com/mattermost/mattermost-plugin-jira/pull/1140

gentcod:
@Kshitij-Katiyar  can you shed more light on specific files you would like to get converted?

Kshitij-Katiyar:
@gentcod This task is up for any Js/Jsx file in the code.
It would be good to start with more important files related to reducer/action/hooks.

gentcod:
@Kshitij-Katiyar I can see it has been implemented for actions and hooks, I want to take up the changes related to the reducers

Kshitij-Katiyar:
@gentcod Go ahead, assigned you the issue. All the best

Kshitij-Katiyar:
@gentcod converted_FILENAME_to_ts

karan-singhh:
has this issue been fixed? if not I'd like to work on it

PR #1177, opened 2025-02-25
Converted reducer/index.js to TS

gentcod:
#### Summary

Based on requirement for issue [1142](https://github.com/mattermost/mattermost-plugin-jira/issues/1142#issuecomment-2681013834)
- Created an interface to specify types for action and state objects
- Refactored needed code


gentcod:
@Kshitij-Katiyar

Kshitij-Katiyar:
> @Kshitij-Katiyar kindly reference

@gentcod Thanks, will review it soon

webapp/src/reducers/index.ts
```diff
@@ -5,7 +5,19 @@ import {combineReducers} from 'redux';
...
-function installedInstances(state = [], action) {
+export type Action<T extends string = string> = {
+    type: T
+}
+
+export interface AnyAction extends Action {
+    [key: string]: any
+}
+
+export interface AnyState {
```
Kshitij-Katiyar, line 16: Is there any way of not using any in the code?

gentcod: I think any is okay in this case. The value could be anything really. Could be an object, a number or a string

webapp/src/reducers/index.ts
```diff
@@ -181,7 +193,7 @@ const channelSubscriptions = (state = {}, action) => {
     case ActionTypes.EDITED_CHANNEL_SUBSCRIPTION: {
         const sub = action.data;
         const newSubs = state[sub.channel_id].concat([]);
-        newSubs.splice(newSubs.findIndex((s) => s.id === sub.id), 1, sub);
+        newSubs.splice(newSubs.findIndex((s: any) => s.id === sub.id), 1, sub);
```
Kshitij-Katiyar, line 196: can a specific type be added to this subscription value? similar for line 176

webapp/src/reducers/index.ts
Kshitij-Katiyar: Got it

webapp/src/reducers/index.ts
Kshitij-Katiyar:
```
export type ChannelSubscription = {
    id: string;
    channel_id: string;
    filters: ChannelSubscriptionFilters;
    name: string;
    instance_id: string;
}
```

can the type be useful here?

gentcod: I had seen this before, this should be the type definition for `action.data`. It could also be stated as a type definition for s if the state looks like this: `{  [channel_id: string]: ChannelSubscription[]  }`.  So if the data to be return is in that form, then it is fine to have `s: ChannelSubscription` but in a case where state is in the form: `{  [channel_id: string]: Subs[]  }`, where Subs is like ` type Subs = { id: string; name: string; }` there will be errors

gentcod: Looking at this though, ` newSubs.splice(newSubs.findIndex((s: any) => s.id === sub.id), 1, sub);`, it is safe to say, s is of type ChannelSubscription, so I'll have it modified

Kshitij-Katiyar:
@gentcod LGTM except some lint issues, can you resolve them please

gentcod:
@Kshitij-Katiyar any update on this?

Kshitij-Katiyar (approved):

Kshitij-Katiyar:
Approved from my side
@wiggin77 will do the final review and then we can have it for the QA

Kshitij-Katiyar:
@gentcod Thanks, LGTM. Approved
@wiggin77 will do the final pass and then we can go with QA

gentcod:
@Kshitij-Katiyar Alright..

Kshitij-Katiyar:
@AayushChaudhary0001 Please test this PR
@gentcod Can you share the impact areas for testing

AayushChaudhary0001 (approved):
The PR is tested and code changes are working fine on changing reducer/index.js to TS. All the basic functionalities are working fine for both Jira local server and cloud instance.
LGTM, Approved

PR #1201, opened 2025-05-13
Added support for Enterprise Advanced license

marianunez:
#### Summary

Added support for Enterprise Advanced sku by:

 - Updating plugin api to get latest support for Enterprise Advanced through the latest server public dependency.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-63646


server/enterprise/license.go
```diff
@@ -27,19 +27,9 @@ func NewEnterpriseChecker(api PluginAPI) Checker {
...
-)
-
 func (e *enterpriseChecker) HasEnterpriseFeatures() bool {
 	config := e.api.GetConfig()
 	license := e.api.GetLicense()
 
-	if license != nil && (license.SkuShortName == e20 || license.SkuShortName == enterprise || license.SkuShortName == professional) {
-		return true
-	}
-
```
marianunez, line 43: Removed this check as the function below in the plugin api takes care of this check and we don't have to mantain the sku names here twice.

harshilsharma63 (approved):

server/enterprise/license.go
marianunez: @wiggin77 actually the test surfaced that this was originally allowing `professional` in the checks I removed but had the higher sku check in the api below. 

I'm updating this to use the Profressional check - is this correct for multi instance feature?

wiggin77: @marianunez  Indeed multi-instance has been available for Professional and up.  I don't think your intention is to change that., although I should probably confirm with product.  I'll do another PR if that needs to change.

wiggin77 (approved):

server/enterprise/license.go
marianunez: Correct, no intention on moving any feature from their current sku

PR #1235, opened 2025-10-28
[MM-66329]: Block attaching post without access to issues via attach-comment-to-issue API

abbas-dependable-naqvi:
#### Summary
Block attaching the post to which user does not have access.

#### Ticket Link
  Fixes https://mattermost.atlassian.net/browse/MM-66329

### DIFF
--- a/server/issue_test.go
+++ b/server/issue_test.go
@@ -537,29 +537,6 @@ func TestGetUserSetting(t *testing.T) {
 }
 
 func TestRouteAttachCommentToIssue(t *testing.T) {
-	api := &plugintest.API{}
-	api.On("GetPost", "error_post").Return(nil, &model.AppError{Id: "1"})
-	api.On("GetPost", "post_not_found").Return(nil, (*model.AppError)(nil))
-	api.On("GetPost", "valid_post").Return(&model.Post{
-		UserId: "userID",
-	}, nil)
-	api.On("GetPost", "0").Return(&model.Post{
-		UserId: "user_not_found",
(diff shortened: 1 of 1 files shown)


server/issue_test.go
```diff
@@ -312,54 +312,61 @@ func TestRouteShareIssuePublicly(t *testing.T) {
...
 		request      *requestStruct
 		expectedCode int
-	}{
+		setupMocks   func(api *plugintest.API)
+	}
+
+	successfulUserPostSetup := func(api *plugintest.API) {
+		api.On("GetPost", "1").Return(&model.Post{UserId: "1", ChannelId: "test_channel"}, (*model.AppError)(nil)).Once()
+		api.On("GetUser", "1").Return(&model.User{Username: "username"}, (*model.AppError)(nil)).Once()
+		api.On("GetChannelMember", mock.AnythingOfType("string"), mock.AnythingOfType("string")).Return(nil, nil).Once()
```
wiggin77: This should probably not return nil, nil.  It can return an empty ChannelMember instead.  
```suggestion
		api.On("GetChannelMember", mock.AnythingOfType("string"), mock.AnythingOfType("string")).Return(&model.ChannelMember{}, nil).Once()
```

abbas-dependable-naqvi:
> @abbas-dependable-naqvi I'm a bit concerned about the unit tests that were removed. Overall it appears coverage has been reduced.

@wiggin77 I might have messed them up while refactoring. I will add all the existing ones back along with the new one

PR #1240, opened 2025-11-05
Strip Go binaries to reduce bundle size

marianunez:
#### Summary

Add -ldflags="-s -w" to all Go build commands to strip debug symbols and DWARF information from binaries, reducing file sizes by ~20-30% while preserving panic stack traces.

Based on https://github.com/mattermost/mattermost-plugin-agents/pull/426

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,5 +21,8 @@ jobs:
       golang-version: "1.23.0"
 
   dist:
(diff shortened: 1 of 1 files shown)


nevyangelova:
Seems like CI is happy with the bundle size @marianunez @wiggin77

nevyangelova (approved):
Thanks @marianunez , approved!

PR #1246, opened 2025-11-24
fix: Jira post-action regressions and add signature verification

nevyangelova:
#### Summary
- restore `share publicly` and `transition` post-actions by allowing signed payloads to run even if the backing ephemeral post is gone
- generate and validate HMAC signatures for all Jira action buttons using the plugin secret
- keep rejecting unsigned payloads, so the security fix remains in place
- fix the malformed JSON response that caused the “invalid character” errors

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66721

### DIFF
--- a/server/http_test.go
+++ b/server/http_test.go
@@ -1462,7 +1462,7 @@ func TestShareIssuePubliclyAuthentication(t *testing.T) {
 	api.On("LogDebug", mockAnythingOfTypeBatch("string", 11)...).Return(nil)
 	api.On("LogWarn", mockAnythingOfTypeBatch("string", 10)...).Return(nil)
 	api.On("SendEphemeralPost", mock.Anything, mock.Anything).Return(&model.Post{})
-	api.On("GetPost", "attacker-post").Return(&model.Post{UserId: "attacker"}, (*model.AppError)(nil))
+	api.On("GetPost", "attacker-post").Return(&model.Post{UserId: "attacker", ChannelId: "channel-id"}, (*model.AppError)(nil))
 
 	p := &Plugin{}
 	p.SetAPI(api)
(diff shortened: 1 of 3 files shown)


server/issue_parser.go
```diff
@@ -32,14 +32,20 @@ func reporterSummary(reporter *jira.User) string {
 	return reporterSummary
 }
 
-func getActions(instanceID types.ID, client Client, issue *jira.Issue) ([]*model.PostAction, error) {
+func getActions(p *Plugin, instanceID types.ID, client Client, issue *jira.Issue) ([]*model.PostAction, error) {
 	var actions []*model.PostAction
 
 	ctx := map[string]interface{}{
 		"issue_key":   issue.ID,
```
wiggin77: Strange that the map key is called `issue_key`, which I assume is the "MM-1000" format, but what is being assigned is a numeric id.

server/issue.go
```diff
@@ -85,40 +130,23 @@ func decodePostActionRequest(r *http.Request) (string, model.PostActionIntegrati
 	return authenticatedUserID, requestData, 0, nil
 }
 
-func (p *Plugin) buildPostActionContext(authenticatedUserID string, requestData model.PostActionIntegrationRequest) (*postActionContext, int, string) {
+func (p *Plugin) buildPostActionContext(authenticatedUserID string, requestData model.PostActionIntegrationRequest, requireStoredPost bool) (*postActionContext, int, string) {
```
marianunez: Do we need this new parameter `requireStoredPost`? It seems that both instances that it's used it's `false`?

marianunez (commented):
Thanks @nevyangelova! Left a non-blocking comment

marianunez (approved):

PR #1252, opened 2025-12-04
Improve Jira create issue UX, team field handling, and slash cmd auto…

nevyangelova:
#### Summary
- render every field Jira exposes in create-metadata by dropping the hardcoded allow lists and adding native inputs for numeric,date and datetime types, and keeping required-field checks in sync with what the UI can actually display
- detect Team custom fields by schema, cache their IDs per instance, and reuse them when writing issues instead of assuming customfield_10001
- advertise `/jira issue create` in slash-command autocomplete so the existing modal workflow is discoverable
- address package vulnerability updates using 'npm audit' 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66787

### DIFF
--- a/server/subscribe.go
+++ b/server/subscribe.go
@@ -180,7 +180,7 @@ func (p *Plugin) getUserID() string {
 	return p.getConfig().botUserID
 }
 
-func (p *Plugin) matchesSubsciptionFilters(wh *webhook, instanceID types.ID, filters SubscriptionFilters) bool {
+func (p *Plugin) matchesSubscriptionFilters(wh *webhook, instanceID types.ID, filters SubscriptionFilters) bool {
 	webhookEvents := wh.Events()
 	foundEvent := false
 	eventTypes := filters.Events
@@ -312,7 +312,7 @@ func (p *Plugin) getChannelsSubscribed(wh *webhook, instanceID types.ID) ([]Chan
 	subscriptionMap := make(map[string]bool)
 	subIds := subs.Channel.ByID
 	for _, sub := range subIds {
-		if p.matchesSubsciptionFilters(wh, instanceID, sub.Filters) {
+		if p.matchesSubscriptionFilters(wh, instanceID, sub.Filters) {
 			if !subscriptionMap[sub.ChannelID] {
 				subscriptionMap[sub.ChannelID] = true
 				channelSubscriptions = append(channelSubscriptions, sub)
--- a/webapp/src/components/input.jsx
+++ b/webapp/src/components/input.jsx
@@ -147,7 +147,6 @@ export default class Input extends PureComponent {
                     type={this.props.type}
                     placeholder={this.props.placeholder}
                     value={value}
-                    maxLength={this.props.maxLength}
                     onChange={this.handleChange}
                     disabled={this.props.disabled}
                     readOnly={this.props.readOnly}


wiggin77 (changes requested):
Looks great.  One race condition to fix.

server/team_fields.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	p.teamFieldCache[instanceID] = current
+}
+
+func (p *Plugin) getTeamFieldKeys(instanceID types.ID) map[string]struct{} {
+	p.teamFieldCacheLock.RLock()
+	cached := p.teamFieldCache[instanceID]
+	if len(cached) == 0 {
+		p.teamFieldCacheLock.RUnlock()
```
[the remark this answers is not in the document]
nevyangelova: You are right and also defer is safer, thanks @wiggin77 . Pushed a fix and added a comment.

server/issue.go
```diff
@@ -539,9 +548,9 @@ func (p *Plugin) GetCreateIssueMetadataForProjects(instanceID, mattermostUserID
 		return nil, err
 	}
 
-	teamIDList := p.conf.TeamIDList
-	if len(teamIDList) > 0 {
-		injectTeamAllowedValues(metaInfo, teamIDList)
+	teamFieldKeys := injectTeamAllowedValues(metaInfo, p.conf.TeamIDList)
+	if len(teamFieldKeys) > 0 {
+		p.cacheTeamFieldKeys(instanceID, teamFieldKeys)
 	}
```
marianunez: In the cases where the JIRA instance does not have team fields, if we skip the cache, it seems it will default to `customfield_10001`. It seems that this may cause `prreProcessTeamFields` and `updateTeamValue` to try to process with a non-existint field key. 

Just double checking that is the behavior we want as I"m not that familiar with the code here

server/issue.go
```diff
@@ -576,15 +587,27 @@ func injectTeamAllowedValues(metaInfo *jira.CreateMetaInfo, teamIDList []TeamLis
...
 				}
 
-				if schemaRaw["custom"] != "com.atlassian.jira.plugin.system.customfieldtypes:atlassian-team" {
+				if schemaRaw["custom"] != teamFieldSchema {
 					continue
 				}
 
-				fieldMap["allowedValues"] = allowedValues
-				issueType.Fields[key] = fieldMap
+				lowerKey := strings.ToLower(key)
```
marianunez: Why do we change to lowercase ? Are we also doing the same when looking up to avoid casing mismatchs?

marianunez:
@nevyangelova can you provide QA Steps on how to test the team field? In the spinwick I connected it to mmtest Jira instance and the team field doesn't show in the Create Issue.

nevyangelova:
> @nevyangelova can you provide QA Steps on how to test the team field? In the spinwick I connected it to mmtest Jira instance and the team field doesn't show in the Create Issue.

Did you configure the team id in plugin settings because Jira's API doesn't include team data in the create-metadata response?

1. Go to System Console → Plugins → Jira
2. In "Team List" setting, add your teams: `[Team Name](team-id),[Another Team](team-id-2)`
3. Save and try /jira create again

To get team IDs from mmtest Jira: Settings → Teams → click a team → ID is in the URL.

nevyangelova:
@marianunez just waiting on your review and we can merge 🎉

marianunez (approved):

PR #1254, opened 2025-12-08
fix: Duplicate Jira comment DMs and skip self-watcher alerts

nevyangelova:
#### Summary
- Filter out comment_* events entirely so we only send one DM per Jira comment even when Cloud fires `jira:issue_updated` and `comment` webhooks
- prevent watchers from receiving a DM when they’re also the author of the comment

#### Jira discussion and next steps
- Ben confirmed we can rely on `jira:issue_updated` for comments and recommend leaving the standalone `comment_*` webhooks disabled to avoid duplicate payloads.
- Watcher/reporter DMs already exist but are opt-in. That covers the customer’s point #3 as a communication item rather than new code.
- Personal DMs for non-comment field changes (point #4) remain out of scope for this PR and I am waiting on Ben’s call whether we should spin that out as a separate feature request.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66750
https://mattermost.atlassian.net/browse/MM-66798

### DIFF
--- a/server/webhook_http.go
+++ b/server/webhook_http.go
@@ -98,11 +98,6 @@ func (p *Plugin) httpWebhook(w http.ResponseWriter, r *http.Request, instanceID 
 		return respondErr(w, http.StatusBadRequest, err)
 	}
 
-	// Skip channel posts for standalone comment events to avoid duplicates
-	if wh.ShouldSkipChannelPost() {
-		return http.StatusOK, nil
-	}
-
 	// Skip events we don't need to post
 	if selectedEvents.Intersection(wh.Events()).Len() == 0 {
 		return http.StatusOK, nil
--- a/server/webhook_parser.go
+++ b/server/webhook_parser.go
@@ -76,25 +76,10 @@ func ParseWebhook(bb []byte) (wh Webhook, err error) {
 		}
 	case commentCreated:
 		wh, err = parseWebhookCommentCreated(jwh)
-		if err == nil {
-			if webhook, ok := wh.(*webhook); ok && webhook != nil {
-				webhook.skipChannelPost = true
-			}
-		}
 	case commentUpdated:
 		wh, err = parseWebhookCommentUpdated(jwh)
-		if err == nil {
-			if webhook, ok := wh.(*webhook); ok && webhook != nil {
-				webhook.skipChannelPost = true
-			}
-		}
 	case commentDeleted:
 		wh, err = parseWebhookCommentDeleted(jwh)
-		if err == nil {
-			if webhook, ok := wh.(*webhook); ok && webhook != nil {
-				webhook.skipChannelPost = true
-			}
-		}
 	case worklogUpdated:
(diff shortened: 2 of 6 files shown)


server/issue.go
```diff
@@ -1640,3 +1652,19 @@ func (p *Plugin) GetProjectListWithAPIToken(instanceID string) (*jira.ProjectLis
 
 	return &projectResponse.Values, nil
 }
+
+func shouldNotifyWatcherUser(watcher jira.Watcher, author *jira.User) bool {
+	if author == nil {
+		return true
+	}
```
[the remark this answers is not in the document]
nevyangelova: Returning an error adds complexity for little benefit. if author data is missing we can't skip the watcher reliably so we default to notifying. It's a safe fallback.

nevyangelova:
@hanzei ready for review

server/webhook_parser.go
```diff
@@ -74,12 +74,8 @@ func ParseWebhook(bb []byte) (wh Webhook, err error) {
 		default:
 			wh, err = parseWebhookUnspecified(jwh)
 		}
-	case commentCreated:
-		wh, err = parseWebhookCommentCreated(jwh)
-	case commentUpdated:
-		wh, err = parseWebhookCommentUpdated(jwh)
-	case commentDeleted:
-		wh, err = parseWebhookCommentDeleted(jwh)
+	case commentCreated, commentUpdated, commentDeleted:
```
hanzei: Let's ensure in testing that this change doesn't have any unexpected side affects.

server/webhook_http.go
```diff
@@ -98,6 +98,11 @@ func (p *Plugin) httpWebhook(w http.ResponseWriter, r *http.Request, instanceID
 		return respondErr(w, http.StatusBadRequest, err)
 	}
 
+	// Skip channel posts for standalone comment events to avoid duplicates
+	if wh.ShouldSkipChannelPost() {
+		return http.StatusOK, nil
+	}
```
hanzei: 3/5 that the API should not show about business logic. Can we handle that somewhere else?

nevyangelova:
@hanzei kind reminder as we would like to roll out this fix as soon as possible

server/webhook_parser.go
```diff
@@ -230,17 +239,6 @@ func parseWebhookDeleted(jwh *JiraWebhook) Webhook {
...
-	// The "comment_xxx" events from Jira Server come incomplete,
-	// i.e. with just minimal metadata. We toss them out since they
-	// are rather useless for our use case. Instead the Jira server
-	// webhooks receive and process jira:issue_updated with eventTypes
-	// "issue_commented", etc.
-	//
-	// Detect this condition by checking that jwh.Issue.ID
-	if jwh.Issue.ID == "" {
-		return nil, ErrWebhookIgnored
-	}
```
marianunez: Were these check removed intentionally?

PR #1258, opened 2025-12-30
fix: Show file attachment info in Attach to Jira Issue modal

nevyangelova:
#### Summary
Fixes the "Attach to Jira Issue" modal showing a blank message when the user is attaching a file only post by adding getPostDisplayMessage() that returns the post message or falls back to displaying file attachment names. 

Single file: "Attached file: document.pdf"
Multiple files: "Attached files: image.png, doc.pdf"

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66431


ogi-m:
@nevyangelova I don't see any text in the modal when it's just an attachment-only post without any text:

image

nevyangelova:
@ogi-m updated

PR #1259, opened 2025-12-30
fix: Attachment filenames showing as ~~~~ in subscription notifications

nevyangelova:
#### Summary
Subscription notifications show ~~~~ instead of attachment filenames. I fixed parseWebhookUpdatedAttachments to populate fieldInfo.from and fieldInfo.to values. When multiple field updates are merged attachment names now display correctly.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66432


ogi-m:
@nevyangelova Could we get rid of the crossed out `none`:

image

PR #1262, opened 2026-01-07
MM-63225: Add support for Advanced Roadmaps Team field

nevyangelova:
#### Summary
Adds support for the Advanced Roadmaps "Team" field type in Jira subscriptions and the Create Issue modal. Customers using Jiras Advanced Roadmaps module have a "Team" field with schema type `com.atlassian.teams:rm-team-custom-field-team` which was not recognized by the plugin. Only the standard team field was supported.

- Added `TEAM_ADVANCED_ROADMAPS` enum for the Advanced Roadmaps team field schema
- Updated `isTeamField()` to recognize both team field types
- Updated server-side team field detection to accept both schemas

## Ticket
https://mattermost.atlassian.net/browse/MM-63225

### DIFF
--- a/server/issue.go
+++ b/server/issue.go
@@ -92,9 +92,9 @@ const (
 
 	expandValueGroups = "groups"
 
-	teamFieldSchema             = "com.atlassian.jira.plugin.system.customfieldtypes:atlassian-team"
-	teamAdvancedRoadmapsSchema  = "com.atlassian.teams:rm-team-custom-field-team"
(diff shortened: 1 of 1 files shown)


nevyangelova:
@ogi-m Is the TeamIDList configured in the plugin settings? If no teams are configured, allowedValues will be empty and the field won't render properly

ogi-m:
Thanks @nevyangelova , after configuring it, the team field shows up in create issue modal and set up team filtering. Confirmed the filter works as expected as well.

PR #1266, opened 2026-01-29
MM-67247: Show display name instead of account ID for unconnected users

nevyangelova:
#### Summary
Fixes notification messages showing raw `~accountid:xxx` format when the mentioned Jira user is not connected to Mattermost so I modify replaceJiraAccountIds() to fetch display name from Jira API when user is not connected to Mattermost. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-67247

#### QA Steps
In Jira, add a comment that mentions a user who is not connected to Mattermost and check the notification in Mattermost.
Expected: Shows display name instead of `~accountid:24842824174714`

### DIFF
--- a/server/utils_test.go
+++ b/server/utils_test.go
@@ -140,8 +140,8 @@ func (m *mockJiraClient) RESTPostAttachment(issueID string, data io.Reader, name
 	return nil, nil
 }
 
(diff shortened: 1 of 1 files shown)


marianunez (changes requested):

marianunez (approved):
Thanks @nevyangelova !

PR #1267, opened 2026-01-29
MM-67214: Upgrade to React 18 and migrate tests to React Testing Library

nevyangelova:
#### Summary
Upgraded React from 16.14.0 to 18.2.0
Upgraded react-dom from 16.8.6 to 18.2.0
Upgraded react-redux from 7.2.9 to 8.1.3
Upgraded react-select from 3.2.0 to 5.10.0
Upgraded react-intl from 4.5.0 to 6.8.0
Removed Enzyme and migrated all tests to React Testing Library.
Updated Babel configuration for React 18 automatic JSX runtime.
Fixed react-select type imports for v5 compatibility. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-67214

### DIFF
--- a/webapp/package-lock.json
+++ b/webapp/package-lock.json
@@ -71,7 +71,7 @@
         "sass": "1.60.0",
         "sass-loader": "12.2.0",
         "style-loader": "3.3.1",
-        "webpack": "5.98.0",
+        "webpack": "5.105.0",
         "webpack-cli": "5.1.4"
       }
     },
@@ -84,8 +84,6 @@
     },
     "node_modules/@ampproject/remapping": {
       "version": "2.3.0",
-      "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.3.0.tgz",
-      "integrity": "sha512-30iZtAPgz+LTIYoeivqYo853f02jBYSd5uGnGpkFV0M3xOt9aN73erkgYAmZU43x4VfqcnLxW9Kpg3R5LC4YYw==",
       "dev": true,
       "license": "Apache-2.0",
       "dependencies": {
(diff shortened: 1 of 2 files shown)


webapp/src/components/data_selectors/jira_epic_selector/jira_epic_selector.test.tsx
```diff
@@ -25,43 +57,45 @@ describe('components/JiraEpicSelector', () => {
...
+        await act(async () => {
+            renderWithRedux(
+                <JiraEpicSelector
+                    {...props}
+                    ref={ref}
+                />,
+            );
+        });
+
+        expect(ref.current).toBeDefined();
```
marianunez: Should we check other elements directly? Seems that just checking that the component exists is lowering the quality of the test compared to snapshot 

This applies to all cases below where the snapshot has been removed.

nevyangelova: The issues are with internal validation that runs when the component is fully mounted. This is done when validator collects validation functions from child components. The issue is that with deep rendering child components register their validators and when handleCreate is called validation fails because the child components aren't fully set up.

The tests were designed for shallow rendering which doesn't run full lifecycle. I would propose to add new tests as we develop new features and fix bugs, as direct rewrite would be very difficult and time consuming atm. but if you think it can be prioritised I can do it :)

webapp/src/components/modals/channel_subscriptions/channel_subscriptions.test.tsx
```diff
@@ -2,20 +2,58 @@
...
 
 describe('components/ChannelSettingsModal', () => {
+    const mockTheme = {
+        centerChannelColor: '#333333',
+        centerChannelBg: '#ffffff',
+        buttonBg: '#166de0',
+        buttonColor: '#ffffff',
+        linkColor: '#2389d7',
+        errorTextColor: '#fd5960',
+    };
```
avasconcelos114: Nit: This is another thing that could be a constant provided by test utils to reduce repetition

avasconcelos114 (approved):
Things are looking quite tidy now :) I think this leaves us in a good spot to continue bolstering the testing suite as we make new contributions

PR #1297, opened 2026-04-13
Validate subscription ownership before editing

nevyangelova:
#### Summary
Load the existing subscription from the store and verify the user has permission to manage it before applying any changes.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-68272

### DIFF
--- a/server/http_test.go
+++ b/server/http_test.go
@@ -636,9 +636,22 @@ func TestEditSubscription(t *testing.T) {
 			expectedStatusCode: http.StatusBadRequest,
 		},
 		"No Permissions": {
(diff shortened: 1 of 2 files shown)


nevyangelova:
@edgarbellot is this the fix you had in mind?

avasconcelos114 (approved):
The logic LGTM! It might be good to also add a test case for when the user lacks permissions for the target channel

ogi-m (approved):
LGTM! 
- User B without access to the channel gets blocked with 403 because the server checks permissions against the original channel (User A's), not the channel User B supplied in the request body
- Fabricated subscription ID correctly rejected with 400
- User A can still edit their own subscription normally, no regression

edgarbellot:
@nevyangelova yep, this correctly fixes the vuln. Thank you!

PR #1302, opened 2026-04-23
Add install token verification to cloud instance lifecycle callback

nang2049:
#### Summary
Add a per installation cryptographic token to the Atlassian Connect  installed lifecycle callback URL. The token is generated when the cloud instance is created and verified when the callback is received.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-68377

Verified as sysadmin by run /jira instance install cloud. from my machine ran: 
```
curl -X POST -H "Content-Type: application/json" \
  -d '{"key":"com.mattermost.jira","clientKey":"attacker","sharedSecret":"attacker_secret","baseUrl":"https://yourorg.atlassian.net","productType":"jira","eventType":"installed"}' \
  https://<MATTERMOST_URL>/plugins/jira/ac/installed
  ```
and got: `403 Forbidden with "invalid install token".`

## Change Impact: 🔴 High

**Reasoning:** The PR adds security-critical install-token verification and changes the cloud installation flow, routing, and persisted state (KV-backed pending routes and altered function signatures), expanding the blast radius across instance resolution, installation handlers, and storage APIs.  
**Regression Risk:** High. The changes touch authentication/installation handlers, instance resolution, storage interfaces, and routing; they introduce new TTL-backed pending routes, path-scoped routing, and function signature changes that can produce integration regressions, race conditions, and untested edge cases.  
** QA Recommendation:** Comprehensive manual QA is strongly recommended: validate successful and failed install callbacks (valid/invalid/missing tokens), instance-scoped vs global routing behavior, pending-route TTL/expiry and concurrent install scenarios, deletion/cleanup of pending routes after install, and end-to-end integration with Atlassian Cloud. Skipping manual QA is not advised due to security and stability implications.

*Generated by CodeRabbitAI*

### DIFF
--- a/server/atlassian_connect.go
+++ b/server/atlassian_connect.go
@@ -4,6 +4,7 @@
 package main
 
 import (
+	"crypto/hmac"
 	"encoding/json"
 	"io"
 	"net/http"
@@ -16,10 +17,21 @@ import (
 const userRedirectPageKey = "user-redirect"
 
 func (p *Plugin) httpACJSON(w http.ResponseWriter, r *http.Request, instanceID types.ID) (int, error) {
+	instance, err := p.instanceStore.LoadInstance(instanceID)
+	if err != nil {
+		return respondErr(w, http.StatusInternalServerError,
+			errors.WithMessage(err, "failed to load instance"))
+	}
+
+	installRoute := routeACInstalled
+	if ci, ok := instance.(*cloudInstance); ok && ci.InstallToken != "" {
+		installRoute = routeACInstalled + "?token=" + ci.InstallToken
+	}
+
 	return p.respondTemplate(w, r, "application/json", map[string]string{
 		"BaseURL":                      p.GetPluginURL(),
 		"RouteACJSON":                  instancePath(routeACJSON, instanceID),
-		"RouteACInstalled":             routeACInstalled,
+		"RouteACInstalled":             installRoute,
 		"RouteACUninstalled":           routeACUninstalled,
 		"RouteACUserRedirectWithToken": instancePath(routeACUserRedirectWithToken, instanceID),
 		"UserRedirectPageKey":          userRedirectPageKey,
@@ -62,6 +74,14 @@ func (p *Plugin) httpACInstalled(w http.ResponseWriter, r *http.Request) (int, e
 	if ci.Installed {
 		return respondErr(w, http.StatusForbidden,
 			errors.Errorf("Jira instance %s is already installed", asc.BaseURL))
+	}
+
+	if ci.InstallToken != "" {
+		providedToken := r.URL.Query().Get("token")
+		if !hmac.Equal([]byte(providedToken), []byte(ci.InstallToken)) {
+			return respondErr(w, http.StatusForbidden,
+				errors.New("invalid install token"))
+		}
 	}
 
 	// Create a permanent instance record, also store it as current
--- a/server/atlassian_connect.go
+++ b/server/atlassian_connect.go
@@ -65,7 +65,7 @@ func (p *Plugin) httpACInstalledGlobal(w http.ResponseWriter, r *http.Request) (
 	return p.processACInstalled(w, r, "", false)
 }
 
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: server/atlassian_connect.go [hunk 1]
```diff
@@ -64,6 +76,14 @@ func (p *Plugin) httpACInstalled(w http.ResponseWriter, r *http.Request) (int, e
 			errors.Errorf("Jira instance %s is already installed", asc.BaseURL))
 	}
 
+	if ci.InstallToken != "" {
+		providedToken := r.URL.Query().Get("token")
+		if !hmac.Equal([]byte(providedToken), []byte(ci.InstallToken)) {
+			return respondErr(w, http.StatusForbidden,
+				errors.New("invalid install token"))
+		}
+	}
```
avasconcelos114: Did we want to clear `ci.InstallToken` once we have run this validation?

avasconcelos114 (commented):
The fix looks good! Only had one general question I wanted to confirm

server/atlassian_connect.go
nang2049: Good catch, but the token is already effectively dead after this point though as the new instance gets Installed: true so any future call gets rejected before it even reaches the token check. and the inactive instance expires after 15 min anyway!

server/atlassian_connect.go
```diff
@@ -16,10 +17,21 @@ import (
...
 func (p *Plugin) httpACJSON(w http.ResponseWriter, r *http.Request, instanceID types.ID) (int, error) {
+	instance, err := p.instanceStore.LoadInstance(instanceID)
+	if err != nil {
+		return respondErr(w, http.StatusInternalServerError,
+			errors.WithMessage(err, "failed to load instance"))
+	}
+
+	installRoute := routeACInstalled
+	if ci, ok := instance.(*cloudInstance); ok && ci.InstallToken != "" {
+		installRoute = routeACInstalled + "?token=" + ci.InstallToken
```
edgarbellot: Nice approach with the install token.

One concern though: could the token be extracted by an unauthenticated caller before the legitimate install completes? The [atlassian-connect.json endpoint](https://github.com/mattermost/mattermost-plugin-jira/blob/965d763/server/http.go#L129) (129) is served without `checkAuth`, and the token gets embedded directly in the response via the [installRoute construction](https://github.com/mattermost/mattermost-plugin-jira/blob/965d763/server/atlassian_connect.go#L26-L28) (26-28). Since the instance ID in the URL is just [base64(jiraURL)](https://github.com/mattermost/mattermost-plugin-jira/blob/965d763/server/http.go#L275-L277) (275-277), an attacker who knows the Jira URL (often guessable as `<company>.atlassian.net`) could GET the descriptor, extract the token from `lifecycle.installed`, and use it to call `/ac/installed?token=<token>` with a crafted payload - all within the 15-minute window.

Note: the unauthenticated descriptor endpoint and the predictable URL scheme are pre-existing, but this PR introduces the token and places it in that response, which is what creates the leak path.

edgarbellot (changes requested):

server/atlassian_connect.go
avasconcelos114: Ah alrighty fair enough that keeping it is pretty low-risk, I had just figured that since it won't be really used from this point onwards, that we could just clean it up but I wouldn't consider this a must-do

avasconcelos114 (approved):

server/atlassian_connect.go
nang2049: Thanks @edgarbellot I removed the token from the public descriptor and switched pending Connect to an instance path with a short lived opaque Jira KV mapping, 404 the Jira‑URL descriptor route until install completes, require POST /ac/installed on that path with the path bound to SetupRoutingSecret and reject the global /ac/installed callback for pending installs. Is that what you had in mind? :D

edgarbellot: This approach works! Thank you @nang2049 for quickly addressing it :)

edgarbellot (approved):

nang2049:
@coderabbitai approve

PR #1303, opened 2026-04-27
MM-68049: Add AdminAPIToken fallback for watcher notifications

nang2049:
#### Summary

When a comment is added to a Jira issue, the plugin attempts to notify all watchers. To fetch the watcher list it needs a valid Jira API client, which it obtains from the first connected Mattermost user among the issue's Creator, Assignee, or Reporter. If none of these three users have connected accounts, the watcher list cannot be retrieved and notifications silently fail.

This PR adds a fallback: when no connected user is available, the plugin uses the **AdminAPIToken** (configured in plugin settings) to call `GET /rest/api/2/issue/{key}/watchers` directly via HTTP Basic auth. This is consistent with how the admin token is already used elsewhere in the plugin (e.g. `GetIssueDataWithAPIToken` for expanding issue data in Cloud OAuth webhooks).

**Changes:**
- **`GetWatchersWithAPIToken`** — New function that fetches the watcher list using the admin API token with Basic auth. Follows the same pattern as `GetIssueDataWithAPIToken`. Works with both Jira Cloud (`email:api-token`) and Server/Data Center (`username:password` or `username:PAT`).
- **`checkIssueWatchers`** — Modified to fall back to `GetWatchersWithAPIToken` when `fetchConnectedUser` returns no client and `AdminAPIToken` is configured. Also adds a safety check: if Jira returns a non-zero `watchCount` but an empty `watchers` array, the admin account likely lacks the "View voters and watchers" project permission, and a warning is logged.
- **Tests** — Added `TestGetWatchersWithAPIToken` covering Cloud (accountId-based), Server/DC (name-based), 404, and 403 scenarios using `httptest.NewServer`.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-68049

#### Manual QA Steps

1. **Setup:** Configure `AdminAPIToken` and `AdminEmail` in **System Console > Plugins > Jira > Admin API Token / Admin Email**. The Jira admin account must have "Browse projects" and "View voters and watchers" permissions.
2. **Create a Jira issue** where the Creator, Assignee, and Reporter are all users who have **not** connected their Jira accounts to Mattermost.
3. **Have at least one connected Mattermost user watch the issue** in Jira (and ensure their watching notification setting is enabled in `/jira settings`).
4. **Add a comment** on the issue from any Jira user.
5. **Verify** the watching Mattermost user receives a DM notification with the comment content and "You are watching this issue" message.
6. **Check server logs** for the info message: `No connected users found for watcher lookup, falling back to admin API token`.
7. **Negative test:** Remove the `AdminAPIToken` from plugin settings and repeat steps 2-4. Verify no watcher notification is sent and an error is logged: `error while fetching connected users for the instanceID`.
8. **Permission test:** Configure an admin token for a Jira account that lacks "View voters and watchers" permission. Repeat steps 2-4. Verify a warning is logged about non-zero watch count with empty watcher list.

## Change Impact: 🟡 Medium

**Reasoning:** Adds an admin-API-token fallback for fetching Jira issue watchers when no connected Mattermost user provides a Jira client. The change is localized to watcher notification logic but introduces an alternate authentication flow and additional REST calls that affect a user-facing notification feature.

**Regression Risk:** Low–Medium. The primary connected-user behavior remains unchanged and the fallback is conditional, but the new admin-token path and added HTTP requests can surface permission-related failures or unexpected Jira responses (e.g., watchCount vs watchers list) that may impact notification delivery.

** QA Recommendation:** Moderate manual QA recommended. Validate: (1) connected users present (no regression), (2) no connected users + AdminAPIToken configured (fallback triggers and returns watchers), (3) no AdminAPIToken configured (graceful behavior), and (4) admin account lacking "View voters and watchers" permission (confirm warning logs and behavior). Unit tests cover Cloud/Server-DC success and 404/403 cases; run integration/end-to-end tests for notification flows.
*Generated by CodeRabbitAI*

### DIFF
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -213,24 +213,36 @@ func (p *Plugin) OnConfigurationChange() error {
 		}
 	}
 
-	jsonBytes, err := json.Marshal(ec.AdminAPIToken)
-	if err != nil {
-		p.client.Log.Warn("Error marshaling the admin API token", "error", err.Error())
-		return err
+	prevExternal := p.getConfig().externalConfig
+
+	switch {
+	case ec.AdminAPIToken == "" || ec.AdminAPIToken == model.FakeSetting:
+		ec.AdminAPIToken = prevExternal.AdminAPIToken
+	case ec.AdminAPIToken == prevExternal.AdminAPIToken:
+	default:
+		if _, decErr := decrypt([]byte(ec.AdminAPIToken), []byte(ec.EncryptionKey)); decErr == nil {
+			break
+		}
+		jsonBytes, err := json.Marshal(ec.AdminAPIToken)
+		if err != nil {
+			p.client.Log.Warn("Error marshaling the admin API token", "error", err.Error())
+			return err
+		}
+		if ec.EncryptionKey == "" {
+			p.client.Log.Warn("Encryption key required to encrypt admin API token")
+			return errors.New("failed to encrypt admin token. Encryption key not generated")
+		}
+		encryptedAdminAPIToken, err := encrypt(jsonBytes, []byte(ec.EncryptionKey))
+		if err != nil {
+			p.client.Log.Warn("Error encrypting the admin API token", "error", err.Error())
+			return err
+		}
+		ec.AdminAPIToken = string(encryptedAdminAPIToken)
 	}
 
-	encryptionKey := ec.EncryptionKey
-	if ec.AdminAPIToken != "" && encryptionKey == "" {
-		p.client.Log.Warn("Encryption key required to encrypt admin API token")
-		return errors.New("failed to encrypt admin token. Encryption key not generated")
+	if ec.AdminEmail == "" {
+		ec.AdminEmail = prevExternal.AdminEmail
 	}
-
-	encryptedAdminAPIToken, err := encrypt(jsonBytes, []byte(encryptionKey))
-	if err != nil {
-		p.client.Log.Warn("Error encrypting the admin API token", "error", err.Error())
-		return err
-	}
-	ec.AdminAPIToken = string(encryptedAdminAPIToken)
 
 	// Default to 30 days if not set
 	if ec.ThreadedJiraCommentSubscriptionDuration == "" {
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -216,8 +216,9 @@ func (p *Plugin) OnConfigurationChange() error {
 	prevExternal := p.getConfig().externalConfig
 
 	switch {
-	case ec.AdminAPIToken == "" || ec.AdminAPIToken == model.FakeSetting:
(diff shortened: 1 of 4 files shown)


avasconcelos114 (approved):
Changes LGTM! Though I'd +1 the minor comment to change the wording to not necessarily specify that it was an issue with the admin account

nang2049:
@avasconcelos114 while testing the watcher fallback I hit a second bug that blocked it on Cloud:

OnConfigurationChange was unconditionally reencrypting AdminAPIToken on every save. When the System Console submits the secret field as the ******** that placeholder was getting encrypted and persisted which broke SetAdminAPITokenRequestHeader on every subsequent reload. Same wipe for AdminEmail. Now we preserve the previous value when inbound is empty and only reencrypt when inbound is actually a new token. Please re-review

server/plugin.go
```diff
@@ -213,24 +213,36 @@ func (p *Plugin) OnConfigurationChange() error {
...
-	}
-
-	encryptionKey := ec.EncryptionKey
-	if ec.AdminAPIToken != "" && encryptionKey == "" {
-		p.client.Log.Warn("Encryption key required to encrypt admin API token")
-		return errors.New("failed to encrypt admin token. Encryption key not generated")
+	prevExternal := p.getConfig().externalConfig
+
+	switch {
+	case ec.AdminAPIToken == "" || ec.AdminAPIToken == model.FakeSetting:
```
avasconcelos114: The `ec.AdminAPIToken == "" ||` condition makes it seem like if we are trying to clear the AdminAPIToken, that the plugin will ignore that and keep the previously stored value, is that the case here?

server/plugin.go
```diff
@@ -213,24 +213,36 @@ func (p *Plugin) OnConfigurationChange() error {
...
+			return err
+		}
+		ec.AdminAPIToken = string(encryptedAdminAPIToken)
 	}
 
-	encryptedAdminAPIToken, err := encrypt(jsonBytes, []byte(encryptionKey))
-	if err != nil {
-		p.client.Log.Warn("Error encrypting the admin API token", "error", err.Error())
-		return err
+	if ec.AdminEmail == "" {
```
avasconcelos114: Somewhat related to the other comment, but I'm guessing that we want to keep the `prevExternal` values of `AdminEmail` in the same scenario that we are keeping the AdminAPIToken as they are essentially a pair? (which would mean we may also want to evaluate for `ec.AdminAPIToken == model.FakeSetting` here)

avasconcelos114 (commented):
Whoa great job finding that out! 

Your approach to resolving it sounds good to me, just have one thing to confirm

nang2049:
@avasconcelos114 updated

avasconcelos114:
@nang2049 LGTM! Thank you :)

PR #1321, opened 2026-06-30
MM-69458 Improving notification deduplication logic

avasconcelos114:
#### Summary

This PR is meant to pay some technical debt that was identified during a review in a backport PR. This is aimed to make 3 general improvements:
1. Moving encryption guard up to prevent mistakenly treating the raw admin API token as an encrypted value
2. Adding the instance ID to the dedup notification logic so we also take instance installations into consideration when evaluating duplicates
3. Ensuring parallel operations can't bypass the dedup flow

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-69458

## Change Impact: 🟠 Medium

**Reasoning:** The PR hardens admin API token handling (encryption guard ordering) and materially changes webhook notification de-duplication semantics under concurrency (atomic claim, instance-scoped keys, longer TTL), both of which affect correctness of sensitive flows but remain contained to server-side plugin/webhook logic.  
**Regression Risk:** Moderate. Token decryption/config-path behavior and user-facing DM notification delivery/dedup behavior are both impacted; however, the change is localized, has accompanying test updates for the dedup key, and rollback should be straightforward (revert server/plugin/webhook logic).  
**QA Recommendation:** Run targeted manual QA focused on (1) admin API token configuration cases (with/without encryption key) and failure logging/error behavior, and (2) webhook notification delivery under concurrent deliveries to confirm duplicates are suppressed only within the intended instance scope. Skipping manual QA is not recommended for these user-visible correctness paths.
*Generated by CodeRabbitAI*

### DIFF
--- a/server/webhook.go
+++ b/server/webhook.go
@@ -220,22 +220,30 @@ func (wh *webhook) PostNotifications(p *Plugin, instanceID types.ID) ([]*model.P
 
 		notification.message = p.replaceJiraAccountIds(instance.GetID(), notification.message, client)
 
-		dedupKey := notificationDedupKey(wh, mattermostUserID, notification.message)
-		var alreadySent bool
-		if kvErr := p.client.KV.Get(dedupKey, &alreadySent); kvErr != nil {
-			p.client.Log.Warn("PostNotifications: failed to read dedup key, sending notification anyway", "key", dedupKey, "error", kvErr.Error())
-		}
-		if alreadySent {
+		// Atomically claim the dedup key before posting so concurrent webhook
+		// deliveries can't both pass the check and send duplicates. SetAtomic(nil)
+		// only writes when the key does not already exist.
+		dedupKey := notificationDedupKey(instance.GetID(), wh, mattermostUserID, notification.message)
+		claimed, kvErr := p.client.KV.Set(dedupKey, true, pluginapi.SetExpiry(notificationDedupTTL), pluginapi.SetAtomic(nil))
+		switch {
+		case kvErr != nil:
+			// Fail open: send the notification rather than dropping it.
+			p.client.Log.Warn("PostNotifications: failed to claim dedup key, sending notification anyway", "key", dedupKey, "error", kvErr.Error())
+		case !claimed:
+			// Another delivery already claimed this notification.
 			continue
 		}
 
 		post, err := p.CreateBotDMPost(instance.GetID(), mattermostUserID, notification.message, notification.postType)
 		if err != nil {
 			p.errorf("PostNotifications: failed to create notification post, err: %v", err)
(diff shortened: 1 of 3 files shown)


server/webhook.go
```diff
@@ -220,23 +220,31 @@ func (wh *webhook) PostNotifications(p *Plugin, instanceID types.ID) ([]*model.P
...
+		case !claimed:
+			// Another delivery already claimed this notification.
 			continue
 		}
 
 		post, err := p.CreateBotDMPost(instance.GetID(), mattermostUserID, notification.message, notification.postType)
 		if err != nil {
 			p.errorf("PostNotifications: failed to create notification post, err: %v", err)
+			// Release the claim so the notification can be retried.
+			if claimed {
```
nang2049: CreateBotDMPost can persist the post and still return an error. Releasing the claim in that case defeats the point of this PR as retry will duplicate. Suggest not releasing on error and letting the TTL expire the key.

```
post, err := p.CreateBotDMPost(instance.GetID(), mattermostUserID, notification.message, notification.postType)
if err != nil {
    p.errorf("PostNotifications: failed to create notification post, err: %v", err)
    // Dont release the dedup claim
    // TTL will clear the claim
    continue
}
```

server/webhook.go
```diff
@@ -250,8 +258,9 @@ func newWebhook(jwh *JiraWebhook, eventType string, format string, args ...inter
 	}
 }
 
-func notificationDedupKey(wh *webhook, recipientID types.ID, message string) string {
-	raw := fmt.Sprintf("%s_%s_%s",
+func notificationDedupKey(instanceID types.ID, wh *webhook, recipientID types.ID, message string) string {
+	raw := fmt.Sprintf("%s_%s_%s_%s",
+		string(instanceID),
```
nang2049, line 260: Including instanceID fixes cross instance key collisions. Good catch :))

nang2049 (changes requested):
Thanks @avasconcelos114 just one minor improvement!

nang2049 (approved):

avasconcelos114:
@ogi-m This should be ready for testing now!

If possible it would be great to have this tested on an HA environment as some reports show that having an `issue_updated` and `issue_assigned` both checked in the Jira webhooks will result in a duplicate on an HA environment when each of these 2 events are processed in different nodes

ogi-m (approved):
LGTM! 

We couldn't reliably reproduce the original issue so I tested by capturing a real Jira issue_assigned webhook payload, then replayed it 5 times simultaneously against the HA server's load balancer endpoint. All 5 requests returned 200 but only 1 DM was delivered, confirming the dedup works correctly across concurrent deliveries.

PR #1322, opened 2026-06-30
MM-69541: enforce channel membership in CreateIssue for both post_id and channel id paths

nang2049:
#### Summary
The membership check in CreateIssue was gated on post_id != "", leaving the direct channel_id path unauthorized. An authenticated user with a connected Jira account could post a notification into any channel, including private ones. The plugin API used to create the post bypasses Mattermost's channel ACLs, so membership must be enforced explicitly.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-69541

## Change Impact: High 🔴

**Reasoning:** This change closes a security/authorization gap in a core, user-facing path (`CreateIssue`) by enforcing Mattermost channel membership for both `post_id`-derived and direct `channel_id` inputs, preventing notification posts from being created in channels the user cannot access.

**Regression Risk:** Moderate: it intentionally changes behavior for previously-allowed (but unauthorized) requests involving private/inaccessible channels and for edge cases where channel resolution is ambiguous; however, the authorization failure cases are explicitly covered and now assert that elevated post-creation methods are not invoked.

**QA Recommendation:** Manual QA strongly recommended (small focused matrix): verify `CreateIssue` over HTTP and internal calls for both `post_id` and `channel_id` in public and private channels, confirm 403 responses for bypass attempts, and confirm that forbidden cases do not trigger `CreatePost` / `SendEphemeralPost`.
*Generated by CodeRabbitAI*

### DIFF
--- a/server/issue.go
+++ b/server/issue.go
@@ -360,10 +360,12 @@ func (p *Plugin) CreateIssue(in *InCreateIssue) (*jira.Issue, int, error) {
 		channelID = post.ChannelId
 	}
 
-	if channelID != "" {
-		if _, err = p.client.Channel.GetMember(channelID, in.mattermostUserID.String()); err != nil {
-			return nil, http.StatusForbidden, errors.New("User does not have access to this channel")
-		}
+	if channelID == "" {
+		return nil, http.StatusBadRequest, errors.New("either post_id or channel_id must be provided")
+	}
+
+	if _, err = p.client.Channel.GetMember(channelID, in.mattermostUserID.String()); err != nil {
+		return nil, http.StatusForbidden, errors.New("User does not have access to this channel")
 	}
 
 	for i, notCovered := range in.RequiredFieldsNotCovered {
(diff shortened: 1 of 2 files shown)


avasconcelos114 (approved):
LGTM!

server/issue.go
```diff
@@ -365,6 +360,12 @@ func (p *Plugin) CreateIssue(in *InCreateIssue) (*jira.Issue, int, error) {
 		channelID = post.ChannelId
 	}
 
+	if channelID != "" {
```
edgarbellot: Verified end-to-end. The original issue is correctly blocked and the happy path shows no regressions, great work!

Found one small edge case: when both `post_id` and `channel_id` are empty, the `if channelID != ""` guard skips the membership check and the request reaches the Jira API, creating an issue before the subsequent CreatePost fails with a 500. The result is an orphaned Jira issue with no Mattermost notification. No impact - nothing is written to any channel - but it may be worth validating that at least one of `post_id` or `channel_id` is present before calling the Jira API, to avoid creating issues the user will never see confirmed.

edgarbellot (approved):

server/issue.go
nang2049: @edgarbellot added the fix. Let me know if its good and ill merge :)

PR #1339, opened 2026-08-14
MM-70280 Applying dedup on channel subscription posts

avasconcelos114:
#### Summary

This PR applies the same dedup mechanism that exists for DM notifications and also applies them to channel subscriptions as well

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-70243

## Change Impact: 🟡 Medium

**Reasoning:** The change updates shared webhook delivery logic and KV-based deduplication for user-facing channel posts. Tests cover concurrent deliveries, overlapping subscriptions, separate channels, and deduplication key differences.

**Regression Risk:** Medium. The change affects notification delivery and KV persistence behavior. KV errors fail open, which limits service disruption.

**QA Recommendation:** Perform targeted manual QA for duplicate deliveries, overlapping subscriptions, separate channels, and KV failures. Skipping manual QA has moderate risk because notification behavior changes.
*Generated by CodeRabbitAI*

### DIFF
--- a/server/webhook.go
+++ b/server/webhook.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"net/url"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
@@ -21,8 +22,9 @@ import (
 )
 
 const (
-	notificationDedupTTL    = 30 * time.Second
+	webhookDedupTTL         = 30 * time.Second
 	notificationDedupKeyFmt = "notif_dedup_%s"
+	channelPostDedupKeyFmt  = "chan_dedup_%s"
 )
 
 const (
@@ -78,8 +80,13 @@ func (wh webhook) PostToChannel(p *Plugin, instanceID types.ID, channelID, fromU
 
 	if wh.headline == "" {
 		return nil, http.StatusBadRequest, errors.Errorf("unsupported webhook")
-	} else if pluginConfig.DisplaySubscriptionNameInNotifications && subscriptionName != "" {
-		wh.headline = fmt.Sprintf("%s\nSubscription: **%s**", wh.headline, subscriptionName)
+	}
+
+	// Keep the dedup identity independent of the subscription name so overlapping
+	// subscriptions on the same channel collapse into one post.
+	headline := wh.headline
+	if pluginConfig.DisplaySubscriptionNameInNotifications && subscriptionName != "" {
+		headline = fmt.Sprintf("%s\nSubscription: **%s**", headline, subscriptionName)
 	}
 
 	post := &model.Post{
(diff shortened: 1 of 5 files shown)


server/webhook.go
```diff
@@ -119,14 +127,29 @@ func (wh webhook) PostToChannel(p *Plugin, instanceID types.ID, channelID, fromU
...
 	} else {
-		post.Message = wh.headline
+		post.Message = headline
+	}
+
+	// Atomically claim the dedup key before posting so concurrent webhook
+	// deliveries can't both pass the check and post duplicates. SetAtomic(nil)
+	// only writes when the key does not already exist.
+	dedupKey := channelPostDedupKey(instanceID, &wh, channelID)
+	claimed, kvErr := p.client.KV.Set(dedupKey, true, pluginapi.SetExpiry(webhookDedupTTL), pluginapi.SetAtomic(nil))
```
nang2049, line 144: The dedup key is claimed here but never released if CreatePost fails a few lines below. If the first of several duplicate deliveries claims the key and then fails to post some deliveries are skipped and the event is lost.

Suggest releasing the claim on the failure path when `claimed` is true:

```go
if err := p.client.Post.CreatePost(post); err != nil {
    if claimed {
        if delErr := p.client.KV.Delete(dedupKey); delErr != nil {
            p.client.Log.Warn("PostToChannel: failed to release dedup key after post failure", "key", dedupKey, "error", delErr.Error())
        }
    }
    return nil, http.StatusInternalServerError, err
}
```

PostNotifications has the same gap so a shared `releaseDedupKey` helper may be cleaner than duplicating this.

server/webhook_test.go
```diff
@@ -0,0 +1,116 @@
...
+func isDedupClaimOptions(opts model.PluginKVSetOptions) bool {
+	return opts.Atomic && opts.OldValue == nil && opts.ExpireInSeconds == int64(webhookDedupTTL/time.Second)
+}
+
+func TestPostToChannelDeduplicatesConcurrentDeliveries(t *testing.T) {
+	t.Run("concurrent deliveries for the same channel post only once", func(t *testing.T) {
+		api := &plugintest.API{}
+
+		var kvWinners atomic.Int32
+		api.On("KVSetWithOptions", mock.AnythingOfType("string"), mock.Anything, mock.MatchedBy(isDedupClaimOptions)).Return(func(string, []byte, model.PluginKVSetOptions) (bool, *model.AppError) {
```
nang2049: This stub matches the key with `mock.AnythingOfType("string")` and decides its return by call count so it cannot observe the dedup. I verified this by patching `channelPostDedupKey` to return a unique key on every call and still passes all three subtests in this test.

A map based fake makes them real tho

```go
var mu sync.Mutex
claimed := map[string]bool{}
api.On("KVSetWithOptions", mock.AnythingOfType("string"), mock.Anything, mock.MatchedBy(isDedupClaimOptions)).
    Return(func(key string, _ []byte, _ model.PluginKVSetOptions) (bool, *model.AppError) {
        mu.Lock()
        defer mu.Unlock()
        if claimed[key] {
            return false, nil
        }
        claimed[key] = true
        return true, nil
    })
```

server/webhook_test.go
```diff
@@ -0,0 +1,116 @@
...
+		post, _, err := newTestChannelWebhook().PostToChannel(p, "instance-1", "channel-1", "bot-user-id", "subscription-b")
+		require.NoError(t, err)
+		require.Nil(t, post)
+
+		api.AssertExpectations(t)
+	})
+
+	t.Run("different channels each get their own post", func(t *testing.T) {
+		api := &plugintest.API{}
+		api.On("KVSetWithOptions", mock.AnythingOfType("string"), mock.Anything, mock.MatchedBy(isDedupClaimOptions)).Return(true, (*model.AppError)(nil)).Twice()
```
nang2049: Same issue as the stub in the first subtest: this always returns `true` regardless of the key, so this test passes even if `channelID` were dropped from `channelPostDedupKey` entirely. The map-based fake plus `require.Len(t, claimed, 2)` is what actually asserts "different channels each get their own post".

nang2049 (commented):
Thanks @avasconcelos114

nang2049 (approved):

nang2049 (approved):
LGTM!

ogi-m (approved):
Working as expected!

PR #1340, opened 2026-08-18
persist Jira team field discovery so team subscription filters match reliably

nang2049:
#### Summary
Team subscription filters silently never matched on most instances. On any instance whose team field is not `customfield_10001` the webhook lookup read the wrong key and the filter rejected every issue.

Because the cache was per process this worked only on nodes where someone had opened the create-issue modal only until the next restart.

- Persist discovered keys in KV per instance under `team_field_keys`.
- Remove the `defaultTeamFieldKey` fallback. Guessing `customfield_10001` risked matching an unrelated field rather than just failing.
- Log a warning when a team filter is evaluated with no resolved keys.

#### Ticket Link
https://hub.mattermost.com/pde/pl/bc87psyf7jgdidfybqynpfgd3a

#### Test
`server/team_fields_test.go` covers the removed fallback the KV round trip and its memoization.

Manual on an instance whose team field is not customfield_10001
1. Disable and re enable the plugin then create a team filtered subscription without opening the create issue modal.
2. Create a matching issue in Jira and confirm the notification posts.
3. Disable and re enable the plugin again and create another matching issue and confirm it still posts.
4. Confirm a subscription filtered on a different team does not post.

## Change Impact: 🟠 Medium

**Reasoning:** The changes span Jira client, subscription, issue, and KV persistence layers. Tests cover key discovery and persistence, but team filtering remains a user-facing path.

**Regression Risk:** Medium. The removed fallback can affect cold-start behavior, and KV or Jira field discovery failures can affect team filters.

**QA Recommendation:** Perform manual QA for team-filter creation, editing, and evaluation across plugin restarts and multiple Jira instances. Skipping manual QA carries moderate risk.

*Generated by CodeRabbitAI*

### DIFF
--- a/server/client.go
+++ b/server/client.go
@@ -27,6 +27,7 @@ import (
 
 const autocompleteSearchRoute = "2/jql/autocompletedata/suggestions"
 const commentVisibilityRoute = "2/user"
+const fieldRoute = "2/field"
 const userSearchRoute = "2/user/assignable/search"
 const unrecognizedEndpoint = "_unrecognized"
 const visibleToAllUsers = "visible-to-all-users"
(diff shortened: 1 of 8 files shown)


avasconcelos114 (approved):
LGTM!

ogi-m (approved):
I don't have an instance where the team field key is not customfield_10001, so I can't reliably test the fix. 

I have verified that subscriptions are working correctly with the team filter.

PR #1353, opened 2026-09-11
MM-70571 Improving cleanup logic for user disconnect flow

avasconcelos114:
#### Summary

This PR improves the logic for disconnecting users and allows them to remove any connection entries that may exist even if the record of the target instance may have been removed from the KV store

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-70571

#### QA Notes
This is self-healing mechanism that should just allow users to be able to disconnect from an instance no matter what. The main flow of insterest is 
1. An admin set up an instance
2. a user connects to that same instance 
3. The admin removes that instance 
4. The user should still be able to disconnect from it rather than being locked and unable to take further actions of their own

## Change Impact: 🔴 High

**Reasoning:** Changes affect core user connection, instance uninstall, and KV persistence flows across multiple modules. They also change shared interfaces and cleanup behavior.

**Regression Risk:** High. Failures may leave stale connections, incomplete cleanup, or inconsistent instance data.

**QA Recommendation:** Perform thorough manual QA for connect, disconnect, uninstall, stale-instance recovery, KV read failures, and Jira error responses. Skipping manual QA carries a high regression risk.

*Generated by CodeRabbitAI*


server/instances.go
```diff
@@ -177,64 +176,111 @@ func (p *Plugin) InstallInstance(newInstance Instance) error {
...
 	if err != nil {
+		p.errorf("UninstallInstance: failed to sweep users of instance %q, leaving their records to self-heal: %v", instanceID, err)
+	}
+
+	// Delete the blob only after the list no longer references it: a list
+	// entry pointing at a deleted blob is the state that strands users. A
+	// leftover blob does not, but nothing can reach it to retry either.
+	if err := p.instanceStore.DeleteInstance(instanceID); err != nil {
+		p.errorf("UninstallInstance: failed to delete instance blob %q: %v", instanceID, err)
+	}
```
nang2049, line 230: Both of these failures are logged and then dropped, so `executeInstanceUninstall` prints "Jira instance successfully uninstalled". I don't think either needs to become a returned error as the ordering argument for continuing is sound. But the silence might be a problem

server/instances.go
```diff
@@ -340,14 +386,26 @@ func (p *Plugin) resolveUserInstanceURL(user *User, instanceURL string) (types.I
...
+
+	connected := NewInstances()
+	for _, id := range user.ConnectedInstances.IDs() {
+		if instances.Contains(id) {
+			connected.Set(instances.Get(id))
+		}
+	}
+	if connected.IsEmpty() {
+		return "", errors.Wrap(kvstore.ErrNotFound, "your account is not connected to Jira. Please use `/jira connect`")
+	}
```
nang2049, line 403: This is the case id like to see covered because it's the flow in the QA notes on a single instance server.
Suggest when `connected` is empty but `user.ConnectedInstances` is not fall back to the stale reference so the disconnect can clean itself up.

server/kv.go
```diff
@@ -278,35 +280,40 @@ func (store store) CountUsers() (int, error) {
...
-				return err
-			}
+	failedReads := 0
+	for _, key := range keys {
+		user := NewUser("")
+		if err := store.get(key, user); err != nil {
+			store.plugin.errorf("MapUsers: failed to load Jira user for key %q: %v", key, err)
+			failedReads++
+			continue
 		}
```
nang2049, line 310: `store.get` delegates to `client.KV.Get`, which returns a nil error and leaves the destination untouched when the key is gone. So for a key that was listed above but deleted before we read it this loop doesn't continue. That might contradicts the new interface doc of "skipping and counting the records it cannot load" and mid-iteration deletion iswhat we want to handle. A guard after the read would close it:

```go
if user.MattermostUserID == "" {
    failedReads++
    continue
}
```

nang2049 (commented):
Thanks @avasconcelos114  a couple of minor comments!

nang2049 (approved):

